PR #302, opened 2012-09-13
Fluentmigrator.Runner support for multiple assemblies at a time

MikeBagos:
Added support for FluentMigrator.Runner to proccess multiple assemblies at a time, passed as a collection of MigrationAssemblyInfo packages. Kept compatibility with older code as existing constructors were replaced to point to new constructors.

Also fixed the test, which pass. Extensive test were not created/performed.


tommarien:
Should be closed by pulling #467

PR #386, opened 2013-03-04
Fluent API for programmatic access to migrations and direct queries

Athari:
Added `Migrator` class for simple and easily readable programmatic access to migrations (it's a facade for the library). Interfaces of a few other classes are wrapped too, to make API simpler and more _fluent_.

In addition to usual methods like `MigrateUp` and `Rollback`, it contains a query class for making direct queries to the database, without recording migrations history. This part of API can be seen as a complement to @markrendle's Simple.Data. Together with FluentMigrator, they provide both DML and DDL.

To be most useful, a new NuGet package should be added with FM.Runner.

Example code with migrations:

``` csharp
using (var migrator = new Migrator()
    .LoadMigrations(typeof(DataMigration).Assembly)
    .OpenConnection("SqlServerCe", "...")) {
    migrator.MigrateUp();
}
```

Example code with direct queries:

``` csharp
using (var migrator = new Migrator().OpenConnection("SqlServerCe", "..."))
    var query = migrator.GetQuery();
    query.Create.Table("Users")
        .WithColumn("Id").AsInt32().PrimaryKey().Identity()
        .WithColumn("Name").AsString()
        .WithColumn("Email").AsString();
    query.Process();
}
```


fubar-coder:
I agreee with @tommarien that the query part doesn't fit in the scope of this project. The fluent migrator part looks good, but IMHO it'd be better to provide a way to integrate into a DI framework nicely.

PR #405, opened 2013-04-27
Fixed some bugs related to Firebird and SqlSchemaDumper

woehrl01:
The Firebird processor had a bug which reverted successfull migrations when dispose on the processor is called at the end of a migration. This patch clears the stack of the changes after commit, to prevent the execution of them at the rollback step (on dispose).

Additionally the SqlSchemaDumper has been improved to have a more correctly DbType mapping. But I don't know how I should map the remaining types.

Finally a bug in SqlSchemaDumper has been fixed which selected the wrong column at the foreignkey extraction part.


tommarien:
I consider this pull as closed, but i am eager to see any clean pr's coming ;)

PR #433, opened 2013-07-16
Allow CurrentUTCDateTime SystemMethods in SQLLite

Merwer:
CURRENT_TIMESTAMP is actually UTC time - I'm not sure what would be best to handle local time in SQL Lite.


fubar-coder:
I'll try to incorporate this changes directly.

fubar-coder:
Integrated PR.

ParagonIG:
The code produces myTimestamp DATE DEFAULT datetime('now','localtime') not myTimestamp DATE DEFAULT (datetime('now','localtime')) as required by SQLite so any migration attempting to set the default value of a datetime to the current local time fails.
see Item #743

PR #474, opened 2014-02-03
Support for multiple databases (internal only, not exposed through console)

ashmind:
This pull request adds support for migrating multiple related databases that have common versioning.
The database with version table is considered default and `Create` etc work as they did before, but the syntax below can be used to redirect certain changes to related databases.

**Syntax**
    
    InDatabase("Other")
        .Create.Table("Test").InSchema("test").WithColumn("Id").AsInt32();

**Notes on implementation**

1. This feature requires `MultiDatabaseMigrationProcessor` which wraps a set of other processors. This processor is currently not created when using console or MSBuild/NAnt tasks. The reason is that I do not need it at the moment (I use Runner manually), but I feel this pull request may be useful nevertheless.

2. Transaction support is rough -- failing transaction on one DB may not revert transactions in others. I do not see a perfect solution for that, unless something custom is done with MSDTC (do all DBs support `TransactionScope` properly?)

3. I tried to avoid breaking changes, but there is one breaking change where a new Process overload is added to the `IMigrationProcessor`. I hope everyone is using `ProcessorBase` and that would be OK.

3. New syntax makes old `IfDatabase` confusing. I would argue `IfDatabase` was already confusing and should be `IfEngine` or something like that. But it may also mean my extension should be renamed. I just did not have enough time for perfect naming.

4. Integration tests on MultiDB are only done in SQL Server 2008.

5. `IInDatabaseExpressionRoot` is a copy of `IIfDatabaseExpressionRoot`. Not sure if it should be reused instead (I noticed that it is not reused between `IfDatabase...` and similar code in `MigrationBase`).

6. I am not sure what level of XML comments is expected. I added some where similar classes had them or where it seemed useful. Let me know if more would be required.

7. For this, and for `InSchema` I feel supporting some kind of `using` may be a good idea. However this is probably not possible to implement without major changes.

### DIFF
--- a/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj
+++ b/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj
@@ -114,7 +114,6 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.configuration" />
-    <Reference Include="System.Transactions" />
     <Reference Include="System.Xml.Linq">
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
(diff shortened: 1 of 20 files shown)


tommarien:
@ashmind I'll close this pull for now until extensibility hook arrives

PR #497, opened 2014-06-05
implemented possibility to migrate up and down between start and target ...

CodeWithCream:
...versions


fubar-coder:
This is already (indirectly) implemented using `IRunnerContext.StartVersion`.

Issue #529, opened 2014-09-21
Add Support to SAP HANA Database


FabioNascimento:
Hello,

I made this [commit!](https://github.com/FabioNascimento/fluentmigrator/commit/e9a7e88c514e53797356cd9f11051f897cd9e234) to add support to the database that I'm work it and I would like to see in excelent library.

tks

PR #536, opened 2014-11-04
Added support do SAP Hana #529 

FabioNascimento:
Added suporte do SAP Hana #529


tommarien:
@FabioNascimento Thank, sorry for the pull delay ;)

FabioNascimento:
@tommarien  You welcome...
Should I write documentation to this resource ? If yes, let me know.

tommarien:
@FabioNascimento I will never stop you from adding documentation, that would be weird :+1:

PR #553, opened 2014-12-17
add comment feature for postgresql

itn3000:
add feature which enables comment on table and column by WithDescription and WithColumnDescription.


itn3000:
I'm sorry for late response,I fixed it.

tommarien:
@itn3000 Thanks

PR #564, opened 2014-12-31
Updated forward only migration resharper file template.

ahmedshuhel:
Included file name in the migration class.


daniellee:
Yes, that was an improvement. I added it to the [wiki](https://github.com/schambers/fluentmigrator/wiki/Resharper-File-Template) too. Thanks again!

ahmedshuhel:
Thank you.

PR #566, opened 2015-01-02
Capitalize SQLite classes and filenames

### LINKED ISSUE #552: SQLite class / namespace capitalisation
screen shot 2014-12-16 at 11 56 05

Should be capital 'QL' in `SQLiteProcessor` and so on... right? Because you've changed the namespaces to be capitalised. Should be consistent...

Suggestion:

`grep -R --include '*.cs' Sqlite .`

daniellee:
Changes Sqlite to SQLite in class names, filenames, and in project references. Should close #552.


tommarien:
@daniellee Thanks

PR #570, opened 2015-01-12
Add options to generate migrations without database connection

andrewhanson:
I'm interested in thoughts and feedback on the following scenario and migration runner changes. 

__Scenario__
In one of our current build scenarios we are required to provide our operations team SQL migration scripts for deployment into our production environment but because of security considerations we are not allowed to connect to the production database server nor are we allowed to run database migrations from within our code.

Because of this we required the ability to generated migrations SQL scripts for a given version range without connecting to a database.

__Proposed Changes__
This pull request adds two options to the IRunnerContext that will allow us to run migrations without a connection to a target database:

New Options:
* startVersion: the version number to start generating migrations from 
* noConnection: indicates that we want to run the migration without a connection string.

_Running a connectionless migration_

    Migrate.exe
    --version 0 --startVersion 2  --output --preview --task rollback -db SqlServer2012 --noConnection 
    --assembly   C:\Code\MigrationScripts.dll
    
To accomplish this we added a new ConnectionLessVersionLoader and ConnectionlessProcessor.  The rest of the included changes are simply used to load these classes when required.

__ConnectionLessVersionLoader__
Instead of querying the version info table for applied migrations like the current VersionLoader this class loads all the available migrations in the target assembly and loads the applied migrations with any migration less than the supplied 'startVersion' parameter

__ConnectionlessProcessor__
Instead of executing SQL against a database connection this processor simply emits the generated SQL to the Announcer.

### DIFF
--- a/src/FluentMigrator.Console/MigratorConsole.cs
+++ b/src/FluentMigrator.Console/MigratorConsole.cs
@@ -48,6 +48,8 @@ namespace FluentMigrator.Console
         public int Timeout;
         public bool Verbose;
         public long Version;
+        public long StartVersion;
+        public bool NoConnection;
         public string WorkingDirectory;
         public bool TransactionPerSession;
         public string ProviderSwitches;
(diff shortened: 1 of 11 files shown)


tommarien:
Very interesting feature, i wonder if we could reuse some api changes to implement #497 ?

andrewhanson:
It's definitely the same versions to apply requirement as #497 I don't see why we couldn't combine the two approaches.

NikGovorov:
How does it handle migrations that use WithConnection method?

andrewhanson:
I don't think "WithConnection" would work in this context.

On of the downsides I saw when working on this was that there are some
features that just seem to need a live connection to work

tommarien:
@andrewhanson Do you mind adding documentation to the wiki about this feature, because i think this is a good start for something beautiful and usefull in environments where you don't have credentials to run the migrations and you need to provide a script ;)

Thanks

Issue #545, opened 2014-11-19
SetExistingRowsTo, changing column length


katlimruiz:
This code:
```C#
            Alter.Table("Attachment")
                .AddColumn("Group").AsString(3).NotNullable().SetExistingRowsTo("ATT");
```
creates the column as nullable, updates the rows, and then sets as not nullable.

However, the length of the column ends up as **nvarchar(255)**!.

The workaround is to not use SetExistingRowsTo, add column nullable, update and the alter as not nullable manually.

Which obviously works, but is less than optimal.

PR #572, opened 2015-01-19
Fixed issue #545 

mstancombe:
Fixed Issue #545 where not all column definition data was set when using SetExistingRowsTo command, changed ColumnDefinition.Clone to be virtual for testability.


tommarien:
@mstancombe Thank you

PR #573, opened 2015-01-29
Provides an ability to create a column and specify that the timezone information should be stored with the date

dealproc:
Provides the ability to store timezone information for a given date within the database.

C# syntax would be:
`Create.Table("Table1").WithColumn("CreatedOnUTC").AsDateTimeOffset();`

Sql Server >= 2008 & Postgres implementations are in place already.

I also added the `DbType.DateTime2` mapping to the `timestamp` type for postgres, as the `timestamp` time in postgres can hold a larger range of date/time values.


tommarien:
Thanks for the clean pull ;)

dealproc:
:+1: 
I hope I have the fix in there, can you double check..

I've been using a custom build since i brought this up 18 months ago... there was an issue i unearthed when i re-worked this PR that if you run the `Insert.Into("Table").Record(new { datetimeoffsetcolumn = DateTimeOffset.UTCNow })` statement, the date/time wasn't being quoted, and the insert statement would fail.  

I know i corrected it in my working folder, but am unsure if you've received it as part of this PR or not.

If you can confirm... or otherwise, I'll see about providing that patch as well.

PR #575, opened 2015-02-03
Fix FirebirdProcessor

amroel:
FirebirdProcessor reverts all actions after it's disposed


tommarien:
@amroel Thanks, seems all very reasonable fixes ;)

PR #576, opened 2015-02-05
When using Firebird quote only keywords otherwise quoting should be optional and not the default

amroel:


tommarien:
@amroel it's sometimes difficult to follow, but i like the multiple commit style, thanks !

src/FluentMigrator.Runner/Generators/Firebird/FirebirdQuoter.cs
```diff
@@ -1,15 +1,64 @@
...
+            "SQLSTATE", "SQRT", "STABILITY", "START", "STARTING", "STARTS", "STATEMENT", "STATISTICS", "SUB_TYPE", "SUBSTRING", "SUM", "SUSPEND",
+            "TABLE", "TAN", "TANH", "TEMPORARY", "THEN", "TIME", "TIMEOUT", "TIMESTAMP", "TO", "TRAILING", "TRANSACTION", "TRIGGER", "TRIM", 
+            "TRUNC", "TWO_PHASE", "TYPE", "UNCOMMITTED", "UNDO", "UNION", "UNIQUE", "UPDATE", "UPDATING", "UPPER", "USER", "USING", "UUID_TO_CHAR",
+            "VALUE", "VALUES", "VARCHAR", "VARIABLE", "VARYING", "VIEW", "WAIT", "WEEK", "WEEKDAY", "WHEN", "WHERE", "WHILE", "WITH", "WORK",
+            "WRITE", "YEAR", "YEARDAY"
+        };
+
+        public override string Quote(string name)
+        {
+            if (_keywords.Any(x => x == name.ToUpper()))
```
hazzik: This can fail on "turkish i problem"

amroel:
What is a "turkish i problem" @hazzik?

hazzik:
@amroel 

* https://msdn.microsoft.com/en-us/library/ms973919.aspx
* http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html
* http://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/
* http://www.i18nguy.com/unicode/turkish-i18n.html
* http://blog.codinghorror.com/whats-wrong-with-turkey/

Basically, `i` in Turkish capitalized to `İ`, and lowercase version of `I` is `ı`

Which makes `"similar".ToUpper()` not equal to `"SIMILAR"`

hazzik:
So, the `_keywords` should be a HashSet<string>() with ordinal ignore case comparer.

amroel:
Thank you @hazzik I see the point.
@tommarien I think a new PR fixing this would be reasonable. What do you think?

tommarien:
@amroel I think a string comparison of ordinalignorecase would be just fine in this case, and yes a pull is fine but if you want i can cherrypick your new commit in your branch ;)

@hazzik Do you agree on the ordinalignore case ?

amroel:
according to https://msdn.microsoft.com/en-us/library/ms973919.aspx posted by @hazzik: 
"**ToUpperInvariant** and **ToLowerInvariant** are also available. **ToUpperInvariant** is the *standard* way to normalize case. Comparisons made using **OrdinalIgnoreCase** are behaviorally the composition of two calls: calling **ToUpperInvariant** on both string arguments, and doing an **Ordinal** comparison."

Wouldn't then a simple switch from ToUpper to ToUpperInvariant suffice in this case?

tommarien:
@amroel i guess but a simple test can prove it ;)

amroel:
I did and it works :)
Just wanted to know your thoughts and/or if you favor a specific solution.

tommarien:
@amroel i think if it works the same toUpperInvariant is fine

amroel:
Ok. Here are the relevant changes:
https://github.com/amroel/fluentmigrator/commit/776af690cf5ef97aac5da7ce8c70cba86b2f62e7
https://github.com/amroel/fluentmigrator/commit/a182406dfbac007963952c133ca27e396f822406

I'll wait to hear hazzik's openion before I submit a pull request. Agree?

tommarien:
@amroel Agreed

@hazzik Any remarks ? (btw thanks for being a critical eye)

jgillich:
@amroel I might be missing something, but how do I re-enable quoting?

amroel:
@jgillich simply quote what you need to be quoted!
Or did I misunderstood your question?

tommarien:
@amroel I think what @jgillich means is how to fallback to the old (always quoted behavior) like for instance on oracle see https://github.com/schambers/fluentmigrator/wiki/Command-Line-Runner-Options#--providerswitchesvalue

amroel:
Ah ... well it's certainly doable but I'm not sure if it's also desirable!
@jgillich do you really need to quote everything when using Fb, which will force you to literally quote each and every Fb object (Tables, columns, FK's, PK's, sequences, indexes etc.) in DDL's and DML's?
Again, you can still manually quote what you need.

hazzik:
>Ok. Here are the relevant changes:
>amroel@776af69
>amroel@a182406

>I'll wait to hear hazzik's openion before I submit a pull request. Agree?

Looks fine for me, however I prefer other way around. But does not matter as soon it solves the problem

PR #577, opened 2015-02-06
Added ability to set Transaction Behavior and use Tags with MaintenanceAttribute

unrealguard:
Added the ability to set the TransactionBehavior for a maintenance migration and added the ability to Tag a maintenance migration. I needed this functionality because as part of our QA database migration, we wanted to load the majority of the database from a backup and this has to be done outside of a transaction with SQL Server.

I added unit test for all the added functionality and some of the existing maintenance functionality. The code has also been tested in our QA database migrations.

### DIFF
--- a/src/FluentMigrator.Runner/IMaintenanceLoader.cs
+++ b/src/FluentMigrator.Runner/IMaintenanceLoader.cs
@@ -16,10 +16,13 @@
 //
 #endregion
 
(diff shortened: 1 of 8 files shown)


tommarien:
@unrealguard Looks great, will try to pull it asap

tommarien:
@unrealguard Thanks

PR #582, opened 2015-02-22
Allows to specify custom version table meta data

istaheev:
Introduction
------------
I have a PostgreSQL database with many independent schemas where all schemas have similar database structure and new schemas can be created on the fly.
Since I am already using FluentMigrator in the project and have excellent experience with it I tried to use it to populate schemas but, unfortunately, it failed.

The problem and solution
-----------
In this scenario every schema can be interpreted as an independent database. Each schema should have its own `VersionInfo` table, so I can run FluentMigrator for every schema and get expected result. For PostgreSQL this can be achieved very easily by setting SearchPath in the connection string ("...;SearchPath=schema1") and all migrations should now be applied to this schema only (and `VersionInfo` table will be created there too).
Unfortunately the PostgreSQL backend has strange behavior to use "public" as the default replacement for unspecified schema names, it makes this scenario impossible.

I can solve this on the migrations level by setting `ApplicationContext` to a schema name and using it everywhere where a schema name is expected, e.g. `Create.Table("test").InSchema((string)ApplicationContext)`. But there is no workaround for `VersionInfo` table, one can specify only static name for it via an object implementing `IVersionTableMetaData` interface which can only be instantiated from the assembly.

This patch allows to specify custom `IVersionTableMetaData` object via `IRunnerContext`. If the custom object is not set then old behavior is used (search for a class in the assembly).

### DIFF
--- a/src/FluentMigrator.Runner/IMigrationRunner.cs
+++ b/src/FluentMigrator.Runner/IMigrationRunner.cs
@@ -1,11 +1,14 @@
 using System.Reflection;
 
+using FluentMigrator.VersionTableInfo;
+
 namespace FluentMigrator.Runner
 {
     public interface IMigrationRunner : IMigrationScopeStarter
     {
         IMigrationProcessor Processor { get; }
         Assembly MigrationAssembly { get; }
+        IVersionTableMetaData CustomVersionTableMetaData { get; }
         void Up(IMigration migration);
         void Down(IMigration migration);
         void MigrateUp();
--- a/src/FluentMigrator.Runner/Initialization/IRunnerContext.cs
+++ b/src/FluentMigrator.Runner/Initialization/IRunnerContext.cs
@@ -18,6 +18,8 @@
 
 using System.Collections.Generic;
 
+using FluentMigrator.VersionTableInfo;
+
 namespace FluentMigrator.Runner.Initialization
 {
     public interface IRunnerContext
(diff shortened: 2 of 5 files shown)


tommarien:
@istaheev How about passing ApplicationContext to VersionTableMetaData ? would that solve the problem ? I know it's not passed at the moment but we could

istaheev:
@tommarien This approach might be even better, but to implement it we need to extend ```IVersionTableMetaData``` with ```ApplicationContext { get; set; }``` property which will be initialized by the ```VersionLoader``` class. This might break backward compatibility with existing user code using FluentMigrator.
Another drawback of this approach is that FluentMigrator doesn't support several VersionTableMetaData definitions per an assembly. I can run migrations from some namespace but I can't tell FluentMigrator to use VersionTableMetaData from same namespace. So, if I have special VersionTableMetaData which uses ApplicationContext to setup proper schema I can't use it for other migrations in the same assembly (which might have different requirements for ApplicationContext).
What do you think about this? I am new to FluentMigrator's internals and can miss something.

istaheev:
@tommarien Just got a thought that we can create another ```IVersionTableMetaDataAnotherExtendedVersion``` (by analogue with ```IVersionTableMetaDataExtended```) with ```ApplicationContext { get; set; }```, and check if ```VersionTableMetaData``` implements it to set proper application context. This will not break backward compatibility, but the solution looks ugly.

tommarien:
@istaheev Mmm, i see your point but again i feel documenting the change as breaking if you don't inherit from defaultversionmetadata seems ok too. If we document it...

istaheev:
@tommarien Ok, sounds good, I will try to implement it.

tommarien:
@istaheev Thanks and looking forward to the pull :)

istaheev:
@tommarien: I created another pull request with new changes. If you are ok with them you can reject this one.

PR #583, opened 2015-02-25
FirebirdQuoter handles the turkish i problem

amroel:


tommarien:
@amroel Thanks for the fix and again props for having clean commits ;)

amroel:
Glad to be of any help @tommarien

PR #585, opened 2015-02-26
Added Transaction to expression.Operation

albatrossen:
This enables .WithConnection expressions in the upgrade for the Jet databases


tommarien:
@albatrossen Thanks

PR #586, opened 2015-02-28
Add application context to versiontablemetadata

istaheev:
Introduction
------------
I have a PostgreSQL database with many independent schemas where all schemas have similar database structure and new schemas can be created on the fly.
Since I am already using FluentMigrator in the project and have excellent experience with it I tried to use it to populate schemas but, unfortunately, it failed.

The problem and solution
-----------
In this scenario every schema can be interpreted as an independent database. Each schema should have its own `VersionInfo` table, so I can run FluentMigrator for every schema and get expected result. For PostgreSQL this can be achieved very easily by setting SearchPath in the connection string ("...;SearchPath=schema1") and all migrations should now be applied to this schema only (and `VersionInfo` table will be created there too).
Unfortunately the PostgreSQL backend has strange behavior to use "public" as the default replacement for unspecified schema names, it makes this scenario impossible.

This can be solved by setting `ApplicationContext` to a schema name and using it everywhere where a schema name is expected, e.g. `Create.Table("test").InSchema((string)ApplicationContext)`, but there is no way to say same for to `IVersionInfoMetaData` implementations. This patch modifies `IVersionTableMetaData` to be able to access `ApplicationContext`.

Another change this patch introduces is ability to search for `IVersionTableMetaData` implementations in particular namespaces (same namespaces used to search for migrations). This allows to have several `IVersionTableMetaData` implementations in one assembly thus allowing to keep database migrations for several independent databases there.

This patch might break backward code compatibility with previous versions of FluentMigrator.


tommarien:
@istaheev Is it possible to rebase your branch to the current version of the master, as a lot changed in release 1.5.0 ?

istaheev:
@tommarien: Sorry for the delay, I was busy with other things. Finally I have merged master and solved conflicts.

tommarien:
@istaheev Don't be sorry we are all doing this in our free time, thanks for the pull ;)

PR #589, opened 2015-03-04
ROWGUIDCOL on SqlServer

stsrki:
Added option to set ROWGUIDCOL attribute when creating table (SqlServer
only)

```
Create.Table( "TestTable" ).InSchema( "TestSchema" )
        .WithColumn( "TestTableID" ).AsGuid().RowGuid().PrimaryKey().WithDefault( SystemMethods.NewSequentialId ).NotNullable();
```

ps. I had to remove previous pull request because I messed something while trying to add another pull request.

### DIFF
--- a/src/FluentMigrator.Runner/Extensions/SqlServerExtensions.cs
+++ b/src/FluentMigrator.Runner/Extensions/SqlServerExtensions.cs
@@ -102,5 +102,12 @@ namespace FluentMigrator.Runner.Extensions
             throw new InvalidOperationException("The seeded identity method can only be called on a valid object.");
         }
 
+        public static ICreateTableColumnOptionOrWithColumnSyntax RowGuid( this ICreateTableColumnOptionOrWithColumnSyntax expression )
+        {
+            CreateTableExpressionBuilder cast = expression as CreateTableExpressionBuilder;
+            if ( cast != null )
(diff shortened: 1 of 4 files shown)


tommarien:
@stsrki Is it possible to add some tests for this feature ? also again i would advise to keep formatRowGuid only on sql server column class instead of adding it to columnbase

stsrki:
Ok I will try to do it as soon as possible.

stsrki:
Everything is moved to the SqlServer provider except for the IsRowGuid in ColumnDefinition class.

stsrki:
I'm now trying to implement replication in PostgreSQL with Slony-I and it seems to have similar logic as SqlServer. They both require that there is a unique key contraint in the replicated table. In SqlServer it is ROWGUIDCOL and in PostgreSQL it is UNIQUE.
What do you recommend before I implement this in FluentMigrator? Do I still need to separate it for every provider or can I define it in ColumnBase. Maybe it can be later user for other database providers.

tommarien:
@stsrki Why not use a sql server specific extension here AsRowGuid ?

fubar-coder:
Are the unit tests the only thing that's missing?

tommarien:
@fubar-coder seems to be ;)

fubar-coder:

PR #590, opened 2015-03-04
Improvements for PostgreSql provider

stsrki:
1. Added ability to set default value for uuid column to generate
sequential uuids.
2. Fixed generation of column/table description by adding quoter to the
generator.


tommarien:
@stsrki looks good, is uuid_generate_v1() a true alternative for sql's SequentialId ?

stsrki:
Well it seems to be sequential. I tested it by populating my test table and it works as expected when sorting.

Quick google search: http://labria.github.io/2013/04/28/rails-4-postgres-uuid-pk-guide/

tommarien:
@stsrki So it's more Guid.Comb alike then, but for me it's fine ;)

tommarien:
@stsrki Thanks, very cleanly done

PR #591, opened 2015-03-05
Default Oracle Quoter now quotes known keywords

### LINKED ISSUE #580: Oracle: auto quote reserved keywords
For instance if you use 

Insert.Into("BLAH").Row(new { GROUP = "TEST"});

It will fail and apart from using the dictionary overload and adding the quotes yourself you can't do anything about it.

See http://docs.oracle.com/cd/B19306_01/em.102/b40103/app_oracle_reserved_words.htm#CHDCGECI for a list of reserved keywords

tommarien:
Did not choose to include every keyword in the tests, as this would lead to a test explosing. Test's just try to prove the fix.

Closes #580


tommarien:
Not correct, seems only happening when using data expressions

PR #592, opened 2015-03-25
improve resharper templates

woehrl01:


woehrl01:
Prepending the classname with the version number allows to have the correct sorting in the project explorer automatically. Setting the description automatically prevents from typing that each time

tommarien:
@woehrl01 You do know that if you don't specify description, it is the classname by default ;)

For the prepending of the versionnumber i would suggest an alternative migration template

tommarien:
@woehrl01 For the moment i'll consider this as closed, unti i see different migrationtemplate that does not include classname

PR #595, opened 2015-03-25
improved character case handling in firebird

woehrl01:


tommarien:
Can you elaborate abit on what was changed and why ?

woehrl01:
sure: 

* FBOptions.PackKeyNames has been extended to allow the configuration of the FirebirdTruncator. Due to firebird allowing only a specific length in names. This one scrambles the name really weird. I wanted to configure to disable that (so instead getting an exception on execution instead of silently creating a weired name)

* the lower(*) stuff is to allow checking for names/constraints in the database, without knowing the correct case.

* The GenerateSetNull has a bug (sets the opposite value), this one fixes that (see http://www.firebirdfaq.org/faq103/)

* commit retain does really weired stuff on firebird. So it's better to use real transaction commiting.

* it should only store undoexpression history if undo enabled (leads to NPE otherwise)

* the SequenceExists crashes with an ArgumentException if the table does not exist, instead it should say no

woehrl01:
I totally understand what you mean. Of course you can reason about possibly. I think that this a little bit more practical. Especially this is the similar behavior to what the sqlserver one returns. 

And I think if you really have this special scenario you always can use the raw sql stuff instead which would make this even more explicit

tommarien:
@woehrl01 Thanks

amroel:
@tommarien, @woehrl01 
sorry to comment on an -meanwhile- "old" topic and an already closed PR.

I have some concerns regarding this PR. First of all I don't understand why some tests are passing on the build server when they should actually fail!
* FirebirdProcessorTests
For example the test [CanCreateAndDropSequenceWithExistCheck should fail on line 146](https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Tests/Integration/Processors/Firebird/FirebirdProcessorTests.cs#L146) since this PR handles everything case insensitive.
Indeed the test fails on my machine.

* Antother question related to the change in [CommitRetaining](https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Runner/Processors/Firebird/FirebirdProcessor.cs#L188):
Ponder for example one test in [FirebirdIndexTests](https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Tests/Integration/Processors/Firebird/FirebirdIndexTests.cs#L55). It uses the class FirebirdTestTable which is storing a [reference to the Transaction object of the processor](https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Tests/Helpers/FirebirdTestTable.cs#L36).
Now each time CommitRetaining is called it generates a new Transaction object which makes the referenced Transaction in FirebirdTestTable no longer valid and all tests should ultimatly fail!
How come they don't on the build server? They do fail on my machine.

Beside the above mentioned pure technical issues I don't see strong semantic reasons for both changes regarding case insesitivity and commit retaining.
* Case insensitivity:
The Public API already allowed work in a case insensitive manner.
For example it was already possible to do Create.Table("table1") which correctly stored "TABLE1" in Firebird.
Now if some one wishes to store table1 say as "TaBLe1" (using the quotes) this PR will prevent that without an obvious reason!

* Commit retaining:
Although it is absolutly true that the commit retaining option in Firebird is not recommended, the Transaction object on the Processor (which the change in CommitRetaining is modifying) is publically accessible!
This lastly means that anything holding a reference to that transaction object (whether it makes sense or not) will be broken after an automatic transaction commit (using CommitRetain() in this PR).
Yes, commit retaining is not recommended in Fb but it worked in the context of FluentMigrator like it should.

woehrl01:
Hi @amroel 

you are absolutely correct that these test should have failed. I don't know why, but both of them (sadly) do work on my machine. 

Regarding your semantic concerns:

I know that everyone has a different usecase so my points are regarding to my daily work and I'll explain why I did these changes.

Foreword: We use Fluent.Migrator to run our migrations for Firebird, Sql Server and Sqlite. That is the reason I want them to behave equally.

* Case insensitivity:
It's true that the public API should work in a case insensitive manner. But it doesn't do it e.g. for ```Schema.XYZ.Exists()``` checks. Sqlite and Sql Server check those case insensitiv, Firebird not. We would need to know the case sensitivity only for firebird. True this is not a big problems, but still a possibility to fail. Case sensitivity is semantically correct, but in my opinion: SQL is case insensitiv, why the hell should tablenames be?
Another thing is regarding the column names. The altering of nullable only works on Firebird if you know the exact name and provide it like that. It wouldn't even fail with an exception! Good luck debugging this mess.
Regarding creating of those tables, I don't see any change which should prevent this. (And why do you even want this kind of names? )

* Commit retaining:
You are totally correct, but commit retaining is not behaving like it should. I had a lot of troubles with this one, because of the totally weird behaviour of transactions within firebird. (Visibility of newly created tables, columns, type changes, creating indizies on newly created columns etc.). And the only working thing is commiting and reopening. 

I personally just want to have these things work the same across those vendors. I know that there are a lot of compromisses to make. I only know that these changes are working amazingly well since about 2 years in production now, with currently more than 300 migrations across all those vendors. (that's the reason why I even made the PR)

amroel:
Hi @woehrl01 
what I'm realy worried about are those tests that should fail!
We should come after the reason. It's making me very nervous :)

You're right regarding the table names. I simply missed that it's still possible using quotes.

Regarding Commit retaining:
where exactly did you have problems in the context of FluentMigrator?
I didn't come against any issues while I was (and I'm still) working with FluentMigrator targeting Firebird!
Are you eventually using the FirebirdProcessor directly in your projects?

In General you can't treat different db engines equally. That's precisley the reason why FluentMigrator has all those different "Processors" implementing a common interface.
I would do the same if I had to.

woehrl01:
@amroel 

Regarding those test, I think that has something to do with the .NET Library. There was a hugh difference between the "old" .NET 3.5 and the newer 4.0 Library. There were a lot of memory/connection leaks in the 3.5 one (and I think this is the one this project uses).

Regarding commit retaining:

I'm not sure what is was (its over a year now) but as I remember correctly, it had something to do with altering a table in the same migration as you insert data into it. (i know you shouldn't do that either, but there were cases I had to). 

Regarding treating equally:

I'm totally aware that I can't treat them equally. But with fluent.migrator I'm near, to write a single fluent statement for most of the simple cases (alter/create/drop of indicies/tables/column). Stored Procedures/Triggers etc. are of course totally different cases. 

That was a huge productivity improvement for us, compared to our old pure sql scripts and with less bugs. (If anyone has created a TIMESTAMP in Firebird and mistakenly uses TIMESTAMP in Sql Server for the same column, know what struggle this can make :) )

amroel:
@woehrl01 
I'm actually having a hard time trying to get those tests **passing** at all! Not even once.
It can't be just the .net library (if it is the cause at all).
They just do what they should, namley fail. Always.

Here is an example (fails because of the change to CommitRetaining):
```xml
<test-case name="FluentMigrator.Tests.Integration.Processors.Firebird.FirebirdIndexTests.CallingIndexExistsCanAcceptIndexNameWithSingleQuote" executed="True" success="False" time="0.069" asserts="0">
	<failure>
		<message><![CDATA[System.InvalidOperationException : Execute requires the Command object to have a Transaction object when the Connection object assigned to the command is in a pending local transaction. The Transaction property of the Command has not been initialized.]]></message>
		<...>
    </failure>
</test-case>
```
I'm astonished that they do pass on your side and on the build server. I simply don't get it.

@tommarien any Ideas?

amroel:
@tommarien 
Is [this](https://github.com/schambers/fluentmigrator/blob/master/.travis.yml#L14) what currently drives the build on travis?
If yes then it means the tests I'm refering to do not run at all! Right?
Is it the same for team city?

tommarien:
Yes that should be the same in teamcity except the notworkingonmono part

I'll try to verify asap

Verstuurd vanaf mijn iPhone

tommarien:
Just verified indeed teamcity is excluding IntegrationTests Only, not NotWorkingOnMono ;)

PR #596, opened 2015-03-25
Fix bugs and mistakes in firebird typemapping

woehrl01:


woehrl01:
don't know why this shows as failed, according to the travis log it seems due to a failure on there side

tommarien:
Thanks ;)

Issue #600, opened 2015-04-16
Migrate.exe --help doesn't work correctly


NetworKKnighT:
Migrate.exe --help doesn't work correctly, because DisplayHelp method throws ArgumentOutOfRangeException while processing "noConnection" option.
This occurs in [line 258](https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Console/MigratorConsole.cs#L258) of MigratorConsole.cs. As I understand OptionSet.WriteOptionDescriptions method throws the exception when length of Option.Description is 134.

tommarien:
@NetworKKnighT Thanks will try to fix asap, issue with optionset object

tommarien:
@NetworKKnighT Doh, i already assigned myself to this bug, have everything ready for the pull request :S

PR #602, opened 2015-04-17
Fixed issue #600

NetworKKnighT:
Fixed issue #600. it was happening because a last character was whitespace.


tommarien:
@NetworKKnighT Thanks for the fix, should be in 1.5.1 ;)

PR #610, opened 2015-05-20
Sql server exists performance

spaccabit:


tommarien:
@spaccabit i know what you are trying to achieve but i don't really like the implementation of it :S

- At first please try to work on an updated fork of fluentmigrator, and please don't rebase/update it until we ask
- SELECT CASE WHEN can be replace with a SELECT COUNT(*) or SELECT 1
- Instead of using datareader and executecommand use executescalar

Thanks

tommarien:
@spaccabit Needed to cherrypick the hell out of this pr, please start with a new fork and create a branch for every pr, before you start, and update your master without any local changes

Also i corrected invalid DBNull checking see commit https://github.com/schambers/fluentmigrator/commit/8ac7aa5db1ca096ab8fec92b906ce84f8608ec81 for more info ;)

Thanks

Issue #612, opened 2015-06-04
Oracle TypeMap must allow up to 4000 characters in VARCHAR2 / NVARCHAR2 fields


jonnybee:
see: [http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1822](http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1828) 

*The VARCHAR2 datatype stores variable-length character strings. When you create a table with a VARCHAR2 column, you specify a maximum string length (in bytes or characters) between 1 and 4000 bytes for the VARCHAR2 column*

FluentMigrator has a limit of 2000 characters and if above will create a CLOB column. The limit should be altered to allow up to 4000 characters.

The same goes for NVARCHAR2 fields. 
*The maximum length of an NVARCHAR2 column is 4000 bytes. It can hold up to 4000 characters. The actual data is subject to the maximum byte limit of 4000.*

I can submit PR if you will accept!

tommarien:
Closed by #613

PR #613, opened 2015-06-04
#612 Update Oracle TypeMap

jonnybee:
Updated Oracle typemap as per Oracle documentation:
https://docs.oracle.com/cd/B28359_01/server.111/b28320/limits001.htm#i287903 

CHAR  - max 2000
NCHAR - max 2000
VARCHAR2 - max 4000
NVARCHAR2 - max 4000
RAW     - max 2000

### DIFF
--- a/src/FluentMigrator.Runner/Generators/Base/TypeMapBase.cs
+++ b/src/FluentMigrator.Runner/Generators/Base/TypeMapBase.cs
@@ -4,7 +4,7 @@ using System.Data;
 
 namespace FluentMigrator.Runner.Generators.Base
 {
-    internal abstract class TypeMapBase : ITypeMap
+    public abstract class TypeMapBase : ITypeMap
     {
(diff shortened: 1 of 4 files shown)


tommarien:
@jonnybee could you add tests for the typemap, you can make OracleTypeMap public if needed ;)

Thanks

jonnybee:
Unit tests for OracleTypeMap has been added to PR.

tommarien:
@jonnybee Thanks, just a remark though, don't test the constants test the typemapping itself, does not matter if the constants have the correct value. What matters if that for certain criteria the typemapping switches accordingly ;)

tommarien:
@jonnybee Also wondering about RAW(max) should it not be just 

Raw up until 2000 bytes
BLOB for beyond ?

jonnybee:
Yes, you are correct. RAW(MAX) is not a valid datatype. 
RAW is up to 2000- and if more then BLOB. 

I have updated the typemap and unit tests.

tommarien:
Thanks

PR #620, opened 2015-07-01
Datetimeoffset feature patch - not quoting the values appropriately

dealproc:
So, when i implemented the `DateTimeOffset` feature, i missed the idea of making sure that, when you created the insert/update scripts within a migration, that the `DateTimeOffset`s were quoted appropriately.

Pull Request is to correct this issue.


tommarien:
Merged only commit https://github.com/schambers/fluentmigrator/commit/551871618715e9c69710d8581a8e7c9fb6b95ab4

PR #629, opened 2015-07-14
changed MSBuild Migrate task to run in separate app domain

zendever:
As discussed in issue:

https://github.com/schambers/fluentmigrator/issues/627


tommarien:
@zendever Thanks ;)

Issue #630, opened 2015-07-23
byte array throws a syntax error when inserting into postgresql


willprice:
```csharp
Insert.IntoTable("file").Row(new
{
     id = 1,
     contents = someByteArray
}
```

Causes the error
```

!!! An error occurred executing the following sql:
INSERT INTO "public"."file" ("id","contents") VALUES (1,0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000);
The error was ERROR: 42601: syntax error at or near "x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"

!!! ERROR: 42601: syntax error at or near "x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
(my byte array is a 100 item long array of zeros)

tommarien:
@willprice Can you try it with manual execute.sql

INSERT INTO "public"."file" ("id","contents") VALUES (1,"\\x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");

I am guessing that the genericquoter formatByteArray is the culprit

willprice:
That too throws the same error, the [postgresql docs](http://www.postgresql.org/docs/9.4/static/datatype-binary.html) show how to escape byte arrays. You use the format `E'\\x<data>'`, perhaps override `FormatByteData` in `PostgresQuoter` with something like
```csharp
@"E'\\x" + string.Concat(data.Select(b => b.ToString("X2")).ToArray()) + "'"
```

I'm currently using that as my workaround and calling `RawSql.Insert`.

tommarien:
@willprice That will be according to spec, i'll mark this as an up for graps issue, so maybe some good person can do the fix ;)

willprice:
Thanks for the tips on where to look on solving this! Let me know if you don't want a dependency on LINQ, if so, I'll do it without.

tommarien:
@willprice No the linq dependency is fine ;) thanks for the pull request ;)

dealproc:
[...]

PR #631, opened 2015-07-23
Fix bug #630 byte arrays aren't serialised correctly in postgresql

willprice:


tommarien:
@willprice Thanks for the pull

PR #634, opened 2015-07-31
Work on a DB invaraiant Select Syntax (WIP)

jogibear9988:
This should not be merged (yet)

It's open for discussion here!

smth like this will be possible:

            var conn = new SQLiteConnection(PersistenceConfig.GetConnectionString(PersistenceMode.FileBased));
            var proz = new SQLiteProcessor(conn, null, null, null, new SQLiteDbFactory());
            var context = new MigrationContext(new MigrationConventions(), proz, new SingleAssembly(GetType().Assembly), null, string.Empty);

            var selExp = new SelectExpressionRoot(context);

            foreach (var table in selExp.Tables)
            {
                var data = table.AllColumns.Read();
            }    

or in a Migration:

     var data = Select.Table("Test").Column("Hallo").Read();


tommarien:
@jogibear9988 Exactly the thing i feared, you end up in writing linq-alike code. What is the problem you want resolved in your specific use case. As for as far as i can see i could handle it with custom sql or scripts 

No pun intended, just asking honest questions ;)

jogibear9988:
Yes, but Linq depends on classes wich look like your db objects! what i want to solve is reading data in migrations db independent! if i use sql, it most times looks the same, but there are errors when using dates, db reserved words, ...

jogibear9988:
I can show you some migrations we have done in our project @work, wich need db data! we used migrator.net there, but for our new project i'd like to switch to fluentmigrator, cause i'd like the syntax much more!

tommarien:
@jogibear9988 ah still a migrator.net user ;) so you are are using different db providers running same migrations ?

jogibear9988:
yes! we having a mfc and wms system supporting mssql, oracle and postgres (and sqlite for development (and mariadb in unit tests))

all are running now using the same migrations! we have around 200 now where only 3 or 4 have provider specific parts!

and we also have migrations wich select various data from db! if we would have used sql directly we would have got problems because in different providers you have to quote differently, or be carful with date time!

Von meinem iPhone gesendet

tommarien:
@jogibear9988 Ok , difficult scenario to maintain but still valid ;)

What about 

Select.FromTable("Table").InSchema("schema")
Columns("Hallo") **OR AllColumns() **

.Where(new  ** OR AllRows()**
{
 Id=1
}

But what will it return a dynamic or a datareader ?

jogibear9988:
I'm closing this issue. Since after more usage, we have decided to switch to Migrator.Net (my fork), because we mostly use SqlLite and in this Project Renaming, Altering Columns is implemented, wich does not work with Fluent Migrator

PR #641, opened 2015-08-23
Add DateTimeOffset to Oracle type map

vgrigoriu:
as TIMESTAMP(4) WITH TIME ZONE


tommarien:
@vgrigoriu Thanks for the pr !

Issue #636, opened 2015-08-06
Default to Unicode strings


vgrigoriu:
Would it be possible, when inserting data, to default to Unicode strings? That is, instead of having to use `ExplicitUnicodeString`, to force people to use `ExplicitNonUnicodeString` when they don't want Unicode? I realize this might be a breaking change, but it seems like the right thing to do. Am I missing something that makes this a bad idea?

tommarien:
@vgrigoriu I guess it all depends on the use case, i never liked the naming though UnicodeString should have been better as a class name.

What you could do and has less side effects is keeping the code as is but

* Add Ansistring in same way as we now have ExplicitUnicode
* Add overload option to insert update etc which is by default false threatallstringsasunicode (expect the ones explicitly using a type like ansistring or unicodestring)

What do you think about this ?

vgrigoriu:
I don't think there's any use case in which defaulting to ANSI strings is correct. Either you only use ASCII characters, so you don't care, or you use accented characters, in which case clearly what you want is an Unicode string. If we make Unicode strings the default and someone's code behavior changes then it was almost certainly a bug.

I think the right thing to do would be:
- deprecate ExplicitUnicodeString
- add AnsiString, as you suggested
- default to formatting normal .NET strings as Unicode SQL strings.

If you agree, I'll try to send a PR this weekend.

cc @Ashthos who originally introduced `ExplicitUnicodeString`.

fubar-coder:
Fixed in 2.0.0

vgrigoriu:
Wow 😄 

Thanks for finally getting this in!

fubar-coder:
It was a rough ride. 1 month of work went into the 2.0 release.

PR #642, opened 2015-08-23
Remove ExplicitUnicodeString and add NonUnicodeString

vgrigoriu:
As I explained [here](https://github.com/schambers/fluentmigrator/issues/636#issuecomment-129391177), code generatiion for SQL Server should default to Unicode string literals, so I removed `ExplicitUnicodeString` and added `NonUnicodeString` instead.

This change affected SQL Server CE too, which I think defaults to Unicode, so `'caș'` and `N'caș'` are equivalent there, but there shouldn't be any harm in using `N'`.

Hana was also using `ExplicitUnicodeString`, so I went ahead and changed it too. @FabioNascimento, is it ok if the Hana generator dafaults to Unicode string literals?

I removed the `ExplicitUnicodeString` class. This means people who were using it will have to change their code. I can add it back and deprecate it if anyone will miss it, but I think migration code is cleaner when you use normal .NET strings.


vgrigoriu:
Fixes #636.

FabioNascimento:
Hello @vgrigoriu ,

Sorry for the late answer, but I just see now. I test your changes in HANA db and everything work fine...

Fabio

vgrigoriu:
Wow, that's a blast from the past 😃 
Who is taking care of this repo now? @fubar-coder ? Any chance of merging this in if I solve the conflicts?

fubar-coder:
This seems to be a breaking change for the end user due to the removal of `ExplicitUnicodeString`. Can you please keep `ExplicitUnicodeString` as obsolete (with explanation)? Can it cause problems when we switch to unicode by default?

vgrigoriu:
All right, I'll put `ExplicitUnicodeString` back in as obsolete.

I explained [here|https://github.com/schambers/fluentmigrator/issues/636#issuecomment-129391177] what I think are the consequences of switching to Unicode.

fubar-coder:
Also, please don't forget to handle `ExplicitUnicodeString` instances 😉 .

@schambers, I guess, that this change is OK for a minor library update when the only user visible change is a deprecated class?

vgrigoriu:
Here you are.

fubar-coder:
Thank you

Issue #645, opened 2015-09-01
Oracle: Improper Currency precision


jarz:
FluentMigrator currently creates a NUMBER(19, 1) column. According to [Oracle](https://docs.oracle.com/cd/B19306_01/gateways.102/b14270/apa.htm), the SQL Server equivalent of MONEY is NUMBER(19,4).

At the very least, the decimal precision should be bumped to two. Otherwise we're rounding to the nearest tenth on insert and update operations.

tommarien:
Closed by merge

PR #646, opened 2015-09-01
Update OracleTypeMap.cs

jarz:
Bump precision to 4 places, matching Oracle's equivalent to SQL Server's MONEY data type.


jarz:
Fixes #645

tommarien:
@jarz Thanks for the clean pull ;)

PR #647, opened 2015-09-03
Allow for types with no namespaces

stevewhite:
I am trying to upgrade to latest version and have a number of extension classes in the same assembly without namespaces, and with t.Namespace being null am getting the dreaded NullReferenceException.

This of course assumes that you put your Migrations in a Namespace.


jarz:
:+1: 

More null checks can't hurt.

tommarien:
@stevewhite Thanks for the patch

Issue #626, opened 2015-07-10
Tags attribute isn't really inheritable (even though it's "tagged" as such)...


zendever:
This may have been a discussion somewhere along the way, or there may be some other way to do this, but setting the Tag attribute on a base migration class doesn't work as I would expect.

Case in point:

on DefaultMigrationConventionTests.cs:

```c#
// inherit from tagged class
public class InheritedFromTaggedWithUk : TaggedWithUk
{
}
...
// test along side tag tester...test will fail
[Test]
public void TypeHasTagsReturnTrueIfTypeHasTagsAttribute()
{
	DefaultMigrationConventions.TypeHasTags(typeof(TaggedWithUk))
		.ShouldBeTrue();

	DefaultMigrationConventions.TypeHasTags(typeof(InheritedFromTaggedWithUk))
		.ShouldBeTrue();
}
```

The code that tests for tag existence clearly only allows the Tags attribute to act as if they were non-inheritable:

in DefaultMigrationConventions.cs

```c#
public static bool TypeHasTags(Type type)
{
	return type.GetOneAttribute<TagsAttribute>() != null; // <-- note the overload with no param
}
```

in ExtensionsForICustomAttributeProvider.cs:

```c#
// the function below IS called for the Tags attribute
public static T GetOneAttribute<T>(this ICustomAttributeProvider member)
	where T : Attribute
{
	return member.GetOneAttribute<T>(false);
}

// the function below is NOT called for the Tags attribute
public static T GetOneAttribute<T>(this ICustomAttributeProvider member, bool inherit)
	where T : Attribute
{
	T[] attributes = member.GetCustomAttributes(typeof(T), inherit) as T[];

	if ((attributes == null) || (attributes.Length == 0))
		return null;
	else
		return attributes[0];
}
```

changing the DefaultMigrationConventions.cs class to the code below seems to pass all unit tests, including the Tag inheritance test above.

```c#
public static bool TypeHasTags(Type type)
{
	return type.GetOneAttribute<TagsAttribute>(true) != null;
}
```

Again, I realize that this may be one of those finer discussion points on the "proper" usage of migrations. Maybe we didn't intend for the Migration class to be turned into a base class. We could certainly argue about whether or not Tags should be "overridden" if they're inherited. We could certainly argue about composition over inheritance...that said, if any of those were really issues, the solution would simply be to make Tags not inheritable, right? At least that way, there'd be no confusion over their purpose.

[...]

PR #648, opened 2015-09-09
Modified DefaultMigrationConventions to allow Tags to be inheritable

cbright:
See  #626 for details


tommarien:
@cbright Thanks for the pr

Issue #651, opened 2015-09-22
IndexOutOfRange Exception when rollingback a firebird migration on a table containing implicitly declared foreign key


amroel:
Having this in a firebird db

``` Sql
create table table1
(
 id bigint not null primary key
);
create table table2
(
 id not null primary key,
 col1 bigint references table1 <-- "Implicit" because it doesn't explicitly mention the pk of table1
);
```

An IndexOutOfRange exception is thrown when the following migration is rolled back:

``` C#
      [Migration(1)]
        public class TestMigration : Migration
        {
            public override void Up()
            {
                Create.Column("col2").OnTable("table2").AsString(30);
            }

            public override void Down()
            {
                Delete.Column("col2").FromTable("table2");
            }
        }
```

PR #652, opened 2015-09-22
Issue 651

amroel:
Fix for Issue #651


amroel:
I see that the 4 new tests I made are failing on Setup/Teardown with an exception "Unable to complete network request to host "127.0.0.1"". 
Strange though is that the new tests share the same code for database access as all other successfuly runing tests!

amroel:
@fubar-coder, @tommarien 
after making this PR I realized that all Firebird integration tests are not run! Neither on the build server nor localy.
This is caused by excluding tests marked with the "Integration" attribute on the build server and also by setting the IsEnabled flag to false in DatabaseServerOptions.
This lastly means that we can never be sure that "our" changes didn't break working code!
(Which obviously did happen in the past)

I have tried solving this Problem but came to the conclusion that we need a whole new test infrastructre allowing us to:
1) manually choose one or more db engines to run all test locally against during development
2) run all test on the build server against all db engines that can be (or are already) installed the server or, alternativly, have an "embeded" version (like firebird).

I might be thinking too complex so any thoughts would be apprciated!

fubar-coder:
@amroel I tested the integration tests on my machine. The idea about a more thorough build (and test) server infrastructure is good, but it's not that easy, because some tests aren't allowed to run in parallel (like the firebird tests). The tests itself have to be restructured too. It's a lot of work to be done.

amroel:
I am fully aware of the amount of work that needs to be done. The real trouble I see is not the net amount of work to restructure the tests (although it is a lot of work of course). It's more the administrative overload while doing the actual work!
Namley merging every change made in the master branch into the "development" branch and possibly fixing what will then be broken.

Although the tests do cover a whole lot of scenarios and they would have detected a lot of issues that were introduced in the past, they are currently simply worthless! Unfortunalty.

amroel:
@fubar-coder I've refactored the integration tests so that:
1) they're runnable locally on the developer machine.
2) they can run on a build server when configured to run the tests against SQLite, SqlCe and Firebird embedded.

Would you (or some one else) review my changes first before I submit a PR?
https://github.com/amroel/fluentmigrator/tree/refactor_tests

fubar-coder:
@reinaldocoelho @RenaldSB Can you please take a look at the refactored test code?

fubar-coder:
@amroel Can you please create a PR anyway? Maybe with a "WIP:" in the PR subject? This would help us/me to identify the changes you've done. Thanks in advance.

reinaldocoelho:
@fubar-coder I can take a look in the end of the day, that´s ok?

amroel:
Done #784

fubar-coder:
@amorel Thank you!
@reinaldocoelho That's OK

reinaldocoelho:
@fubar-coder I looked to refactored code, and at first it´s look well.

The system compiles without problems.
And the tests run with success, but run only on MSSQL 2014 and FireBird Embedded examples.

When running tests, the process warn all tests of other Database Engines with Ignored tests, but I don´t know if it's a problem.

The steps to run tests now is:
1. On the code root folder, create a folder with name "CurrentTestConfig".
1. Inside that folder, create database config file based on examples in "AvailableTestConfigs" folder.
1. Rename <config-file>.xml to TestConfig.xml
1. On the root folder, run tests on DLL:
 1. Command: tools\NUnit\nunit-console.exe src\FluentMigrator.Tests\bin\Debug\FluentMigrator.Tests.dll

PR #654, opened 2015-09-28
fixes #650.

### LINKED ISSUE #650: Oracle: Error when using single quote character ( ' ) as argument for WithDescription(string) method.
Oracle: Error when using single quote character ( ' ) as argument for WithDescription(string) method.

Please consider adding checks into OracleDescriptionGenerator.GenerateColumnDescription/GenerateTableDescription methods.

barser:
Please review a fix for issue #650


fubar-coder:
This seems like a partial fix only, because the description may contain both ' and $. I'll use #679 with the unit test from this PR.

fubar-coder:
Added, thanks.

Issue #637, opened 2015-08-12
DateTime2 support


krillgar:
The project that I'm currently working on requires DateTime2 support. Is there reasons why that and the other date types from SQL Server 2008+ haven't been implemented yet? I started a fork, but didn't want to go too far into it in case there was some conscious decision that was made.

immeraufdemhund:
https://github.com/schambers/fluentmigrator/blob/master/src/FluentMigrator.Runner/Generators/SqlServer/SqlServer2008TypeMap.cs

Looks like it is implemented to me (unless I'm reading the code incorrectly)

krillgar:
I added the AsDateTime2() method to [IColumnTypeSyntax](https://github.com/krillgar/fluentmigrator/commit/9ce6b31e30677cff400eaf6cd5b4ccfa6b030560). 

Does the SqlServer2008TypeMap allow me to actually create a column with those types, or does it just enable its use?

immeraufdemhund:
Well using the fluent api I tried all the date variants (as I'm sure you did as well) and couldn't get it to generate a datetime2. I see DateTime2 in a few of the Unit Tests, but it must not be hooked up. I did however get it to work by using AsCustom("DateTime2") and that made the DateTime2 column

rasmuschristensen:
Thanks @immeraufdemhund .Works on sql azure, but not sqlexpress. Needed to change it manually afterwards

fubar-coder:
Fixed in release/2.0.0

PR #657, opened 2015-10-06
Add db type datetime2

stevewhite:
DateTime2 was in the SqlServer2008 Map but not available as an option when altering or creating columns for a table.


janv8000:
Workaround: see issue #637

fubar-coder:
Only commit 9ffa3fe9bcbe9bd48c1673703da56a9b61335de2 is a bug fix and a candidate for 1.6.2. The DateTime2 is SQL server specific and I guess that it shouldn't be part of the base interface (`IColumnTypeSyntax`). Instead, it should be an SQL Server specific extension.

fubar-coder:

Issue #658, opened 2015-10-07
Support SQLAnywhere


eloekset:
I've started using FluentMigrator against SQLAnywhere 16 and now I've got an implementation that works for all my currently existing scripts.

Will a pull request be accepted? The fork is eloekset/fluentmigrator and SQLAnywhere implementation is done in a branch called dev-SqlAnywhere.

fubar-coder:
Fixed in release/2.0.0

PR #659, opened 2015-10-09
Support SQLAnywhere

eloekset:
First implementation of SQLAnywhere 16 support as requested in #658.


fubar-coder:
There are some changes to the base classes/interfaces and I'm not sure if an alternative take would be better.

1. Move changes to `MigrationRunner` (and probably other base classes) into a separate pull request, because those changes seem to be fixes?
2. Don't enhance the base classes for SQL Anywhere specific features. Instead, try to put everything in database specific classes (e.g. for `CreateIndex` with distinct null value) with the help of extension methods.
3. Schema creation with password should IMHO be more "fluent" like ' Create.Schema("schema").WithPassword("password")` The changes to this base classes should be in a separate pull request.

It would also be nice to have a documentation about the changes to the base syntax, because that would allow us to add the new feature description to the Wiki.

@tommarien Do you have something to add? (Maybe I'm wrong about the second point).

eloekset:
This PR has become quite large, so I'll try to separate out some of the commits like you say @fubar-coder. It's my first large PR, so maybe I need some help as I'm not that experienced to Git yet. I guess that I should: 

1. Create new branches from master and redo all the changes to other files than SqlAnywhereProcessor etc in logical PRs by feature.
2. Have the small PRs merged into trunk.
3. Rebase the branch for this PR onto master once the smaller PRs are merged.
4. Fix conflicts or redo individual commits somehow. Maybe I have to close this PR and create a new branch and PR or use merge instead of rebase? Anyway I need to keep the current changes as is until the cleanup work is done, since we now use FluentMigrator with SQLAnywhere in production.

eloekset:
@fubar-coder My first rebase onto schambers/master seems to have reapplied all my own commits as well. If this PR is going to be closed anyway, I don't need to remove the duplicate commits.

On the good side, #710 is my first PR to move a commit out of this large PR.

eloekset:
@fubar-coder @tommarien Do we have an example of how other provider specific stuff is implemented without enhancing the base class IndexDefinition that @fubar-coder wants to avoid?

I've done a hard reset on my local dev-SqlAnywhere branch and cherry picked all commits until the cc97bbb commit, so I've now got a much cleaner branch. I don't want to sync it to the remote branch and modify this PR yet, so I've published it to a new branch https://github.com/eloekset/fluentmigrator/commits/dev-CleanSqlAnywhereSupport for reference.

Since C# doesn't support extension properties (yet), I don't know how to implement support for an IndexDefinition.IsNullDistinct property without modifying the IndexDefinition class. Any suggestion on how this can be done? If not, is it really a problem that the property is added to the IndexDefinition class, as the GenericGenerator simply generates an error message using the CompatabilityMode.HandleCompatabilty pattern if the nullable property is set to either true or false for generators not supporting this feature?

fubar-coder:
@eloekset I'll have to take a look first, but not today anymore. It might be necessary to modify some base classes to achieve this level of extensibility.

eloekset:
@fubar-coder I've found that SQL Server 2008 supports unique indexes with multiple null values as well. It's called filtered index. I moved the cc97bbb commit into pull request #716 and implemented support in the SqlServer2008Generator. Please consider merging PRs #701, #710 and #716 before this one, and I'll do more cleanup to this PR after rebasing on master.

fubar-coder:

Issue #617, opened 2015-06-11
SqlServerCe: Specifies NVARCHAR(16) as typemap for time


tommarien:
According to me this should be DATETIME

tommarien:
Closed by merge

PR #663, opened 2015-10-18
Fixed SqlServerCe typemap

nakataaaa:
SqlServerCe: Specifies NVARCHAR(16) as typemap for time #617


tommarien:
@nakataaaa Thanks for the fix

PR #665, opened 2015-10-20
[T4]Missing quotes on AsCustom

spaccabit:


tommarien:
@spaccabit Thanks for the fix ;)

PR #672, opened 2015-11-16
Insert and Update support DBNull.Value in criteria expression.

eloekset:
Since a where clause cannot use anonymous types with properties set to null, support for DBNull.Value is implemented.

This expression is invalid:
`.Where(new { StatusId = null })`

This expression is valid:
`.Where(new { StatusId = DBNull.Value })`

Support is added for all Generator types in master branch and new unit tests for all Generator types has been created to verify that both delete and update expressions support DBNull.Value.


tommarien:
@eloekset Hi

I am looking for a usecase of this? what about
.Where(new { StatusId = (int)null })

eloekset:
I didn't think of that option. I tested it now and can confirm that it works fine. 

Still, if you want to support DBNull.Value as well, this pull request can be merged. I think DBNull.Value is a common way to describe the DB null value in .NET, so I was surprised to find it wasn't supported in Where expressions.

fubar-coder:

PR #679, opened 2015-11-21
Invalid syntax error if description contains quote

vilinski:
See PR https://github.com/logary/logary/pull/91

SqlServer description generator is not allowed to have quotes in column or table description.


barser:
Please see issue #650 - same problem for Oracle

vilinski:
Oracle problem is already fixed in [9047d8c](https://github.com/barser/fluentmigrator/commit/9047d8c78be1e1d66709728494e198227ccd082e)

Issue #618, opened 2015-06-16
SqlServerCe: Typemap is wrong


tommarien:
see https://technet.microsoft.com/en-us/library/ms172424%28v=sql.110%29.aspx

* VARCHAR AND CHAR are not supported only NVARCHAR AND NCHAR
* TEXT is not support, NTEXT is up until 536,870,911 characters
* VARBINARY(max) is not supported

eloekset:
I've sent pull request #681

fubar-coder:
Resolved

PR #681, opened 2015-11-29
Fixed issue #618 SqlServerCe Typemap.

eloekset:


eloekset:
I haven't used SqlServerCe myself, but I assume the test in commit 3f85827 was incorrect, so I renamed it from CanCreateTableWithNvarcharMax to CanCreateTableWithNtextSizeUpTo536870911 as NVARCHAR(MAX) is not supported according to the documentation: https://technet.microsoft.com/en-us/library/ms172424%28v=sql.110%29.aspx

Can someone please verify the correctness of this change before merging?

PR #683, opened 2015-12-04
dev-574-OverloadIfDatabaseToAcceptPredicate

### LINKED ISSUE #574: expose the current database type to Up/Down
Would it be possible to expose the current database type to the migration? This would allow me to write logic in C# to perform migrations in different ways depending on the database I was using. IfDatabase is OK but requires a lot of duplicate code if more than two databases are used. For example many parts of my migrations may be the same for postgres and sqlite but would be different only for oracle. Having something like an ElseDatabase would have been nice but it would have been really nice and simple if I could just write `if(DatabaseType=="oracle"){}else{}` . My tale does have a happy ending though as we just ended up moving to postgres finally, so maybe this is all unneeded, but exposing just a string would have helped a lot during that transition and with tests.

RobPethick:
This should fix #574 but I haven't implemented the Enum argument type. I will look into that later, but looked like a more complex change


fubar-coder:

PR #685, opened 2015-12-07
Removed methods from AutoReversingMigration that can create non-autor…

### LINKED ISSUE #439: Migrate allows non-reversable auto-reversable migrations
Using auto-migrate is a very handy tool, however the auto-reversing migration class still allows many non-reversible methods. Ideally these should not be included in the class, or should throw errors when migrating, or at least warnings in the code, but they do not.

Rather what happens is I rollback a change and I get an error then. I should not be allowed to migrate something that does not have a down method IMO

SNeugber:
…eversible migrations.

Fixes #439 .

In response to the ticket: I think it is worth actually defining which migrations are possible and which aren't. Why not remove the ability to even make the mistake by disallowing the use of the 'Create' etc. functions?


fubar-coder:

PR #488, opened 2014-02-27
Added ability to pass VersionMetaData directly to MigrationRunner

ioleksiy:
Added ability to pass VersionMetaData directly to MigrationRunner, rather than reflection search.

PR #692, opened 2015-12-18
Add optional constructor VersionTableMetaData parameter to MigrationRunner and VersionLoader

fratle:
Add optional constructor VersionTableMetaData parameter to MigrationRunner and VersionLoader.
This enables runtime configuration of VersionTableMetaData.

This change is based on @ioleksiy change #488.
#### Use case

We are developing a multitenant application with one schema per tenant and one common schema that is shared between all tenants. We therefore need one VersionInfo table per schema, the existing solution does not support this. 
Currently fluentmigrator only supports one VersionTable (https://github.com/schambers/fluentmigrator/wiki/Create-custom-metadata-for-the-VersionInfo-table).

This change enables us to dynamically alter the VersionTableMetaData in runtime.

Example usage:

``` csharp
private static string _schema;

[VersionTableMetaData]
public class VersionTable : DefaultVersionTableMetaData
{
    public override string SchemaName => _schema;
}

public void MigrateToLatest(string connectionString, string schema) {
    _schema = schema;

    using (var processor = _factory.Create(connectionString, _announcer, _options))
    {
        IVersionTableMetaData versionMetadata = new VersionTable { };
        var runner = new MigrationRunner(_assembly, migrationContext, processor, versionMetadata);
        runner.MigrateUp(true);
    }
}
```
#### Related issues

Added ability to pass VersionMetaData directly to MigrationRunner #488
Enable FluentMigrator extensibility with new IRunnerFactory #287

### DIFF
--- a/src/FluentMigrator.Runner/ConnectionlessVersionLoader.cs
+++ b/src/FluentMigrator.Runner/ConnectionlessVersionLoader.cs
@@ -13,7 +13,7 @@ namespace FluentMigrator.Runner
     {
         private bool _versionsLoaded;
 
-        public ConnectionlessVersionLoader(IMigrationRunner runner, IAssemblyCollection assemblies, IMigrationConventions conventions, long startVersion, long targetVersion)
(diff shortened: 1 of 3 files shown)


fubar-coder:
Overloads for the `MigrationRunner` are needed to avoid breaking existing code.

fubar-coder:

PR #695, opened 2016-01-21
Update column data with other column in the same table

RenaldSB:
Ability to update data of a column with data from another column in the same table.
Sample of line of code:
Update.Column("NAME").From("FIRST_NAME").OnTable("USER");
SQL generated:
UPDATE USER SET NAME = FIRST_NAME WHERE 1=1

### DIFF
--- a/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
+++ b/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
@@ -296,9 +296,16 @@ namespace FluentMigrator.Runner.Generators.Generic
             List<string> updateItems = new List<string>();
             List<string> whereClauses = new List<string>();
 
-            foreach (var item in expression.Set)
+            if (expression.IsFromColumn)
             {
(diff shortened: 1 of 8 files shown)


fubar-coder:
I think that this use case is too special and should be instead handled by a db-specific SQL script.

eliezeralmeida:
Think of a table structure:
`table example(id int not null, amount decimal(10,4) not null, total decimal(10,2) not null)`

I now apply the following change:
`Alter.Table("example").AddColumn("unitary_value").AsDecimal(10,4).SetExistingRowsTo(0).NotNullable();`

and soon after I would like to update all my lines, doing a math operation taking advantage of the columns of the table:
`Update.Table("example").Set(new { unitary_value="total/amount"}).AllRows();`

today is it possible something like that? or do I need to use SQL?

fubar-coder:
You'd have to use SQL. It sounds like a computed column would be a better fit for this kind of problem, but this isn't supported by FluentMigrator either, because it depends on the database if/how it can be used via SQL.

The reason why I closed this issue is mainly, that FluentMigrator is a migration framework having its main focus on DDL statements while this PR adds more DML features.

A better solution would be the development of a separate NuGet package that implements this feature using extension methods.

Issue #700, opened 2016-02-06
TaskExecutor HasMigrationsToApply method


eloekset:
When migrations are run automatically on startup of an application, it should be possible to make a database backup before TaskExecutor.Execute() is called. However the backup is not necessary unless any migrations will actually be executed. 

This functionality could be implemented in a subclass of TaskExecutor, but to make it easier to decide whether or not to do a backup, the TaskExecutor can instead provide a public method HasMigrationsToApply() that any custom runner can call.

```
if (executor.HasMigrationsToApply())
    BackupDatabase();

executor.Execute();
```

fubar-coder:
Fixed by #701

PR #701, opened 2016-02-06
TaskExecutor got method HasMigrationsToApply

eloekset:
…to check whether migrations will run before calling Execute.

Fixes issue #700

### DIFF
--- a/src/FluentMigrator.Runner/IMigrationRunner.cs
+++ b/src/FluentMigrator.Runner/IMigrationRunner.cs
@@ -19,5 +19,8 @@ namespace FluentMigrator.Runner
         void MigrateDown(long version);
         void ValidateVersionOrder();
         void ListMigrations();
+        bool HasMigrationsToApplyUp(long? version = null);
+        bool HasMigrationsToApplyDown(long version);
+        bool HasMigrationsToApplyRollback();
     }
 }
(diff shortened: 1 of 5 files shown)


eloekset:
Before this pull request I created a custom TaskExecutor that extends the default with a GetMigrationsCount() method.

```
public int GetMigrationCount()
{
    this.Initialize();
    try
    {
        var versionLoader = new VersionLoader(this.Runner, this.Runner.MigrationAssemblies, ((MigrationRunner)this.Runner).Conventions);
        versionLoader.LoadVersionInfo();

        var migrationsToRun = from m in this.Runner.MigrationLoader.LoadMigrations()
                              where !versionLoader.VersionInfo.HasAppliedMigration(m.Key)
                              select m;

        return migrationsToRun.Count();
    }
    finally
    {
        DisposeProcessor();
    }
}
```
This implementation assumes the IMigrationRunner is the default MigraitonRunner implementation, and that the runner wants to execute all migrations up. It doesn't support down, rollback or up to a specific version. I also struggled with a connection leak, and that's why I made a DisposeProcessor() method in this commit (in another pull request): https://github.com/eloekset/fluentmigrator/commit/836ba304c53c2df688d9e08bfa47d2511e79fe0a

This is no longer necessary if this pull request is merged. And the implementation is also much better than the code I've shown above.

fubar-coder:

PR #716, opened 2016-04-13
Unique index with support for multiple null values

eloekset:
Related to PR #659 unique indexes that allow multiple rows with null values is a feature to support for SQL Anywhere databases. After further investigation, I found that a similar unique index is supported by SQL Server 2008 and up as well: http://stackoverflow.com/questions/767657/how-do-i-create-a-unique-constraint-that-also-allows-nulls

This PR replaces commit cc97bbb in #659 and additionally adds support for SQL Server 2008 and up.

### DIFF
--- a/src/FluentMigrator.Runner/Generators/Firebird/FirebirdGenerator.cs
+++ b/src/FluentMigrator.Runner/Generators/Firebird/FirebirdGenerator.cs
@@ -87,7 +87,8 @@ namespace FluentMigrator.Runner.Generators.Firebird
                 , indexDirection == Direction.Ascending ? "ASC " : "DESC "
                 , Quoter.QuoteIndexName(expression.Index.Name)
                 , Quoter.QuoteTableName(expression.Index.TableName)
-                , indexColumns.ToString());
+                , indexColumns.ToString()
+                , GetWithNullsDistinctString(expression.Index));
         }
 
         public override string Generate(AlterColumnExpression expression)
--- a/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
+++ b/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
@@ -33,7 +33,7 @@ namespace FluentMigrator.Runner.Generators.Generic
         public virtual string AlterSchema { get { return "ALTER SCHEMA {0} TRANSFER {1}.{2}"; } }
         public virtual string DropSchema { get { return "DROP SCHEMA {0}"; } }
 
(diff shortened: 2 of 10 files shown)


eloekset:
The fluent syntax that I suggest in this PR is:
```
Create.Index("IX_TestTable_SomeUniqueColumn")
    .OnTable("TestTable")
    .OnColumn("SomeUniqueColumn")
    .Ascending()
    .WithOptions()
    .Unique().WithNullsNotDistinct();
```

tommarien:
Should this not be a choice that is defined per column in your index composition ?

eloekset:
SQL Anywhere doesn't support ignoring NULL values on a per column basis in unique indexes. But I see your point when it comes to SQL Server. In fact the SQL Server syntax is a WHERE condition, which can be very flexible.

I tested a more complex unique index by creating an index like this:
```
CREATE UNIQUE INDEX IX_Test2_UniqueColumn ON mts.Test2 (StupidUniqueColumn ASC) 
WHERE (StupidUniqueColumn IS NOT NULL) AND (StupidUniqueColumn IN ('a', 'b', 'c'))
```

I can now insert multiple rows with NULL values and multiple with 'dc' values, but only one with 'a', 'b' or 'c' values:


Actually I first tried creating an index with this criteria:
```
CREATE UNIQUE INDEX IX_Test2_UniqueColumn ON mts.Test2 (StupidUniqueColumn ASC) 
WHERE (StupidUniqueColumn IS NOT NULL) AND (StupidUniqueColumn NOT IN ('a', 'b', 'c'))
```

The only difference here is that I want to allow multiple rows with 'a', 'b' and 'c' values and have all other values unique except NULLs. But for some reason NOT isn't supported in SQL Server 2014:


If we extend the FluentMigrator Create.Index() syntax to support unique indexes ignoring NULL values, we can still extend the Create.Index() syntax for even more complex indexes in the future.

eloekset:
I'll try an alternative approach some time next week. The syntax can be changed to... 
```
Create.Index("IX_TestTable_SomeUniqueColumn")
    .OnTable("TestTable")
    .OnColumn("SomeUniqueColumn")
    .Ascending().UniqueWithNullsNotDistinct()
    .WithOptions()
    .Unique();
```
...without adding all the complexity that SQL Server 2008 can possibly support. I can add another pull request for that approach, so they can be compared before deciding which one to go for.

eloekset:
Close this PR if #717 is considered a better approach. #717 is more aligned to SQL Server's implementation.

fubar-coder:

PR #717, opened 2016-04-18
Unique index with support for multiple null values per column

eloekset:
Related to PR #659 unique indexes that allow multiple rows with null values is a feature to support for SQL Anywhere databases. After further investigation, I found that a similar unique index is supported by SQL Server 2008 and up as well: http://stackoverflow.com/questions/767657/how-do-i-create-a-unique-constraint-that-also-allows-nulls

This PR replaces commit  cc97bbb  in #659 and additionally adds support for SQL Server 2008 and up.

The difference between this and PR #716 is that this PR supports setting the `WithNullsNotDistinct` option on a per-column basis, which is possible in SQL Server 2008 and up.

### DIFF
--- a/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
+++ b/src/FluentMigrator.Runner/Generators/Generic/GenericGenerator.cs
@@ -54,6 +54,20 @@ namespace FluentMigrator.Runner.Generators.Generic
             return string.Empty;
         }
 
+        public virtual string GetWithNullsDistinctString(IndexDefinition index)
+        {
+            if (index.Columns.Where(c => c.IsNullDistinct.HasValue).Any() && !index.IsUnique)
+            {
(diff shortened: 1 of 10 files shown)


eloekset:
The fluent syntax that I suggest in this PR is:
```
Create.Index("IX_TestTable_SomeUniqueColumn")
    .OnTable("TestTable")
    .OnColumn("SomeUniqueColumn")
    .Ascending().UniqueWithNullsNotDistinct()
    .WithOptions()
    .Unique();
```
This is in response to @tommarien's comment in PR #716.

fubar-coder:
This PR is now part of the release/2.0.0 branch. Thank you for your hard work 👍

fwd079:
Hi, 
Nice that'd be great, when is going to be released please? Currently it gives me this:

image

Regards.

fubar-coder:
It's already included. You have to reference the NuGet package `FluentMigrator.Extensions.SqlAnywhere` to be able to use this extension method.

Issue #709, opened 2016-03-24
Migrate.exe could be more helpful in argument parsing error situations..


untilted-document:
I'm struggling to use Migrate.exe. I wrote a command like this, following the tool's output help:

`Migrate.exe -a "PathToMyMigrations.dll" -dbtype SqlServer -conn "MyNiceConnectionStringHere"`

And I run it - Migrate just prints its arguments again and exits with code 1. What was wrong? Why didn't it tell me what it didn't like? Here's a go with definitely broken args (fake assembly and no "conn"):

![image](https://cloud.githubusercontent.com/assets/18049243/14015844/adaa6678-f1b3-11e5-95b0-4724b33257ff.png)

A message like "--conn is required" or "Assembly 'a' could not be loaded" would be good. That's if conn is indeed required - wiki says it is, console help is inconclusive

I actually cribbed my first stab from a blog:

`Migrate.exe --assembly="PathToMyMigrations.dll" --dbtype=SqlServer --conn="MyNiceConnectionStringHere"`

Which led me to have to read a lengthy wiki article on the arguments format, and why one example looked completely different to the other. I'm now educated (not that it helps solve my original problem) but feel the way you do your arguments help text is a little confusing. I appreciate that you're trying to offer maximum flexibility with them but this just reads like the author is deranged:

> "If a parameter has an = sign after it, then it expects some text value to be set ... [and] ... the = sign is optional ... Single dashes or double dashes is also optional
> --verbose=VALUE ... This should be a boolean switch but is not. VALUE can be anything to set to true and the switch should not be specified to set it to false"

So --verbose=false is verbose? What if I say -verbose? Is it true or not, because it's present but doesnt have a value. Erm.

Examples that present --conn and -c like the number of hyphens is significant are also misleading? When a super-duper easy-to-use-because-it-can-make-sense-of-any-old-garbage-passed-in algorithm starts needing a convoluted explanation, it's not really achieving its design goal any more

Do you think, for the sake of consistency, you could please settle on one format, make all your help texts show examples in that format and then discreetly stop talking about (even if you still support for legacy reasons) the argument formats of having space/equals or not, values or not.. (And also tell me what it doesn't like about the arguments I'm passing)

chairman-mawh:
Ah.. I've just worked it out. I was passing "dbtype" (**just like the wiki says I can**) and it was barfing (silently) - quick change that t -> T and it all starts working.

[...]

PR #719, opened 2016-04-30
709 case insensitive arguments and give more instructive messages when required options are not given

RobPethick:
Doesn't completely solve #709 but certainly would help to know which option is not getting picked up


RobPethick:
Should help with #709 again re:case insensitive option names

fubar-coder:

Issue #712, opened 2016-03-30
Support for .NET Core


fubar-coder:
We should think about starting to support .NET Core as soon as RC2 is out and supported by VS 2015. With the change to a `project.json`, we're still able to target platforms like .NET 3.5/4 in addition to netstandard1.3 (or higher?).
- [.NET Core project system](https://github.com/dotnet/core-docs/blob/master/docs/project-model/index.md)
- [.NET Platform Standard](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md)
- [Target frameworks](https://docs.nuget.org/Create/TargetFrameworks)
- What is [netstandardapp](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-standard-applications.md)?

Conversion steps:
- [ ] Move from `IDbConnection`, etc to `DbConnection`
  - `System.Data.Common` is available since .NET 2.0
- [ ] Remove the need for ICloneable
  Switching to immutable types?
- [ ] Replace usage of DataSet
  - Using the solution from @dmirmilshteyn (https://github.com/schambers/fluentmigrator/issues/621#issuecomment-154342973)?
  - Use a DataReader instead?

All this work shouldn't be done in a single step. It should be done in multiple pull requests (or a single pull request with squashed commits to have one commit per step) for easier code review.

eloekset:
#621 is also about this topic.

fubar-coder:
We're tracking the progress here.

dmirmilshteyn:
@fubar-coder Another factor to consider is integration with the dotnet CLI. It would be nice to be able to add a runner to the "tools" section of project.json, then execute commands such as "dotnet fm migrate [flags]". It would likely be as simple as renaming the current console runner to dotnet-fm and then packing it up in nuget. In the short term, to avoid breaking backwards compatibility, we can create a new runner project which references the console runner, and have the input forwarded to that until the next major release.

I've started working on a port with the RC2 bits, will have some PRs in shortly...

giggio:
Just FYI, I got my runner for .NET CLI for Fluent Migrator working. The license is compatible, you may want to take a look. I think that supporting .NET Core is very important, I am just posting this to help people who would like to use Fluent Migrator until this step is completed.
https://github.com/giggio/FluentMigrator.Runner.Cli

jayalakshmis:
@giggio, do you by chance have a pull request out for merging this upstream?

giggio:
[...]

PR #721, opened 2016-05-29
Switch from DataSet to IDataReader

dmirmilshteyn:
This replaces usage of DataSet with IDataReader in the core library and associated runners/tests. This is initial work to add support for .NET Core RC2, related to #712

### DIFF
--- a/src/FluentMigrator.Runner/Processors/Firebird/FirebirdProcessedExpression.cs
+++ b/src/FluentMigrator.Runner/Processors/Firebird/FirebirdProcessedExpression.cs
@@ -125,19 +125,19 @@ namespace FluentMigrator.Runner.Processors.Firebird
             CanUndo = true;
             FirebirdSchemaProvider schema = new FirebirdSchemaProvider(Processor);
             FirebirdTableSchema table = schema.GetTableSchema(expression.TableName);
-            using (DataSet ds = Processor.ReadTableData(String.Empty, expression.TableName))
+            using (IDataReader ds = Processor.ReadTableData(String.Empty, expression.TableName))
             {
                 foreach (DeletionDataDefinition deletion in expression.Rows)
                 {
                     InsertDataExpression insert = new InsertDataExpression() { SchemaName = String.Empty, TableName = expression.TableName };
-                    foreach (DataRow dr in ds.Tables[0].Rows)
+                    while (ds.Read())
                     {
                         bool match = true;
                         if (!expression.IsAllRows)
                         {
                             foreach (var where in deletion)
                             {
-                                if (dr[where.Key].ToString() != where.Value.ToString())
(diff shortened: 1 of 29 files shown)


src/FluentMigrator.Runner/Processors/Firebird/FirebirdProcessedExpression.cs
```diff
@@ -125,19 +125,19 @@ protected void SetupUndoDeleteData(DeleteDataExpression expression)
             CanUndo = true;
             FirebirdSchemaProvider schema = new FirebirdSchemaProvider(Processor);
             FirebirdTableSchema table = schema.GetTableSchema(expression.TableName);
-            using (DataSet ds = Processor.ReadTableData(String.Empty, expression.TableName))
+            using (IDataReader ds = Processor.ReadTableData(String.Empty, expression.TableName))
```
DustinVenegas: This variable should be renamed to `dr` instead of `ds`.

src/FluentMigrator/IMigrationProcessor.cs
```diff
@@ -29,8 +29,8 @@ public interface IMigrationProcessor : IQuerySchema, IDisposable
         string ConnectionString { get; }
 
         void Execute(string template, params object[] args);
-        DataSet ReadTableData(string schemaName, string tableName);
-        DataSet Read(string template, params object[] args);
+        IDataReader ReadTableData(string schemaName, string tableName);
+        IDataReader Read(string template, params object[] args);
```
DustinVenegas: I don't know how strict we are. Do we need to rev a major version for this change?

marcrocny: Is breaking for strict semver. There's no references to such a policy anywhere I could find.  But I would say that OSS should default to strict semver.  This is a breaking change to a public interface, and interface stability is paramount.  It may not be in common use, but this also presents a breaking change with the separately-delivered FluentMigrator.Runner package, and version-disagreement breakage there should be well-advertised.

* is the broken API in common use externally?  -- would require polling to determine, but I suspect not.
* internally, does it present a breaking change to older versions of other, separately released packages that

marcrocny: Is breaking for strict semver. There's no references to such a policy anywhere I could find.  But I would say that OSS should default to strict semver.  This is a breaking change to a public interface, and interface stability is paramount.  It may not be in common use, but this also presents a breaking change with the separately-delivered FluentMigrator.Runner package, and version-disagreement breakage there should be well-advertised.

dmirmilshteyn: @marcrocny At this point, I don't think this PR is relevant anymore. Netstandard2.0 reintroduces DataSet, so there's no real reason to go with this.

fubar-coder:

PR #725, opened 2016-06-10
AsString() -> Nvarchar() in MySql

hollisyao:
If you use MySql database, there is a error that AsString method generates varchar data type.
So, i fix it that AsString method generates nvarchar data type.


DustinVenegas:
Interesting catch @Hollisyao. It seems this project is a bit behind in PRs. Before anyone gets to this one you'll have to update the tests to pass as well. 

It looks like this is going to be a breaking change for MySQL if accepted.

fubar-coder:

Issue #727, opened 2016-06-27
Output file should not have the command preceded by itself within comments ("ExecuteSqlStatement ...")


arfmach:
If we Execute.Sql a query that has a "GO" command (example: dropping/recreating a stored procedure) and `/* comments */` in it, output file repeats the query within a comment, producing an invalid SQL file, as confirmed in [this article](http://blog.sqlauthority.com/2007/08/03/sql-server-two-different-ways-to-comment-code-explanation-and-example/), under topic “_Additional Notes_”: “**GO Command is not allowed anywhere in commenting text**”.

► Please add an option to suppress echoing commented command to output file.

![go-within-comments](https://cloud.githubusercontent.com/assets/7331118/16387464/6054887e-3c6b-11e6-9927-94868129a69f.png)

arfmach:
Actually, it should be an option to **enable** echoing commented command to output file, with its default value as *false*.

arfmach:
I've branched the repo and made the change; just waiting for being added as a contributor in order to publish the branch and create the pull request.

eloekset:
If you fork the repo, you can send a pull request without being assigned contributor access to the main repo. See one of the existing pull requests, they have a link to the forked repo.

fubar-coder:
Filtering out comments might not an easy task, because some kind of interpretation might be necessary. The easiest solution is to avoid problematic text in comments.

PR #728, opened 2016-06-30
Execute.Sql option to enable echoing command to output file

arfmach:
[ Issue #727 ] Added an option to Execute.Sql() to enable/disable (default=false) echoing command (comment) to output file, because command may have comments and it sometimes messes with the execution of the file.

P.S. I don't know why GitHub file comparison shows all lines being deleted and added again; diff should show only changed lines.


DustinVenegas:
Hey @arfmach the files are showing up as all changed because your git line endings are off. You should make sure you have a recent git client and that `git config core.autocrlf true` is set for this repository. 

Can you either rebase/squash or recreate this branch so the line endings are correct?

PR #739, opened 2016-10-09
NuGet-ize Test Framework Packages

DustinVenegas:
This pull request adds NuGet to the repository so packages are downloaded to ./lib/nuget. NUnit and Moq specifically use NuGet packages in preparation for a .NET Core migration. 

_Note:_ The Albacore build was updated, but not tested due to the Ruby requirement. I'm assuming TeamCity will catch anything wrong there.


DustinVenegas:
I'm not seeing the cause for the Moq reference build break. NuGet appears to work fine and I don't see errors related to NUnit. Ideas?

fubar-coder:
Mostly done in the release/2.0.0 branch

PR #740, opened 2016-10-09
WIP - Convert NUnit to Xunit

DustinVenegas:
This pull request is a work-in-progress for migrating NUnit to a .NET 3.5 compatible Xunit. 

To my knowledge NUnit isn't supported in .NET Core yet and Xunit has excellent tooling there. Logic is not expected to be changed for anything. 

**Outstanding Question**: Do we want to use `Assert.<func>(expected, actual);` calls or look for an equivalent Xunit should package? 

Please shoot feedback early since this is going to touch pretty much every file in a test project.


fubar-coder:
NUnit now supports .NET Core, so there is currently no need to change the test runner.

PR #742, opened 2016-10-14
Execute.Sql option to enable echoing command; changed timestamps.

arfmach:


fubar-coder:
Not feasible for inclusion into the standard project. I'd suggest to resolve this problem by developing your own announcer.

PR #750, opened 2016-11-29
ListMigrations says "(not applied)" for migrations not applied.

### LINKED ISSUE #749: Listmigrations lists all migrations but does not show which has been applied in previous then current version
I can't find the documentation that specifies how this should exactly work and maybe this is more of a feature request. Currently it lists all Migrations and which is the current migration you're on. To me I would also think that all the listed migrations of previous then current version had been applied but that is not the case.

My request is that listmigrations would show migrations not applied with a [ - ] instead of [ + ] infront. Those migrations not implemented that is of previous version then current should perhaps also get a red or orange color so it's clear it hasn't been applied.

Example:
If two people work on same migration project, Foo checks in first, with migration version 3, and Bar checks in last, with migration version 2. Foo has already  run migration 1 and 3 then pulls down the latest code before he builds the project. He then runs <Migration.exe> -task listmigrations to see what migrations he's missing/version he's on. It will show him all 3 migrations and that he's on the latests (current) while which is true of the point is only to show in what version he's on. List migrations would give a lot more if it also showed him that version 2 hadn't been run.

Example 2:
(Those in red circle is migrations missing and hasn't been applied)
image
image

eloekset:
Fixes #749 

Announcers don't have a way to say the [-] sign requested, so I suggest saying "(not applied)" instead.

### DIFF
--- a/src/FluentMigrator.Runner/MigrationRunner.cs
+++ b/src/FluentMigrator.Runner/MigrationRunner.cs
@@ -537,9 +537,10 @@ namespace FluentMigrator.Runner
             {
                 string migrationName = migration.Value.GetName();
                 bool isCurrent = migration.Key == currentVersion;
+                bool isApplied = VersionLoader.VersionInfo.HasAppliedMigration(migration.Value.Version);
                 string message = string.Format("{0}{1}",
                                                 migrationName,
-                                                isCurrent ? " (current)" : string.Empty);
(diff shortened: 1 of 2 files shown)


ArneBK:
Looking good! I think this is a nice feature that listmigrations is missing!

fubar-coder:

PR #751, opened 2016-11-30
Modified IntegrationTestBase so it Assert failure when no database se…

ericburcham:
**As discussed with Tom on 11/28/2016**:

Modified IntegrationTestBase so it Assert failure when no database server options are configured.  This prevents false positives when running tests using the ExecuteWithSupportedProcessors method.


ericburcham:
The Travis-CI build failed on my first commit because it doesn't use support string interpolation.  I just replaced that with Assert.Fail's support for the old string.Format behavior with indexed {0} replacement tokens.

ericburcham:
In my last commit, I removed the usage of string interpolation, but left the nameof keyword.  Because I'm _that smart_.  I have removed the nameof keyword now, and the Travis-CI build is passing.

fubar-coder:

PR #753, opened 2016-12-05
use unique constraint instead of unique index in ColumnExpressionBuilderHelper.Unique

jcberleur:
This fixes the convention used `GetConstraintName` instead of `GetIndexName`.

Unfortunately, this might be a breaking change for those that use the Index directly after (for Delete/Rename/etc.)

### DIFF
--- a/src/FluentMigrator.Tests/Integration/MigrationRunnerTests.cs
+++ b/src/FluentMigrator.Tests/Integration/MigrationRunnerTests.cs
@@ -162,6 +162,24 @@ namespace FluentMigrator.Tests.Integration
         }
 
         [Test]
+        public void CanApplyUniqueConvention()
+        {
+            ExecuteWithSupportedProcessors(
+                processor =>
+                {
+                    var runner = new MigrationRunner(Assembly.GetExecutingAssembly(), _runnerContext, processor);
+
+                    runner.Up(new TestUniqueConstraintNamingConvention());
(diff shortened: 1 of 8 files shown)


fubar-coder:
I think that this has serious implications for SQLite which doesn't know named constraints. Instead, you can have a UNIQUE constraint on a column, so I guess we'll have to use INDEXes instead. Another problem is, that the UNIQUE column constraint is unnamed and a removal of just this constraint isn't easily done.

One has to create a new TABLE with the modified COLUMN and move all data from the old TABLE to the new one, drop the old table and rename the new one to the name of the old one. This would probably also mean the recreation of INDEXes, TRIGGERs and maybe VIEWs.

And dropping the CONSTRAINT by name still wouldn't be supported by SQLite.

@escarls @eloekset What is your opinion on this topic?

jzabroski:
@jcberleur Just wondering , why did you close this PR, other than the fact it is 2 years old?

jzabroski:
Re-reading the comments, it looks like the reason this wasn't merged was due to SQLite and lack of a "metamodel" for FluentMigrator, see #1032 for a task which attempts to build such a meta model so that composing such complex features doesn't cause unexpected bugs in SQLite and other vanilla sql providers.

jcberleur:
> @jcberleur Just wondering , why did you close this PR, other than the fact it is 2 years old?

that was the only reason

Issue #680, opened 2015-11-24
Logging from within the migration


LSTANCZYK:
I would like to be able to log information from inside of the migration. This should be included in the output log. Is this currently possible and if not, please consider it as a feature request.

ROIInstitute:
I would also like to know this.

sjorsmiltenburg:
I would also like to know this..

eloekset:
I don't think it's possible. I've been using Console.WriteLine() from within my migrations, but that obviously works when run in a console runner only.

sjorsmiltenburg:
what I did was create a custom announcer which writes to log4net.

`public class Log4netAnnouncer : Announcer
	{
		private static readonly log4net.ILog Log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);

		public override void Write(string message, bool escaped)
		{
			Log.Debug(message);
		}

		public override void Error(Exception exception)
		{
			base.Error(exception);
			Log.Error(exception);
		}
	}`

fubar-coder:
I think about using Microsoft.Extensions.Logging. This would allow log message filtering and would at least a partial solution for this problem.

fubar-coder:
Fixed in develop branch

Issue #738, opened 2016-10-09
Single Build System


DustinVenegas:
I was slowly looking at what it would take to port things to .NET Core. While poking around, I noticed there are three different build systems: NAnt, MSBuild, and Rake/Albacore. 

I would like to consolidate the current build systems into CAKE or FAKE. This would assist with contributions (No Ruby on Windows required) and would simplify the build (one place to look). Are there any hard dissents or opinions?

dmirmilshteyn:
I might be wrong, but I think the NAnt portions just provide support for integrating the migration runner into an existing NAnt workflow. I've been able to build the solution using solely MSBuild. In terms of Rake, the amount of work done in the rakefile seems fairly minimal. It may be simpler to just write sh/ps1 scripts instead of bringing in another build system like CAKE, especially since some of the multi-targetting and NuGet management functions are already part of the .NET Core CLI. 

Either way, I'm definitely in favour of simplifying the build system, regardless of how it ends up being done.

DustinVenegas:
@dmirmilshteyn, I believe you're correct for NAnt. Those files probably just need to be moved to reduce confusion. 

Since this project gets compiled cross-platform I figured some meta-runner would be useful. At the least, those options are still fewer dependencies (-Ruby, already depend on .NET). Thoughts?

escarls:
When I look at it, it seems the Rake script is for building the NuGet package and the MSBuild is used by Travis CI.
Not sure what the NAnt script is used for, except it's called from Build.bat.

escarls:
The NAnt script is used by the TeamCity.

escarls:
I think your idea for a single build system is good.
However, I think that to minimize impact of changing, maybe it would be better to go just MSBuild at the start, and then change it to Cake for the next big version change?

fubar-coder:
Solved in the release/2.0.0 branch by using the .NET Core tooling/msbuild

PR #757, opened 2016-12-17
Optional NuGet package for a Log4Net announcer

eloekset:
After writing a Log4Net announcer for a few projects, maybe it's a good idea to build it into an optional NuGet package distributed like the other official FluentMigrator packages.

Hope it provides enough flexibility. I've assumed INFO is the desired log level for typical messages, WARN for emphasized messages and ERROR for error messages.

This is related to issue #680.

### DIFF
--- a/FluentMigrator.Log4NetAnnouncer.Tests/FluentMigrator.Log4NetAnnouncer.Tests.csproj
+++ b/FluentMigrator.Log4NetAnnouncer.Tests/FluentMigrator.Log4NetAnnouncer.Tests.csproj
@@ -1,0 +1,84 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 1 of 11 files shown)


eloekset:
The build fails because there's no x86|Debug configuration for the two new projects. Anyone got VS2010 installed and can fix the VS2010 solution file. I've tried editing it manually.

```/home/travis/build/schambers/fluentmigrator/FluentMigrator (2010).sln:  warning : The project configuration for project 'FluentMigrator.Log4NetAnnouncer' corresponding to the solution configuration 'Debug|x86' was not found.```

fubar-coder:
I don't have access to VS2010 any more. @schambers Can we switch to the VS2015 solution on the build server?

escarls:
I have a VS2010 install. I did fix the solution, but the issue was not that the Debug|x86 was missing.
The issue is that Travis runs on Linux and nuget command line doesn't work good there.
At least not on the old mono version the CI is running on.

So I changed the project to grab the dlls from lib like the other projects and I was able to compile it with xbuild on mono.

Also, it looked like it was possible to run xbuild on the FluentMigrator.sln solution also, so I guess that is something we could try. Should be just to switch it in CI.proj
@fubar-coder Do you know if there will be run ci tests if you just try this out on a branch of this repo?

fubar-coder:
Sorry, I don't have much insight about the configuration or functionality of the currently used CI. Maybe @tommarien or @schambers can answer this question?

escarls:
I've made myself a travis account and testing it in my repo here https://github.com/escarls/fluentmigrator/commits/ci_builds
So far, it goes ok, but I need to get higher on the mono versions before I can get it to use NuGet in the builds. I guess my experimentation relates to this pr, along with some others and also #738.

fubar-coder:
This looks good, but - to be honest - I flirt with the idea of using `Microsoft.Extensions.Logging`. Do you have a problem with dropping support for .NET Framework versions < 4.6.1?

eloekset:
I agree it's better to go through Microsoft.Extensions.Logging and get support for all outputs already implemented for that. So I guess we can just close this PR?

Issue #756, opened 2016-12-08
Support new features in SqlServer and SqlAzure 


jonnybee:
FluentMigrator should support new feature for Index creation such as 
INCLUDE (field1, field2) and WITH should support new options such as ONLINE that will allow creation of index without locking the entire table.

eloekset:
Ref MSDN docs on nonkey columns:
```
-- Creates a nonclustered index on the Person.Address table with four included (nonkey) columns.   
-- index key column is PostalCode and the nonkey columns are  
-- AddressLine1, AddressLine2, City, and StateProvinceID.  
CREATE NONCLUSTERED INDEX IX_Address_PostalCode  
ON Person.Address (PostalCode)  
INCLUDE (AddressLine1, AddressLine2, City, StateProvinceID); 
```
https://msdn.microsoft.com/en-us/library/ms190806.aspx#Anchor_2

eloekset:
I started adding INCLUDE syntax, but then I found this was added for SQL Server 2012 in commit ceca4ea2.

eloekset:
Ref MSDN docs on WITH (ONLINE = ON|OFF):
https://msdn.microsoft.com/en-us/library/ms177442.aspx#Anchor_2

Should be available for CREATE INDEX, ALTER INDEX, DROP INDEX and ALTER TABLE:
https://msdn.microsoft.com/en-us/library/ms177442.aspx#Anchor_0

eloekset:
@jonnybee, do the two PRs referencing this issue contain the features you requested?

jonnybee:
Yes, this would solve our immediate issues. Create index WITH (ONLINE=ON|OFF) was actually first introduced in SQL Server 2005 Enterprise Edition. This is also supported for SqlServer 2008 and 2012 Enterprise Edition .

eloekset:
@jonnybee, can you take a look at PR #788 and comment whether you agree on the implementation? 

I don't think it's possible to test for editions of SQL Server when writing migrations, so the user will have to take the limitions of other editions into account, and avoid using the new option unless target engine is Enterprise Edition.

fubar-coder:
Fixed by PR #788 in release/2.0.0

PR #759, opened 2016-12-18
Syntax for creating indexes with nonkey columns

eloekset:
Related to issue #756 

If someone who knows the history can say why the Include syntax was left out, we probably have a reason to not merge this PR as is. But as far as I can see, there's no reason why the IndexIncludeDefinition created in commit ceca4ea shouldn't be supported.


fubar-coder:

PR #761, opened 2016-12-25
move from IDbConnection to DbConnection

govorovvs:


fubar-coder:

PR #762, opened 2016-12-26
remove System.Runtime.Serialization usage

govorovvs:
Part of #712 . I've removed System.Runtime.Serialization usages as they aren't longer supported in netstandard


fubar-coder:

PR #772, opened 2017-02-03
Ability to specify the Default Schema Name

eveilleux:
Following the work started by @daniellee in 2013, I completed the work to be able to override the default schema globally.  There are 2 ways to do it:
1- Define a MigrationConventions class in the migration assembly, and it will be used to determine the default schema to use.
2- As a parameter in the RunnerContext.

The 1 takes precedence on 2.


fubar-coder:

PR #782, opened 2017-03-06
Map Guid to char(36)

### LINKED ISSUE #769: Column type for Guid
I noticed that with MySQL a varchar(40) column is created for storing Guids. Wouldn't it be more efficient to create a char(36) column?

marcrocny:
The standard string-format Guid is a fixed, 36-character string; fixes #769

### DIFF
--- a/src/FluentMigrator.Runner/Generators/MySql/MySqlTypeMap.cs
+++ b/src/FluentMigrator.Runner/Generators/MySql/MySqlTypeMap.cs
@@ -37,7 +37,7 @@ namespace FluentMigrator.Runner.Generators.MySql
             SetTypeMap(DbType.Decimal, "DECIMAL(19,5)");
             SetTypeMap(DbType.Decimal, "DECIMAL($size,$precision)", DecimalCapacity);
             SetTypeMap(DbType.Double, "DOUBLE");
-            SetTypeMap(DbType.Guid, "VARCHAR(40)");
+            SetTypeMap(DbType.Guid, "CHAR(36)");
             SetTypeMap(DbType.Int16, "SMALLINT");
             SetTypeMap(DbType.Int32, "INTEGER");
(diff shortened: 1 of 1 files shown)


fubar-coder:
I believe that this is a non-breaking change and therefore a candidate for 1.6.3

fubar-coder:
@RenaldSB @reinaldocoelho What do you guys think? Is it ok for 1.6.3 or is it something for 1.7?

marcrocny:
Not a feature-add, so on semver grounds I'd agree that 1.6.3 would be appropriate.

rhegner:
Unfortunately this does seem to be a breaking change. In my database I have tables created with an old version of fluentmigrator (GUID -> VARCHAR(40)) and newer tables with CHAR(36).
With MySQL this does not seem to be a problem, but with MariaDB creating a new foreign key between and old and a new table fails with "Foreign key constraint is incorrectly formed".
I do have this problem with MariaDB on a live system :(

Any idea how to resolve this? Changing the datatype of the existing VARCHAR(40) columns to CHAR(36) does not seem to work because these columns are already part of foreign keys...

fubar-coder:
The only way to solve this problem is by providing your own MySQL generator by copying the code from [src/FluentMigrator.Runner.MySql/Generators/MySql](https://github.com/fluentmigrator/fluentmigrator/tree/master/src/FluentMigrator.Runner.MySql/Generators/MySql) and updating the type map - or by using `IfDatabase("mysql").etc.etc.AsCustom("VARCHAR(40)")`.

rhegner:
In the meantime I dumped all the data of the live database, deleted the schema, let the new version apply all the migrations on a fresh database, and restored the data into the new database.
That was a bit of a pain, but fortunately the live system was not so busy these days.

fubar-coder:
Glad that you could work it out.

PR #783, opened 2017-03-06
Add MySQL alter and drop of defaults

marcrocny:
While MySQL doesn't treat default values as named constraints, there is
a syntacitical facility under ALTER TABLE for manipulating them; see
https://dev.mysql.com/doc/refman/5.7/en/alter-table.html


fubar-coder:
I added this to 1.6.3, because this feature exists since 4.0.22 (2005)

fubar-coder:

PR #787, opened 2017-03-13
Allow before-each & after-each Maintenance only if migrations are run; DRY on MigrateUp.

marcrocny:
Would really like a few sets of eyes on this one, especially original authors.  

This PR consolidates the logic for MigrateUp by (1) pre-filtering the migration set in the "full install" polymorph and (2) adding ApplyMaintence to the version-specific polymorph.

Further, it limits the execution to only run in the case where there are migrations to apply.  *This materially changes the Maintenance Migration logic,* though in a way that seems closer to the original intent--previously, because filtering for "full install" MigrateUp occurred in ApplyMigrationup, all maintenance *for every migration* would run, even if no migrations were actually run.

I also need some guidance on testing this, since the Maintenance feature does not seem very test-friendly.

This is hypothetically breaking, if there are Maintenance Migrations out there that are depending on the "always-run" "feature."  But it is certainly at least a minor-version change.

### DIFF
--- a/src/FluentMigrator.Runner/MigrationRunner.cs
+++ b/src/FluentMigrator.Runner/MigrationRunner.cs
@@ -126,39 +126,7 @@ namespace FluentMigrator.Runner
 
         public void MigrateUp(bool useAutomaticTransactionManagement)
         {
-            var migrations = MigrationLoader.LoadMigrations();
-
-            using (IMigrationScope scope = _migrationScopeHandler.CreateOrWrapMigrationScope(useAutomaticTransactionManagement && TransactionPerSession))
-            {
-                try
-                {
-                    ApplyMaintenance(MigrationStage.BeforeAll, useAutomaticTransactionManagement);
-
-                    foreach (var pair in migrations)
-                    {
-                        ApplyMaintenance(MigrationStage.BeforeEach, useAutomaticTransactionManagement);
-                        ApplyMigrationUp(pair.Value, useAutomaticTransactionManagement && pair.Value.TransactionBehavior == TransactionBehavior.Default);
-                        ApplyMaintenance(MigrationStage.AfterEach, useAutomaticTransactionManagement);
-                    }
-
-                    ApplyMaintenance(MigrationStage.BeforeProfiles, useAutomaticTransactionManagement);
-
-                    ApplyProfiles();
-
-                    ApplyMaintenance(MigrationStage.AfterAll, useAutomaticTransactionManagement);
-
-                    scope.Complete();
-                }
-                catch
-                {
-                    if (scope.IsActive)
-                        scope.Cancel();  // SQLAnywhere needs explicit call to rollback transaction
-
-                    throw;
-                }
-            }
-
-            VersionLoader.LoadVersionInfo();
+            MigrateUp(long.MaxValue, useAutomaticTransactionManagement);
         }
 
         public void MigrateUp(long targetVersion)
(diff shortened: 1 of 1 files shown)


src/FluentMigrator.Runner/MigrationRunner.cs
```diff
@@ -169,17 +137,28 @@ public void MigrateUp(long targetVersion)
         public void MigrateUp(long targetVersion, bool useAutomaticTransactionManagement)
         {
             var migrationInfos = GetUpMigrationsToApply(targetVersion);
+
+            if (!migrationInfos.Any()) return;
```
fubar-coder: Please apply `.ToList()` to `GetUpMigrationsToApply` and use `.Count` to avoide reenumeration.

src/FluentMigrator.Runner/MigrationRunner.cs
```diff
@@ -280,7 +259,6 @@ public virtual void ApplyMigrationUp(IMigrationInfo migrationInfo, bool useTrans
                 _alreadyOutputPreviewOnlyModeWarning = true;
             }
 
-            if (!migrationInfo.IsAttributed() || !VersionLoader.VersionInfo.HasAppliedMigration(migrationInfo.Version))
```
fubar-coder: Why was this line removed?

marcrocny: The logic indicated in that `if` is duplicated by local callers to the function; that is, for every invocation within the project, it evaluates to "true".  There were no independent tests of that function, but being public it could be consumed elsewhere.

marcrocny:
I'm having difficulty running tests locally (test-assembly load fails in NUnit).  From Travis I see that the failing tests were because of the expectation for ApplyProfiles to run, but apparently without migrations running.  I can adjust for that.

src/FluentMigrator.Runner/MigrationRunner.cs
fubar-coder: Due to the fact that this function is public, the removal of this condition would be a change in behaviour. When you revert the removal of this code, then I'll happily apply your PR.

fubar-coder:
It seems that adding the `ToList` fixed the unit tests 👍

src/FluentMigrator.Runner/MigrationRunner.cs
fubar-coder: BTW: You may create a private function that takes an additional boolean that might skip the validation like e.g.:

```
 if (!skipAttributedCheck && (!migrationInfo.IsAttributed() || !VersionLoader.VersionInfo.HasAppliedMigration(migrationInfo.Version)))
```

You'd then be able to call this private function from all other code paths, but with `skipAttributedCheck` set to `true`.

marcrocny:
Actually it was the removal of the short-circuit when the list was empty.  The unit tests expect Profiles to run even when there are no migrations to run.  That seems counterintuitive to me, though probably rarely encountered for the common Profile use-case of dev-db data population.  Even so, going to treat the tests as definitive of expected public behavior.

Making suggested changes now.  Since conforming to the public interface removes the behavior change, I'm changing the title on this one, too.

marcrocny:
Mark, I hadn't restored that missing line yet.  How do I go about reopening for that commit?

fubar-coder:
A PR cannot be reopened, I'll add the change myself

tommarien:
@marcrocny & @fubar-coder, imho this seems like a huge breaking change. Maintenance migrations    main usecase is not for populating dev db's, there are more there for people who still use stored procedures an want to recreate them every time to avoid something being broken by a schema change !

fubar-coder:
@tommarien There is no change in behaviour any more, because the guarding `.Any()` condition got removed and the previously removed `if (!migrationInfo.IsAttributed() || !VersionLoader.VersionInfo.HasAppliedMigration(migrationInfo.Version))` line got added again. This PR is currently a mere code clean-up to avoid duplicate code in `void MigrateUp()` and `void MigrateUp(long targetVersion)`.

tommarien:
@fubar-coder Thanks ;)

marcrocny:
Hi @tommarien, above when I wrote about the "usecase...for populating dev db's" I was writing about *Profiles*, not Maintenance.  In the latter case, if there a maintenance migration is being used for recreating procedures in case of schema changes (the use case that I'm concerned about!) doesn't "schema change" imply an active, unapplied migration needs to be run? In which case, even with the guard, the required maintenance would still be run.  I know this is now moot for this PR, but I just want to make sure I understand in case there's something I'm missing.

I also need to note: there are still two public behavioral changes from this PR, but I'm not sure how "breaking" they are:

# Maintenance migrations will be run when a specific-version variant of `MigrateUp()` is called.  Previous to this PR, migrations were only run on the latest-version variants of `MigrateUp()`.
# `BeforeEach` and `AfterEach` migrations will only be run before and after applied migrations.  Previously, because each migration was only tested against `VersionLoader` in `ApplyMigrationUp`, `*Each` maintenance would be run for every migration returned by `MigrationLoader.LoadMigrations()`.

fubar-coder:
@marcrocny You're right. There are behaviour changes. I misread the diff 😟 .

@tommarien **Not** executing `ApplyMaintenance` for a `void MigrateUp(long)` seems like an oversight (or even a bug). Running `BeforeEach`and `AfterEach` even though the migration isn't applied seems at least strange and TBH I don't believe that it's a breaking change. A not-running migration doesn't change the DB which means that there is (IMHO) no need to run some user code.

tommarien:
@marcrocny you are right there be no schema change except for maybe a stored proc change that could be in the maintenance after migration. 

@fubar-coder tying profiles to the fact that a migration was run is weird. Suppose a team member creates a profile to populate some users in the db. Given the change one would have to have a schema change before the population can be run. That sounds as unpredictable behavior to me. If you ask for a profile to be run you should get that exactly, not given when then

marcrocny:
@tommarien, regarding "tying profiles to migration run:" no worries, that behavior change was removed, because it broke tests that defined the behavior for profiles.

PR #790, opened 2017-03-23
Progress towards .NET Core support

eloekset:
Most of the work done by @volkanceylan to port the main projects to .NET Standard 1.6.

His work was meant to be a temporary solution for Serenity until an official .NET Core version of FluentMigrator is available. Ref #712


eloekset:
Do we have access to a build server that supports the latest version of MSBuild? If not, maybe we can write a script that calls the dotnet CLI commands for restore, build and pack against the `FluentMigrator.DotNetCore.sln`.

johnkors:
@eloekset I use appveyor and download what I need at build time:

https://github.com/johnkors/NoCommonsNET/blob/master/build.bat#L1
https://github.com/johnkors/NoCommonsNET/blob/master/InstallDotNetCore.ps1

eloekset:
After a lot of attempts, I finally found a way to build the solution on travis. I had to move to Trusty CI Environment (https://docs.travis-ci.com/user/trusty-ci-environment/#Using-Trusty) since the .NET Core SDK won't install on Ubuntu 12.04 (14.04 and 16.04 are supported).

Now this PR builds .NET Standard 1.6 libraries only. That's a starting point for the `feature/dotnet-core` branch. If this gets merged, more contributors can send PRs to add back mono builds and maybe target lower .NET Standard API levels (1.3 or maybe 1.2 https://docs.microsoft.com/en-us/dotnet/articles/standard/library#net-platforms-support).

PR #792, opened 2017-03-30
Write to log file

jcperada:
Added option for recording logs to a log file the same messages displayed on console. Removed `$(MSBuildBinPath)\Microsoft.CSharp.targets` on Import option on `FluentMigrator.Runner.csproj` to eliminate warning on build.


fubar-coder:
Please use a `CompositeAnnouncer` in combination with a `TextWriterAnnouncer`.

PR #793, opened 2017-04-04
MySqlProcessor was not announcing the SQL of embedded scripts.

BillVolz:
It looks like this might have been overlooked.

### DIFF
--- a/src/FluentMigrator.Runner/Processors/MySql/MySqlProcessor.cs
+++ b/src/FluentMigrator.Runner/Processors/MySql/MySqlProcessor.cs
@@ -88,6 +88,10 @@ namespace FluentMigrator.Runner.Processors.MySql
 
         public override void Execute(string template, params object[] args)
         {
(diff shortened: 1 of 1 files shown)


fubar-coder:
I'm not sure if the SQL scripts should be printed, because they are not dynamically generated by FluentMigrator. IOW: The statements are already known by the developer and announcing the migration version should be enough.

BillVolz:
Looking at the Oracle Processor, from what I can see it is logging scripts.  This is why I thought this was a specific issue with MySql.  My use case is that I want to use the Verbose Preview-Only mode to create a final script to audit the changes before a production deployment.   Manually going through the scripts to create this is not practical and is error prone.  

[Oracle Processor](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Runner/Processors/Oracle/OracleProcessor.cs#L139)

fubar-coder:

PR #795, opened 2017-04-11
Added StopOnError flag in order to pause console if some migration fails

LaykTaviskaron:
When some migration fails it is good for developer to somehow pause the console and take a look at the error message


fubar-coder:

PR #796, opened 2017-04-18
Migration gate date

samukce:
In some scnearios is needed to have a way to orchestrate when a specifc data migration can be run. That can be apply in a situation that specific data should be migrated in a specific date. 
So, I implemented a way to set a Start date, End date or Start and End date gap;

For instance, to setting a specific migration that can run only between 2017-04-20 to 2017-04-22, the class should be:

```c#
   [Migration(0001)]
   [StartGate(2017, 04, 20)]
   [EndGate(2017, 04, 22)]
   public class MigrationWithGateAttribute : Migration
   {
       public override void Up() { }
       public override void Down() { }
   }
```

That can be applied together feature toggle.


samukce:
@daniellee can you take a look in this PR?

tommarien:
@samukce Why not creating you own inherited migration class that contains this logic ?

samukce:
@tommarien yes, I can but I know more people can have the same request.

samukce:
@fubar-coder can you take a look in this PR?

samukce:
@tommarien make sense?

tommarien:
@samukce Still wondering what would be the use case of these gated migrations, but i understand why you want it to be in the framework itself. I just don't get the need for it :S

samukce:
@tommarien In traditional big companies that do not apply continuous delivery yet and that many teams involved in this process, it's hard to keep the version synchronized in the right time with the data that it need at the right time.

My main problem that I tried to solve was:
* I have big pipeline to deploy one new version;
* I have one set of data that need to go with this specific version;
* The set of data just make sense if was executed in a range of time;
* Think in this range of time as a seasonality, for instance; or one promotional campaign;

samukce:
@tommarien what you think? make sense?

fubar-coder:
Closed, as the implementation for #838 and #839 is the way to go.

Issue #801, opened 2017-05-09
Add SystemMethods.CurrentDateTimeOffset


Arithmomaniac:
For example, in SQL Server, it would resolve as `sysdatetimeoffset()`.

fubar-coder:
Fixed in release/2.0.0

PR #803, opened 2017-05-09
Add SystemMethods.CurrentDateTimeOffset

### LINKED ISSUE #802: SqlServer2005GeneratorTests use SqlServer2008Generator instead of 2005
I noticed, when working on a PR for #801, that SqlServer2005GeneratorTests creates an instance of SqlServer2008Generator and not SqlServer2005Generator. It has always been like this since the test class was created in commit 0d00e840.

I'll fix it in a separate commit in the upcoming PR for #801.

eloekset:
Solves #801 for SQL Server:
- Add SystemMethods.CurrentDateTimeOffset
- For SQL Server 2008 and newer, generate SYSDATETIMEOFFSET() to get values relative to the local time zone as opposed to GETUTCDATE().
- For SQL Server 2005 and older, fallback to GETUTCDATE() since that's the closest supported.

In addition, this PR fixes #802.


fubar-coder:

PR #806, opened 2017-05-22
Fix for issue #805 - use system methods in update statements in SQL Server

### LINKED ISSUE #805: SystemMethods.NewGuid used in .SetExistingRowsTo() generates an invalid value for SQL Server
This is related to https://github.com/fluentmigrator/fluentmigrator/issues/394. Same bug, different place.

Given this FluentMigrator code:

`Alter.Table("User")
.AddColumn("UniqueIdentifier").AsGuid().NotNullable().SetExistingRowsTo(SystemMethods.NewGuid)`

It generates the following SQL:

> UPDATE [dbo].[User] SET [UniqueIdentifier] = 'NewGuid' WHERE 1 = 1

And fails, because 'NewGuid' is a string and an invalid value at that, too. Correct SQL should look like this:

> UPDATE [dbo].[User] SET [UniqueIdentifier] = NEWID() WHERE 1 = 1

See: https://docs.microsoft.com/en-us/sql/t-sql/functions/newid-transact-sql

pkrakowiak:
Fixes #805 

It is now possible to use system methods (except CURRENT_USER) for update statements with SQL Server.

### DIFF
--- a/src/FluentMigrator.Runner/Generators/SqlServer/SqlServerQuoter.cs
+++ b/src/FluentMigrator.Runner/Generators/SqlServer/SqlServerQuoter.cs
@@ -24,6 +24,21 @@ namespace FluentMigrator.Runner.Generators.SqlServer
             {
                 return string.Format("N{0}", FormatString(value.ToString()));
             }
+
+            if (value != null && value is SystemMethods)
+            {
+                switch ((SystemMethods)value)
+                {
+                    case SystemMethods.NewGuid:
+                        return "NEWID()";
+                    case SystemMethods.NewSequentialId:
+                        return "NEWSEQUENTIALID()";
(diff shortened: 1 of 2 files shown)


fubar-coder:
Thank you. This looks very good 👍

pkrakowiak:
One thing - this is a more generic change that applies to all kinds of SQL Server queries. I needed it specifically for an UPDATE (via SetExistingRowsTo()), but as you can see I edited the quoter, so it will output proper server function names wherever it's used. This is actually good, I think - users can now call those methods where they need them as part of other queries and if they call it in a wrong place (like NEWSEQUENTIALID() outside of a DEFAULT constraint) they will get a SQL error.

This is new codebase for me, so correct me if I'm wrong.

fubar-coder:
I don't see this change as a problem.

Issue #748, opened 2016-11-21
Issue with ForeignKey() on column with other than default schema


ArneHB:
If you use the .ForeignKey method extension for .WithColumn then it uses the default schema and I'm not able to change the schema with .InSchema.

Example:
```
Create.Table("Company").InSchema("Foo")
    .WithColumn("Id").WithColumn("Id").AsInt32().NotNullable().PrimaryKey().Identity()
    .WithColumn("Name").AsString().NotNullable()

Create.Table("Users").InSchema("Foo")
    .WithColumn("Id").WithColumn("Id").AsInt32().NotNullable().PrimaryKey().Identity()
    .WithColumn("Name").AsString().NotNullable()
    .WithColumn("Company_Id").AsForeignKey("Company", "Id").InSchema("Foo");
```

If I try to run this then I get the following error message:
```
!!! An error occured executing the following sql:
ALTER TABLE [Foo].[Users] ADD CONSTRAINT [FK_Users_Company_Id_Company_Id] FOREIGN KEY ([Company_Id]) REFERENCES [dbo].[Company] ([Id])
The error was Foreign key 'FK_Users_Company_Id_Company_Id' references invalid table 'dbo.Company'.
Could not create constraint. See previous errors.

!!! Foreign key 'FK_Users_Company_Id_Company_Id' references
invalid table 'dbo.Company'.
```

I know you can use the Create.ForeignKey() instead as workaround but I'd like the other method to work too since it has been implemented and can be used.

RyanOC:
I was having some issues with this too. I had to do what you mentioned ArneBK. It would be nice for this to work as expected though.

Create.ForeignKey() 
     .FromTable("Users").InSchema(SchemaName).ForeignColumn("CompanyId")
     .ToTable("Company").InSchema(SchemaName).PrimaryColumn("Id");

BJMdotNET:
That "other then" in the title is really bothering me... ;-) http://writingexplained.org/then-vs-than-difference

eloekset:
I added some tests to verify this issue. Turns out the syntax used is a bit misleading:
```.WithColumn("Company_Id").AsForeignKey("Company", "Id").InSchema("Foo");```

```InSchema("Foo")``` here refers to the schema of the table to be created, which is already set to "Foo" in ```Create.Table("Users").InSchema("Foo")```. It doesn't however specify any custom schema for the primary key table of the foreign key expression. To do so, a slightly different syntax must be used:
[...]

PR #808, opened 2017-06-14
Avoid confusion reported in issue #748

eloekset:
Syntax InSchema() should not be available after ForeignKey() syntax on a column. It looks like InSchema() is applied to the primary key table referenced by the foreign key, but instead it is applied to the table created earlier in the CreateTable syntax.

Simply returning a syntax interface without the InSchema part from the ForeignKey syntax avoids this confusion. InSchema should be used immediately after Create.Table(...) anyway.


fubar-coder:

Issue #810, opened 2017-07-07
Allow --timeout=0 to mean infinite


jasondentler:
This matches the behavior of SqlCommand.CommandTimeout, which the documentation references. 

Here's the offending code:

`
            if (RunnerContext.Timeout == 0)
            {
                RunnerContext.Timeout = 30; // Set default timeout for command
            }
`

AdGalesso:
+1

fubar-coder:
Fixed in release/2.0.0

PR #811, opened 2017-08-03
Remove default connection timeout. Use provider's default instead.

eloekset:
To fix issue #810 I've made Timeout property of configurations nullable. The value is no longer set to a default value by FluentMigrator. When null, a connection's Timeout property doesn't get set, resulting in the provider's default timeout to be used.

In case of SQL Server, setting Timeout to 0 now results in an infinite timeout as requested in the issue.


fubar-coder:

PR #825, opened 2017-12-04
Update Max Decimal for MySql

fugaku:
Update Maximum Decimal Size for MySQL based on documentation stated [here](https://dev.mysql.com/doc/refman/5.7/en/precision-math-decimal-characteristics.html)

This is to fix issue #823

PR #779, opened 2017-02-24
Changed Nant build to use FluentMigrator.sln

escarls:

PR #777, opened 2017-02-19
Changed Travis CI to use FluentMigrator.sln

escarls:
- moved the build parameters from CI.proj to parameters in .travis.yml
- removed CI.proj as it's no longer in use (no references from any file).
- Added the missing projects from FluentMigrator (2010).sln.
- Set the build config for FluentMigrator.sln to the same as FluentMigrator (2010).sln, except for the changes to FluentMigrator.sln from checkins.
- Fixed some build warnings.

The Travis CI logs are now the same as they were before the changes, except for some fewer warnings.
The Nant logs (build.bat) are the same as they were before the changes, except for some fewer warnings.
Rake still builds equal nuget packages.

PR #827, opened 2017-12-15
Fixed nunit tests in Travis

escarls:
Due to Travis switched default test image, I've updated (hack) the script, so the NUnit tests will run again, as discussed in #825.

I will do a better and more correct fix, but those changes depends on that we're running on ONE sln file (#777 and #779 for the start).


escarls:
Refer to the [Travis blog](https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch) about the change from Precise (Ubuntu 12.04) to Trusty (Ubuntu 14.04)

PR #829, opened 2018-01-09
add breaking change identification

AlexCuse:
This change should allow developers to mark migrations as breaking changes.  These changes are not run by default but can be enabled by passing the flag --allow-breaking-changes / --abc. We have been using this for a while to identify changes that require special deployment steps, though our initial implementation was done against the 1.0.6 tag since we needed to keep transaction-per-session behavior.  Closing the loop now since transaction-per-session option has become available.

### DIFF
--- a/src/FluentMigrator.Console/MigratorConsole.cs
+++ b/src/FluentMigrator.Console/MigratorConsole.cs
@@ -52,6 +52,7 @@ namespace FluentMigrator.Console
         public bool NoConnection;
         public string WorkingDirectory;
         public bool TransactionPerSession;
+        public bool AllowBreakingChange;
         public string ProviderSwitches;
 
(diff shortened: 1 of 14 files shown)


AlexCuse:
Let me know how you'd like me to address the failing integration tests.  I just need to be able to define one breaking change migration to use for unit tests, can do so in another assembly or similar.  I don't *think* this is something that should require cross-platform testing.

fubar-coder:
This change looks good, but what are the failing integration tests you mentioned?

AlexCuse:
I could have sworn I saw something in the travis build output suggesting that it was trying to run "TestBreakingMigration" (defined in FluentMigrator.Tests.Integration.Migrations) but could be wrong - definitely not seeing any now.

PR #830, opened 2018-01-11
Update rake gems and build instructions

escarls:
I have updated the GemFile.lock to be able to setup the rake build environment easily on a Windows machine.
Changed Ruby version from 2.2.4 to 2.2.6, as it's impossible (without a lot of hacking) to make it work with 2.2.4 now due to 2.2.4 having issues reading the SSL cert of the rubygems.org site.
Basically, I've updated the nokogiri gem version (and requirements) in the lock file, so it's possible to install the build environment with the command bundler install, as it was intended.
This is tested with 2.2.6, 2.3.3 and 2.4.3 and will probably also work with versions less than 2.3.3 and 2.4.3 as long as those versions are updated for the SSL issue.
I did not update the albacore gem to avoid issues with the build.


escarls:
The build fails in Travis due to #827 is not merged yet.
Here's the [log](https://travis-ci.org/escarls/fluentmigrator/builds/327590009) of a Travis build of the PR where #827 is merged in.

haf:
If you're upgrading to this version of ruby, you should really consider going for v3 of albacore.

PR #831, opened 2018-01-20
.NET Standard 2.0 projects

eloekset:
Minimal changes to try to build for .NET Standard 2.0 and .NET 3.5, ref latest comments in issue #712 

- This PR will probably have to be updated several times before it can be merged into master. Maybe it's better to send it to a new branch instead.
- And I guess #827 and #830 should be merged first, so this PR can be rebased on that work.


eloekset:
Travis complains about the project format not supporting MSBuild 2003 format. We won't be able to do that for .NET Standard 2.0 projects. I guess we should should use .NET CLI now.

This is from the build log of travis before trying to use "dotnet build" instead of MSBuild:
/home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj: error : /home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj: The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.

eloekset:
I've been trying to install and run travis locally to investigate what's needed to use .NET CLI, following this guide: https://stackoverflow.com/a/41935867/1421847
But I give up for now. Maybe we should try AppVeyor, since that should be very easy to use for .NET projects? 
image

escarls:
@eloekset There should be a lot easier if you switch to csharp in language so you get support for dotnet without all the manual stuff.
I do not know if that fixes your issue, but it's worth a try.
When I played with this a year ago, I found that switching to csharp will give you more support.
Now, I did never try dotnet core in those builds, as I wanted my older PRs merged first.

eloekset:
```dotnet build ./src/FluentMigrator/FluentMigrator.csproj``` fails because it can't build with target framework net452 on Linux, which isn't very surprising. 

It should be super simple to build NuGet packages that targets both .NET Standard 2.0 and .NET 4.5.2 using .NET CLI. But that requires a Windows build environment. 

We could tweak this with mono and duplicate csproj files to build .NET Standard libs and do nuget pack manually, but that's a lot of unnecessary work.

Build error:
```
/usr/share/dotnet/sdk/2.1.4/Microsoft.Common.CurrentVersion.targets(1124,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.5.2" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed.
```

eloekset:
So now the travis build at least completes successfully. Then the AutomatedRelease config of TeamCity build and nuspec files are left to be updated.

eloekset:
Did someone disable TeamCity builds right now?

eloekset:
No, it started after a while. Seems like my attempt at migrating from NUnit 2 to 3 failed. Apparently it's not the local NUnit Console in the tools folder that is used by TeamCity, but a JetBrains.BuildServer.NUnitLauncher.exe, which I suppose can be upgraded to NUnit 3 by some configuration changes to the build definition? Does @escarls or @fubar-coder know how to fix this?

From the build log:
image

eloekset:
I'm closing this PR as I'm preparing a new one from a much cleaner branch.

PR #832, opened 2018-02-04
Another attempt at building .NET Standard 2.0 assemblies

eloekset:
Minimal changes to try to build for .NET Standard 2.0 and .NET 3.5, ref latest comments in issue #712

This PR replaces #831, which became a mess. This one will probably also have to be updated several times before it can be merged into master. And maybe it's better to send it to a new branch instead.

Now I'm trying to reuse existing projects to build .NET 3.5 assemblies and just build .NET Standard assemblies from new csproj files to avoid breaking any of the existing build stuff. The ideal solution using .NET Standard projects with several target frameworks turned out to break the build, and forced an upgrade to .NET 4.5.2, which turned out to be more difficult than expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
-language: c
-
+language: csharp
+dotnet: 2.1.4
 install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y mono-devel mono-gmcs nunit-console libfbembed2.5
+  - sudo apt-get install -y nunit-console libfbembed2.5
 
 script:
   - xbuild CI.proj
+  - dotnet build ./FluentMigrator.NetStandard.sln
   - sudo cp /var/lib/firebird/2.5/system/default-security2.fdb /var/lib/firebird/2.5/system/security2.fdb
   - sudo chmod a+rw /var/lib/firebird/2.5/system/security2.fdb
   - dpkg -L libfbembed2.5
--- a/FluentMigrator.NetStandard.sln
+++ b/FluentMigrator.NetStandard.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.27130.2027
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FluentMigrator.NetStandard", "src\FluentMigrator\FluentMigrator.NetStandard.csproj", "{658B1572-14A4-4E73-978D-9E49DE5D6E34}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FluentMigrator.Runner.NetStandard", "src\FluentMigrator.Runner\FluentMigrator.Runner.NetStandard.csproj", "{959D10C7-95F1-43AC-920C-C70E8BE73F60}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{658B1572-14A4-4E73-978D-9E49DE5D6E34}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{658B1572-14A4-4E73-978D-9E49DE5D6E34}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{658B1572-14A4-4E73-978D-9E49DE5D6E34}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{658B1572-14A4-4E73-978D-9E49DE5D6E34}.Release|Any CPU.Build.0 = Release|Any CPU
+		{959D10C7-95F1-43AC-920C-C70E8BE73F60}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{959D10C7-95F1-43AC-920C-C70E8BE73F60}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{959D10C7-95F1-43AC-920C-C70E8BE73F60}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{959D10C7-95F1-43AC-920C-C70E8BE73F60}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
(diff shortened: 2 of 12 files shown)


eloekset:
Seems like the **rakefile.rb** needs to be updated to run .NET CLI for dist builds, and **packages/packaging.rb** needs to be updated to generate more .nuspec content to make .nupkg files support both .NET 3.5, 4.0 and .NET Standard 2.0.

I won't be able to guess what changes are needed to make that work, so I'll either have to setup an environment to test this locally or call for @daniellee to have a look at it.

eloekset:
image

escarls:
@eloekset Did you try merging with #830?

haf:
@eloekset You're failing because you're not using the right version of albacore. In v3 I did a single breaking change, specifically for test_runner; we now have Expecto and it has a C# API, so I'm not maintaining NUnit test running any more. v3 supports building .net core nugets too.

haf:
PS; don't maintain your own "packaging" files, it's such a waste of time. Just send PRs if you want changes to albacore. Also, I've done the work you mention for supporting .net core now.

eloekset:
@escarls I did a rebase onto your branch in #830 now, but I still get the same error, so I didn't force push my branch in this PR.
image

@haf sounds like you know a lot about ruby and this build script. From what I can see out of **packages/packaging.rb** the .nuspec files are not manually maintained. 

I can call for @fubar-coder to create another feature branch in the main repo, and open another PR against that branch instead of master, so we can work together to make this build stuff work. We should probably keep the **feature/dotnet-core** branch in case we want to continue development to target lower .NET Standard versions. What about a **feature/netstandard20** branch for this easier job at targeting .NET Standard 2.0?

eloekset:
@fubar-coder also if you create a **feature/netstandard20** branch, maybe you can see if it's possible to merge this PR into that branch even if I opened it against **master**. If not, tell me, and I'll close this one and open a new one.

haf:
@eloekset Either run

```
gem uninstall albacore
gem install bundle
bundle
bundle exec rake
```

Or upgrade your Rakefile

```
bundle update
# then replace test_runner with `system` to just invoke `nunit-console-runner.exe` with your dlls.
```

haf:
https://github.com/Albacore/albacore/wiki/test_runner

haf:
https://github.com/Albacore/albacore/wiki

eloekset:
I have no idea what I'm doing, but this is the outcome of the first alternative: 😆
image

haf:
Darn, sorry.

```
gem uninstall bundle
gem install bundler
```

(note the r)

And remove the custom packaging; you haven't done your package management for it. I believe the error you're getting is because your own `packaging.rb` is dependent on a very old version of nokogiri.

Replace it with
https://github.com/Albacore/albacore/wiki/nugets_pack_v3

eloekset:
image

haf:
Hmm, right.

To get the second screenshot to work, you could upgrade albacore:

```
bundle update
```

For the first screenshot you need to find the version of v2 of albacore where I fixed your problem a few years ago.

eloekset:
Do we have a Docker image with everything pre-configured?

haf:
Yes, haaf/qvitoo-baseline has it all

But again, you need to update your bundle for it to work.

AeonLucid:
So, I tried to use this branch as suggested here https://github.com/fluentmigrator/fluentmigrator/issues/712#issuecomment-366898195.

I receive a null reference exception here https://github.com/eloekset/fluentmigrator/blob/master/src/FluentMigrator.Runner/Processors/DbFactoryBase.cs#L78 because `CreateDataAdapter` returns `null`.

eloekset:
@AeonLucid which provider are you using for SQLite? I created a sample project in this branch: https://github.com/eloekset/fluentmigrator/tree/feature/712-SampleProject
It has a solution file called **Fluent.NetStandardTest.sln** which has a project with a simple console runner that is hard typed to use **SQLServer** provider.

At first, I got the same problem as you did. I couldn't believe what I saw, because **System.Data.SqlClient 4.1.0** is included by default by the **FluentMigrator.Runner.NetStandard** project, so I expected that to just work. 

But I had to install **System.Data.SqlClient 4.4.2** in the sample project to get an adapter from the DbFactory. This is a screenshot from the sample app running, where the adapter is successfully created:
image

eloekset:
@AeonLucid I'm trying to make `SQLiteDbFactory` support **Microsoft.Data.Sqlite**, but then I find that it doesn't implement the `CreateDataAdapter()` method: https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite.Core/SqliteFactory.cs
It falls back to `System.Data.Common.DbProviderFactory`, which returns `null`: https://github.com/dotnet/corefx/blob/master/src/System.Data.Common/src/System/Data/Common/DbProviderFactory.cs

I published it in yet another branch if you want to take a look at it: https://github.com/eloekset/fluentmigrator/tree/feature/712-SampleProjectUsingMicrosoftDataSqlite

AeonLucid:
@eloekset Thanks, I will take a look at the Sqlite branch. Hopefully it is possible to get it working 😄

eloekset:
Ok, since they don't want to support DbDataAdatper, FluentMigrator needs another provider for SQLite. I don't think it's a good idea to base migrations on typed models like EFCore and Dapper are meant for. DataSets are great for migrations, and many(?) migration steps must be converted if the future version of  FluentMigrator drops support for that.

AeonLucid:
Yeah.. I'm going to take a different approach at porting the app to dotnetcore. Starting to run in a lot of other issue's as well.

eloekset:
Since I can't wait until the build scripts support the new .NET Standard libraries, I've built some scripts to build NuGet packages for internal use. It supports SQL Server on win10-x64 only, since that fits my needs right now.

If anyone are in need on something similar, I've published a branch where preview of NuGet packages for .NET Standard only can be built: https://github.com/eloekset/fluentmigrator/tree/feature/712-ConsoleRunnerNetStandard2_0
At the root folder there's a powershell script to publish and pack **FluentMigrator**, **FluentMigrator.Runner** and **FluentMigrator.Tools** (only Console project) for **win10-x64**. These packages shouldn't be published on the nuget.org feed. So I've just added them to my own feed at myget.org for internal use.

I've also added a sample repo here: https://github.com/eloekset/FluentMigrateConsoleNetCoreApp/

It uses the preview to run a sample migration against SQL Server.
image

giggio:
@fubar-coder Awesome, I see that the `release/2.0.0` is being frequently updated and is building with .NET Standard now. Any idea when 2.0 is to be released?

fubar-coder:
Its release is planned to be in two weeks. Please download it from MyGet and test it 😁

PR #865, opened 2018-04-27
Added ability to override the default conventionset

vincentparrett:
This allows users to actually use the IConventionSet architecture as intended. If not supplied, it will use the default convention set as before. 

PS. Apologies for the few (minor) formatting changes, I guess vs formatted the file when I saved.


fubar-coder:
Be aware that I will change your PR, because adding a default parameter broke binary compatibility and I'm not ready to release 2.1 yet 😀

fubar-coder:
No problem. Thank you for fixing the problem 👍

vincentparrett:
Thanks for the fast response, I am now using 2.07 with my own conventions and migrations running as expected.

PR #878, opened 2018-05-08
Fixed connections[machinename] never reached in code

escarls:
Not sure if this was on purpose or not.
If it was, then please decline the PR :)


fubar-coder:
@escarls Do we need a 3.0.1 now or do we want to wait a few days to accumulate enough changes?

PR #891, opened 2018-05-29
using the oracle beta driver that is compatible with dotnet core.

cgibbons:
In order for oracle driver to work with dotnet core you have to use the beta version of the driver, this will obviously need to be updated when they release the final version, last word from oracle is 3rd quarter 2018


fubar-coder:
I'll close this patch, because I'll only apply it partially (only for the `dotnet-fm` tool).

Oracle plans to release a non-beta version of the driver in Q3, but
it's the only Oracle driver that works under Linux/MacOS. The console
tool (`Migrate.exe`) is more Windows-centric and will therefore keep
using the standard Oracle ADO.NET library. The `dotnet-fm` is mostly
used on non-Windows platforms and is therefore predestinated to use
the new beta driver.

The statement from Oracle can be found on the
[Oracle website](http://www.oracle.com/technetwork/topics/dotnet/tech-info/odpnet-dotnet-ef-core-sod-4395108.pdf).

The console tool will switch to the new driver when it becomes stable.

PR #907, opened 2018-07-03
Added support to default schema name for VersionInfo table [dotnet fm --default-schema-name]

Alegrowin:
When I was using the Cli to create Pogo.User and Pogo.EatenBanana, the dbo.VersionInfo table was created or conflicted with other dbo.VersionInfo table. 

dotnet $Cli list migrations -p SqlServer2016 -a $Assembly -c $ConnectionString --namespace $Namespace --default-schema-name $DefaultSchemaName


fubar-coder:
Thank you, looks good to me 👍

Alegrowin:
Great, thank you! :) Do you need me to do something to release a new version of the CLI? Or you already have a process that handle that?

fubar-coder:
The CLI build is coupled to the library build to reduce problems with differences between CLI/library versions. We don't have enough changes yet that may justify a new version. Maybe I can/should upload a 3.1.4 only for the CLI? I'll take a look.

Alegrowin:
No need, I have generated a temporary version and hosted it in our local nuget repo. I wanted to make sure there was nothing else to do on my side. 

Thank you very much!

PR #919, opened 2018-08-29
Add ability to configure connection string dynamically.

TerraVenil:


fubar-coder:
Is this really dynamic when added as singleton? This sounds like it cannot be changed after it's been used at least once.

When you really want a dynamic connection string, then you should add your own *scoped* implementation of `IConnectionStringAccessor`.

fubar-coder:
Can you please add a xmldoc comment for the new function? Other than that, it really looks good to me.

TerraVenil:
My intension was to give ability to developer to initialize in lazy manner connection string for ProcessorOptions, for example usage of current extension method will be
```
.WithGlobalConnectionString(x =>
{
    var connectonStringService = x.GetService<IConnectonStringService>();
    return connectonStringService.Get();
})
```

TerraVenil:
Added xmldoc.

PR #940, opened 2018-10-25
Oracle 12C supports object name max length = 128

dfaruque:
Please consider this in next version.
https://oracle-base.com/articles/12c/long-identifiers-12cr2


fubar-coder:
A new alpha build should be available soon.

PR #943, opened 2018-10-26
Properly quote byte array for SQLite

elken:
The default method doesn't work for SQLite, using the below change it's now possible to store `byte[]`.

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteQuoter.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteQuoter.cs
@@ -25,7 +25,7 @@ namespace FluentMigrator.Runner.Generators.SQLite
 
         protected override string FormatByteArray(byte[] value)
         {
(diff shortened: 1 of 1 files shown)


src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteQuoter.cs
```diff
@@ -22,5 +22,18 @@ public override string QuoteSchemaName(string schemaName)
         {
             return string.Empty;
         }
+
+        protected override string FormatByteArray(byte[] value)
+        {
+            var hex = new System.Text.StringBuilder((value.Length * 2) + 2);
```
fubar-coder: Shouldn't this be a `+ 3`? The `X` and both single quotes `'`.

elken: Yes! My bad! Amending now.

elken:
Any idea when this will be merged into a release?

fubar-coder:
In the next couple of days. There are enough changes to justify a 3.2.0 release.

PR #944, opened 2018-10-26
Snowflake database support

ikudjoi:
Here is the first version of Snowflake support for FluentMigrator.

This was more laborious than I thought, problems that I had to overcome:
* There is no default schema in Snowflake. Quite a few tests do not specify schema, had to rewrite them for Snowflake.
* Snowflake identifier quoting is different at least from SQL Server that I'm familiar with. I wanted to support both quoting and leaving quotes away, so I implemented SnowflakeOptions.
https://docs.snowflake.net/manuals/sql-reference/identifiers-syntax.html
* The .NET connector of Snowflake does not support multiple statements in batch. I worked over this by implementing BatchParser that splits the commands separated by semicolon. I had to build support to SearchStatus and SpecialTokenInfo classes for this.
* There was a bug in class SqlBatchParser (explicitly specified newline was replaced by environment newline after first match.) 
* Another bug was found from SemicolonSearcher, it did not start search from the current index but from beginning every time.

The .NET connector is included here as compiled binaries. My intention is to replace that with NuGet package but we need the get my pull requests first merged.
https://github.com/snowflakedb/snowflake-connector-net/pulls


fubar-coder:
Is there a way for me to test your changes (e.g. a test installation)?

ikudjoi:
Hi, I'm sorry, Snowflake is cloud-only. If you utterly require I think it might be possible to create a temporary account for you on our cost.

I guess I need to work on quality review & check the CI results.

PR #947, opened 2018-10-30
Snowflake support

ikudjoi:
Support for Snowflake database
https://snowflake.com

I already had placed a PR but I rebased my work on current develop branch, deleted my branch on Github and pushed the new one. But my PR also disappeared. That was probably something that I shouldn't have done, sorry.

I updated the precompiled Snowflake .NET connector binaries and added references to the CLI projects, otherwise this is the same as before, but rebased.

My pull requests to the .NET connector project might get accepted today, so I might be able to replace the references to my own builds with NuGet package sometime soon.


ikudjoi:
The old PR seems to be still there but was automatically closed.
https://github.com/fluentmigrator/fluentmigrator/pull/944

fubar-coder:
OK, I'll wait a little bit when there is a chance to use a NuGet package for an updated Snowflake connector.

ikudjoi:
Pushed fixes for alter column syntax & more tests. @fubar-coder, have you noticed my private messages in gitter.im?

fubar-coder:
@ikudjoi I saw them just now. Thank you 👍 .

ikudjoi:
I had to do a lot of changes to the test because I thought there is no default schema in Snowflake. But today I learned I was wrong, there is one. New version supporting default schema is now pushed.

ikudjoi:
And now the new version of the .NET connector is out, removed own builds and replaced them with the new NuGet package.

ikudjoi:
I'm not sure if you want to embrace the latest commit I just pushed. I ran into a problem when using the FM without quoting identifiers. Snowflake converts unquoted identifiers to uppercase, so the versioninfo table contains column "VERSION". But FluentMigrator is trying to read column "Version".

I suggest it would not be too harmful to match the column case insensitive in this context. I found no other way to match it case insensitive, setting caseSensitive = false didn't work.

src/FluentMigrator.Runner/VersionLoader.cs
```diff
@@ -147,9 +147,23 @@ public void LoadVersionInfo()
             if (!AlreadyCreatedVersionTable) return;
 
             var dataSet = _processor.ReadTableData(VersionTableMetaData.SchemaName, VersionTableMetaData.TableName);
-            foreach (DataRow row in dataSet.Tables[0].Rows)
+            var dataTable = dataSet.Tables[0];
+
+            // Ignore case when looking for correct column.
+            foreach (DataColumn col in dataTable.Columns)
```
fubar-coder: This change is not optimal. It might be easier to just set `dataTable.CaseSensitive` to `true`

fubar-coder: > setting caseSensitive = false didn't work.

Uh, ok. But it still looks strange. Just get the column index and use that instead - but only if `Version` couldn't be found.

fubar-coder: And please use `OrdinalIgnoreCase`.

ikudjoi: I'm trying to find better solution.

src/FluentMigrator.Runner/VersionLoader.cs
```diff
@@ -147,9 +147,28 @@ public void LoadVersionInfo()
             if (!AlreadyCreatedVersionTable) return;
 
             var dataSet = _processor.ReadTableData(VersionTableMetaData.SchemaName, VersionTableMetaData.TableName);
-            foreach (DataRow row in dataSet.Tables[0].Rows)
+            var dataTable = dataSet.Tables[0];
+            var versionColumnIndex = dataTable.Columns.IndexOf(VersionTableMetaData.ColumnName);
+
+            // Find out correct column by case insensitive matching if column was not found. Setting dataTable.caseSensitive = false does not help for some reason.
+            if (versionColumnIndex < 0)
```
fubar-coder: Reminder for me: Check for -1 instead.

test/FluentMigrator.Tests/FluentMigrator.Tests.csproj
```diff
@@ -76,6 +78,10 @@
       <Link>%(Filename)%(Extension)</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="../../lib/Snowflake/*.*">
```
fubar-coder: Reminder for me: Check if it's still required

ikudjoi: That folder was removed, so no, it is not. Thanks for accepting the PR!

PR #959, opened 2018-11-20
Remove PoundSignSingleLineComment and add DoubleSlashSingleLineComment

igitur:
Hash signs are not comments on SQL Server. The `PoundSignSingleLineComment` (pound?) was stripping out references to temporary tables in my migration scripts.

Same for SQLite https://www.sqlite.org/lang_comment.html

I see it's used in `SqlAnywhereBatchParser` too, but i'm unsure about whether a hash sign represents a comment on SQL Anywhere. From what I could find at http://dcx.sybase.com/1200/en/dbreference/syco.html it seems hash signs aren't comments, but double slashes are. So I added a commit for SqlAnywhere too.

Let me know if I need to change anything. Would be awesome if this could be released in a 3.1.4 version.

### DIFF
--- a/src/FluentMigrator.Runner.Core/BatchParser/RangeSearchers/DoubleSlashSingleLineComment.cs
+++ b/src/FluentMigrator.Runner.Core/BatchParser/RangeSearchers/DoubleSlashSingleLineComment.cs
@@ -1,0 +1,32 @@
+#region License
+// Copyright (c) 2018, Fluent Migrator Project
+//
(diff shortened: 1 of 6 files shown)


igitur:
`//` is not a valid singe line comment token for MS SQL Server. I have already added the `SqlAnywhereBatchParserTests.cs` file to test the double slash on `SQL Anywhere`.

igitur:
Or do you mean you want an explicit test for MS SQL Server that `//` is not treated as a comment?

PR #961, opened 2018-11-20
Allow quoting to be disabled from dotnet-fm tool for postgres migrations.

groogiam:


groogiam:
Any idea when you might be able to publish this change to nuget?

fubar-coder:
No, not yet. I plan on releasing a 3.2.0, but I have to work through all those issues first, but I have currently little time.

lucasrossoni:
Hey All,

I've been trying to run my migration with dotnet -fm on a Postgres database, but it keeps putting double quotes, any idea what parameter I have to use to disable it?

I have created a class inherited from PostgresQuoter and registered on my IOC, but when I run a migration using dotnet-fm it doesn't work

PR #963, opened 2018-11-21
Add IMigrationRunner.LoadVersionInfoIfRequired()

### LINKED ISSUE #960: MigrationRunner.VersionLoader.LoadVersionInfo not executed in MigrationRunner constructor
Previously, in the constructors of `MigrationRunner` the `VersionLoader` property was constructor normally, but this has recently changed to a `new Lazy<IVersionLoader>(...)` implementation. 

The consequence is that `VersionLoader.LoadVersionInfo()` is not called when the `MigrationRunner` instance is constructed, but rather delayed. It would be called when `.MigrateUp()`, `MigrateDown()`, `Rollback()` or `RollbackToVersion()` is called.

However, these methods aren't the only ones that are used. For example, in my case, I'm constructing a `MigrationRunner` and deliberately using `runner.Up(migration)` to migrate only certain migrations. In my case, I deliberately want to exclude migrations related to seeding a database, but that's besides the point.

`MigrationRunner.Up()` doesn't call `VersionLoader.LoadVersionInfo()`, so the appropriate `VersionInfo` table is never created. And that creates other obvious problems.

My suggestion is to revert the lazy loading of `VersionLoader` to a normal constructor, or, alternatively, also call `VersionLoader.LoadVersionInfo()` in the other methods, where applicable. I can submit a PR, but I need your guidance first.

igitur:
Fixes #960

### DIFF
--- a/src/FluentMigrator.Runner/IMigrationRunner.cs
+++ b/src/FluentMigrator.Runner/IMigrationRunner.cs
@@ -108,5 +108,11 @@ namespace FluentMigrator.Runner
         /// </summary>
         /// <returns><c>true</c> when there are migrations available for a rollback</returns>
         bool HasMigrationsToApplyRollback();
+
+        /// <summary>
+        /// Creates the required version info database tables
+        /// </summary>
+        /// <returns><c>true</c> if the table had to be created or <c>false</c> when it already existed</returns>
+        bool LoadVersionInfoIfRequired();
     }
 }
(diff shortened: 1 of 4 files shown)


fubar-coder:
Can you retarget this PR to `fluentmigrator/develop`?

igitur:
ok, done.

igitur:
@fubar-coder I don't understand what Codacy's issue is. The new method contains a reference to instance data (`VersionLoader`) therefore the new method cannot be static.

fubar-coder:
Yeah, Codacy is sometimes wrong. Just ignore it.

Issue #957, opened 2018-11-20
FluentMigrator.DotNet.Cli --no-connection throws exception


groogiam:
I am trying to generate a migration script without a live database running. The following command results in an exception.

`dotnet-fm migrate -p Postgres --no-connection -a Migrations.dll -o=db.sql`

```text
-------------------------------------------------------------------------------
2018111900: Migration2018111900 migrating
-------------------------------------------------------------------------------

Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: Format of the initialization string does not conform to specification starting at index 0.
   at System.Data.Common.DbConnectionOptions.GetKeyValuePair(String connectionString, Int32 currentPosition, StringBuilder buffer, Boolean useOdbcRules, String& keyname, String& keyvalue)
   at System.Data.Common.DbConnectionOptions.ParseInternal(Dictionary`2 parsetable, String connectionString, Boolean buildChain, Dictionary`2 synonyms, Boolean firstKey)
   at System.Data.Common.DbConnectionOptions..ctor(String connectionString, Dictionary`2 synonyms, Boolean useOdbcRules)
   at System.Data.Common.DbConnectionStringBuilder.set_ConnectionString(String value)
   at Npgsql.NpgsqlConnection.GetPoolAndSettings()
   at FluentMigrator.Runner.Processors.GenericProcessorBase.<>c__DisplayClass6_1.<.ctor>b__1() in /home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner.Core/Processors/GenericProcessorBase.cs:line 102
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at FluentMigrator.Runner.Processors.GenericProcessorBase.EnsureConnectionIsOpen() in /home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner.Core/Processors/GenericProcessorBase.cs:line 128
   at FluentMigrator.Runner.Processors.GenericProcessorBase.BeginTransaction() in /home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner.Core/Processors/GenericProcessorBase.cs:line 146
[...]

PR #964, opened 2018-11-22
Fix for https://github.com/fluentmigrator/fluentmigrator/issues/957.

groogiam:
This allows --no-connection to work when used in conjunction with --preview.

There is a deeper issue in which the --no-connection options is not passed to the MigrationScopeHandler to prevent a connection attempt when creating a transaction. 

Ideally we could just check for a null connection string but the lazy connection string initialization assigns the provider name as the connection string. I was unable to determine if this was done for a particular reason.


fubar-coder:
There are some unit tests failing, e.g.: `ProfilesAreAppliedWhenMigrateUpIsCalledWithNoVersion`.

PR #1003, opened 2019-03-20
Update sample to benefit from .NET SDK csproj type

jzabroski:
@fubar-coder Mark, this is a slight improvement to the sample, in that you will be able to automatically embed all .sql files in a solution as embedded resources.  This will help FM users avoid merge conflicts.

Related to:
* https://github.com/dotnet/docs/pull/11424
* https://github.com/dotnet/sdk/issues/3026

### DIFF
--- a/samples/FluentMigrator.Example.Migrations/FluentMigrator.Example.Migrations.csproj
+++ b/samples/FluentMigrator.Example.Migrations/FluentMigrator.Example.Migrations.csproj
@@ -2,11 +2,19 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <Description>Assembly containing migrations for the example project</Description>
+    <!-- By default, *.cs files will be automatically included as compile items,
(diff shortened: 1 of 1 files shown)


samples/FluentMigrator.Example.Migrations/FluentMigrator.Example.Migrations.csproj
```diff
@@ -2,11 +2,19 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <Description>Assembly containing migrations for the example project</Description>
+    <!-- By default, *.cs files will be automatically included as compile items,
+    but *.sql files will be automatically included as None items.  This disables that behavior. -->
```
dsplaisted: In this comment, it seems unclear whether "that behavior" that is disabled includes including *.cs files as compile items (which it doesn't).  So I'd replace "that behavior" with "automatically including None items", or similar.

Issue #1026, opened 2019-05-09
Allow for TagsAttribute to be inherited from interfaces


zendever:
This is an extension of older tickets submitted #626 and #648.

I found a use case for allowing tags to be "combined" by various classifications. So, for example, we have environment specific tags ("Production", "Non-Production", "QA", etc.) but we also have database state tags ("New") which only be allowed to run on a new database instance, or "Testing" which might only seed data for testing.

Because both classifications can exist simultaneously, I found a suitable solution would be to allow tags to be inherited from multiple interfaces as well. So, for a case where we might want to do QA-specific test data when working on standing up automated integration tests. Since we have 2 different types of QA-level testing, it's possible that a QA automated test would have specific information that something like a exploratory QA test environment wouldn't.

I will submit a pull request for this change.

jzabroski:
For testing, we use maintenance stage AfterAll and a `UnitTest` tag.  Actual seeding migrations (e.g., for production fixes such as updating federally mandated rates on a certain effective date) are tagged with `[Tags(TagName.Development, TagName.Alpha, TagName.Beta, TagName.Production)]` - effectively everything but `TagName.UnitTest`

I also recommend using an object hydration library like AutoFixture, ObjectHydrator, etc. to create sample data.  I prefer AutoFixture for testing, and tools like ObjectHydrator for populating QA / SalesDemo databases since it's an easy way to add some humor to the QA environment and Sales demoes.

jzabroski:
@zendever How do you want this to work, exactly?

Taking the example from the documentation (which should be updated if we change how this works)...

```csharp
[Tags("DK", "NL", "UK")]
[Tags("Staging", "Production")]
[Migration(1)]
public class DoSomeStuffToEuropeanStagingAndProdDbs : Migration { /* ... etc ... */ }
```

Would you refactor this into the following:

```csharp
[Tags("DK", "NL", "UK")]
public interface ITagsCountryDkNlUk { } // marker interface for applying tags
[Tags("Staging", "Production")]
public interface ITagsStagingOrProduction { }
[Migration(1)]
public class DoSomeStuffToEuropeanStagingAndProdDbs : 
    ITagsCountryDkNlUk,
    ITagsStagingOrProduction,
    Migration
{
    /* ... etc ... */
}
```

[...]

PR #1027, opened 2019-05-16
Added ability for TagsAttribute to be applied to and inherited from i…

zendever:
…nterfaces. Applies to issue #1026

### DIFF
--- a/src/FluentMigrator.Abstractions/TagsAttribute.cs
+++ b/src/FluentMigrator.Abstractions/TagsAttribute.cs
@@ -5,7 +5,7 @@ namespace FluentMigrator
     ///<summary>
     /// Used to filter which migrations are run.
     ///</summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = true)]
     public class TagsAttribute : Attribute
     {
         /// <summary>
--- a/src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
+++ b/src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
@@ -92,12 +92,25 @@ namespace FluentMigrator.Runner.Infrastructure
 
         private static bool TypeHasTagsImpl(Type type)
         {
-            return type.GetCustomAttributes<TagsAttribute>(true).Any();
+            return GetInheritedCustomAttributes<TagsAttribute>(type).Any();
+        }
+
+        private static IEnumerable<T> GetInheritedCustomAttributes<T>(Type type)
+        {
+            var attributeType = typeof(T);
+
+            return type
+                .GetCustomAttributes(attributeType, true)
+                .Union(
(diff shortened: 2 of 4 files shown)


fubar-coder:
This PR looks good to me. What do you think, @jzabroski .

jzabroski:
:shipit: - But can @zendever  please also update the documentation for tags?  Michael, the documentation is under : https://github.com/fluentmigrator/documentation/blob/master/articles/migration/migration-filter-tags.md

jzabroski:
@zendever @mrmikal I'm wondering if this PR caused regressions in the 3.2.4 release around running migrations without any Tags. Have you tried 3.2.4, and do any of your migrations have zero tags?

jzabroski:
See #1211

jzabroski:
@zendever @mrmikal Actually, it's possible you fixed a bug, but that bug likely breaks a lot of users (including a project I work on).  Investigating further.

zendever:
@jzabroski - If I did somehow unintentionally break something for you, I apologize. I haven't tried 3.2.4. Looking at the change set #1027 , I'm not exactly sure how the functionality was materially changed from how the Tags feature was implemented previously, aside from adding the union of the inheritance chain. Let me take a look at the latest version and see what's what.

jzabroski:
Thanks, I've been looking at this for a couple of hours trying to understand what's being reported.

jzabroski:
@zendever After much debugging, I think the issue reported is a false positive.  I think you did great work... but there was only a handful of commits in the last release and this was the only one that touched tags, so it seemed reasonable place to look.  Sorry for the grief.  I owe you a beer.

zendever:
@jzabroski - No worries. Glad to hear it was a false positive. In looking at the unit tests for the DefaultMigrationInformationLoader, I see that only marginal testing was done there, but it did conclude that untagged migrations would be included by default. In researching, I modified the tests to use the DefaultMigrationConventions.Instance and added a few extra tests surrounding compounding and inheriting tags and they all ended up passing too...so I figured it may have been something else. I had been using a fork for the longest time and hadn't really pulled from the origin in a long while. It's good to see that everything is still working. Looks like now, I can probably pull in the full release!

jzabroski:
@zendever If you want to help improve FluentMigrator, your contributions would be extremely valuable.  Most of my time is spent trying to be the [KeeperOfTheFlame](https://wiki.c2.com/?ArchitectAsKeeperOfTheFlame), although I'm optimistic this year I will get a chance to write code, as some of the trickier problems are solved.

The main issue I want help with is re-doing how connection management works, to allow passwordless connections and better decouple FluentMigrator transactions from connection transactions via unit of work pattern.  I think someone of your skill level likely could tackle this.

As for your additional tests, please feel free to submit a PR to contribute them.  I feel those tests need to be re-written as orthogonal array testing strategies (OATS) in order to correctly guarantee coverage of all pairwise combinations.  Last night, I went through all the various ways you could conceivably filter migrations, and just in the default code, there is `MigrationConstraintAttribute`, `MigrationTraitAttribute`, `TagsAttribute`, and `MaintenanceMigration` vs. `Profile` vs `Migration` (regular).

Issue #890, opened 2018-05-29
OracleManaged Migrations fail with runtime Exceptions


cgibbons:
When trying to migrate an oracle database with version 3.0, runtime exceptions occur.  If using oraclemanaged the following exception occurrs

```text
Unhandled Exception: System.TypeLoadException: Could not load type 'System.Security.Principal.WindowsImpersonationContext' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
   at OracleInternal.ConnectionPool.PoolManager`3.CreateNewPRThreadFunc(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

This is using the oracle beta driver that supports dotnet core.

If using the Devart DotConnect driver, it fails because it says it can't find the correct version of Devart.Data, even though I can see it in the output directory.

To repro this is a netcoreapp2.1 project with the following dependencies

```xml
<ItemGroup>
        <!--<PackageReference Include="Oracle.ManagedDataAccess" Version="12.2.1100" />-->
        <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="2.12.0-beta2" />
        <PackageReference Include="FluentMigrator" Version="3.0.0" />
        <PackageReference Include="FluentMigrator.Console" Version="3.0.0" />
        <PackageReference Include="FluentMigrator.Runner" Version="3.0.0" />
        <PackageReference Include="FluentMigrator.Runner.Oracle" Version="3.0.0" />
</ItemGroup>
```

cgibbons:
This can also be recreated following your quickstart guide, but substituting in the oracle libraries instead.

cgibbons:
I've also tried using the devart driver, but it's failing with assembly load exceptions.  Do you want me to file that as a separate bug with repro steps?  Or append it this one?

cgibbons:
@fubar-coder you're right, I substituted the beta driver in place of the managed driver in the code and it runs file.  I'll submit a PR and you can choose to merge it if you'd like.

fubar-coder:
Please file a separate bug. A PR would be welcome. It's good that you were able to solve the problem.

cgibbons:
[...]

PR #1028, opened 2019-05-22
Update to Oracle.ManagedDataAccess.Core (v2.19.3)

kcrossman:
The Oracle Managed driver for .NET Core 2 has been out for some time. It should be suitable for your needs. This is related to an existing Issue (https://github.com/fluentmigrator/fluentmigrator/issues/890) and PR (https://github.com/fluentmigrator/fluentmigrator/pull/891), but they were dismissed at that time due to using the beta driver.

### DIFF
--- a/src/FluentMigrator.Console/FluentMigrator.Console.csproj
+++ b/src/FluentMigrator.Console/FluentMigrator.Console.csproj
@@ -23,8 +23,8 @@
     <PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="5.12.1" />
     <PackageReference Include="FSharp.Core" Version="4.6.2" />
     <PackageReference Include="Microsoft.Data.Sqlite" Version="2.0.1" />
-    <PackageReference Include="Oracle.ManagedDataAccess" Version="12.2.1100" />
     <PackageReference Include="MySql.Data" Version="8.0.11" />
+    <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="2.19.3" />
   </ItemGroup>
   <ItemGroup Condition=" '$(Platform)' == 'x86' ">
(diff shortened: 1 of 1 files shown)


jzabroski:
@kcrossman Curious... how does this compile/run?  Did Oracle mislabel this as requiring .NET Core?  The Console project targets net461.  It's also strange how Nuget.org says the package has zero dependencies.

kcrossman:
@jzabroski I was able to build. I believe it uses .NET Standard 2.0. The console does target 4.61, so that wouldn't make much sense. Should probably update the Console to at least 4.7.1 and retry... Is there a particular reason it hasn't been updated to a newer Framework version or .NET Core 2.2?

fubar-coder:
We're already using the beta Oracle.ManagedDataAccess.Core driver for (IIRC) `dotnet-fm`. I think that it would be good to update all places where we reference an Oracle ADO.NET driver.

It **should** be compatible with .NET 4.6.1 as it is (with some additional assemblies) compatible with .NET Standard 2.0.

kcrossman:
@fubar-coder I'd rather not rely on specific parts being usable or not and make exceptions everywhere in my code just to satisfy a third party. Either this library supports .NET Core 2 and future .NET development or it doesn't. I'll look at this later when I have another moment to get it working.

kcrossman:
I'll close this for now given the current issue(s) and create a new PR with a more-complete change at some point soon.

jzabroski:
Hey @kcrossman I looked into it and can confirm that Oracle is using that
newer package widely - it's the one referenced in
https://www.nuget.org/packages/Oracle.EntityFrameworkCore/ - so in general
I agree that is what we should be using going forward.  However, we
probably want to have some way for people still on net461 to target the
older provider.  I believe FM 3.2.1 can load in an net461 assembly context.

PR #1042, opened 2019-07-05
Fix Oracle dialect create sequence NOCACHE

foo-543674:
Oracle raise error when below.
```csharp
Create.Sequence().Cache(1);
```
`ORA-04010 the number of values to CACHE must be greater than 1`
Need's to be
```sql
CREATE SEQUENCE [sequence_name] NOCACHE
```
[Docs](https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571)

### DIFF
--- a/src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
+++ b/src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
@@ -415,9 +415,18 @@ namespace FluentMigrator.Runner.Generators.Generic
                 result.AppendFormat(" START WITH {0}", seq.StartWith);
             }
 
+            const long MINIMUM_CACHE_VALUE = 2;
             if (seq.Cache.HasValue)
             {
+                if (seq.Cache.Value < MINIMUM_CACHE_VALUE)
+                {
+                    return CompatibilityMode.HandleCompatibilty("Cache size must be greater than 1; if you intended to disable caching, set Cache to null.");
+                }
                 result.AppendFormat(" CACHE {0}", seq.Cache);
+            }
+            else
+            {
+                result.Append(" NO CACHE");
             }
 
             if (seq.Cycle)
--- a/src/FluentMigrator.Runner.Hana/Generators/Hana/HanaGenerator.cs
+++ b/src/FluentMigrator.Runner.Hana/Generators/Hana/HanaGenerator.cs
@@ -92,9 +92,18 @@ namespace FluentMigrator.Runner.Generators.Hana
                 result.AppendFormat(" START WITH {0}", seq.StartWith);
             }
 
+            const long MINIMUM_CACHE_VALUE = 2;
             if (seq.Cache.HasValue)
             {
+                if (seq.Cache.Value < MINIMUM_CACHE_VALUE)
+                {
+                    return CompatibilityMode.HandleCompatibilty("Cache size must be greater than 1; if you intended to disable caching, set Cache to null.");
+                }
                 result.AppendFormat(" CACHE {0}", seq.Cache);
+            }
+            else
+            {
+                result.Append(" NO CACHE");
             }
 
             if (seq.Cycle)
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
@@ -124,16 +124,18 @@ namespace FluentMigrator.Runner.Generators.Oracle
                 result.AppendFormat(" START WITH {0}", seq.StartWith);
             }
 
+            const long MINIMUM_CACHE_VALUE = 2;
             if (seq.Cache.HasValue)
             {
-                if(seq.Cache == 1)
+                if (seq.Cache.Value < MINIMUM_CACHE_VALUE)
                 {
-                    result.Append(" NOCACHE");
+                    return CompatibilityMode.HandleCompatibilty("Oracle does not support Cache value equal to 1; if you intended to disable caching, set Cache to null. For information on Oracle limitations, see: https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571__GUID-7E390BE1-2F6C-4E5A-9D5C-5A2567D636FB");
                 }
-                else
-                {
-                    result.AppendFormat(" CACHE {0}", seq.Cache);
-                }
+                result.AppendFormat(" CACHE {0}", seq.Cache);
+            }
+            else
+            {
+                result.Append(" NOCACHE");
             }
 
             if (seq.Cycle)
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
@@ -126,7 +126,14 @@ namespace FluentMigrator.Runner.Generators.Oracle
 
             if (seq.Cache.HasValue)
             {
-                result.AppendFormat(" CACHE {0}", seq.Cache);
+                if(seq.Cache == 1)
+                {
+                    result.Append(" NOCACHE");
+                }
+                else
+                {
+                    result.AppendFormat(" CACHE {0}", seq.Cache);
+                }
             }
 
             if (seq.Cycle)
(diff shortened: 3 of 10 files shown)


jzabroski:
> The change sounds reasonable.
> 
> @jzabroski Maybe we should check for `< 2` instead of `== 1` ?

1. The issue seems to be that if you don't specify CACHE or NO CACHE in Oracle, it defaults to 20. Quoting [the link @foo-543674 provided](https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571__GUID-7E390BE1-2F6C-4E5A-9D5C-5A2567D636FB):
    > ### `NOCACHE`
    > Specify `NOCACHE` to indicate that values of the sequence are not preallocated. If you omit both `CACHE` and `NOCACHE`, then the database caches 20 sequence numbers by default.
2. DB2 supports [only values 2 and above](https://www.ibm.com/support/knowledgecenter/en/SSEPEK_10.0.0/sqlref/src/tpc/db2z_sql_createsequence.html): 
    > The minimum value is 2. The default is CACHE 20.
3. SQL Server _may_ interpret 1 as NO CACHE, but this is unofficial/un-documented behavior and my analysis below suggests even SQL Server is internally inconsistent here.  [SQL Server `SEQUENCE` documentation ](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-sequence-transact-sql?view=sql-server-2017) does not say what the default cache size is. It uses an example of 50, but does not say 50 is the default.

# My proposed behavior in pseudo-code

1. if Cache == null, then append `NO CACHE`.
2. If Cache.Value == 1, then `throw new ArgumentOutOfRangeException(nameof(Cache), "Oracle does not support Cache value equal to 1; if you intended to disable caching, set Cache to null. For information on Oracle limitations, see: https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571__GUID-7E390BE1-2F6C-4E5A-9D5C-5A2567D636FB");`

The general idea here is we should never use db engine specific defaults, because they're inconsistent across engines and FM should provide consistency across engines.  Ergo, treating `CACHE == null` as `NO CACHE` makes a ton of sense.

# My Analysis
I don't love this PR, because it assumes you'll know that if you pass in 1, it means NOCACHE.  On the other hand, I just tested the following on SQL Server, and it seems "CACHE 1" is treated (with a warning) as equivalent to `NO CACHE`, and **this behavior is not officially documented by Microsoft**:

> ```sql
> CREATE DATABASE Test;
> GO
> 
> USE Test;
> GO
> 
> CREATE SCHEMA TestSchema;
> GO  
> 
> CREATE SEQUENCE TestSchema.CountBy1
> 	START WITH 1
> 	INCREMENT BY 1
> 	CACHE 1;
> GO
> ```
> Outputs:
> `The cache size for sequence object 'TestSchema.CountBy1' has been set to NO CACHE.`

The concept of [`SEQUENCE` was introduced in SQL:2003 standard](https://en.wikipedia.org/wiki/SQL:2003#New_features), but, if you read the standard, you will see the `CACHE` concept is **not defined**.  A different way to confirm this is through running the following query after my example above, and you'll see the ANSI interface for querying `SEQUENCE`s does not mention caching:

> ```sql
> SELECT * FROM INFORMATION_SCHEMA.SEQUENCES
> ```
> Outputs:
> ```
>SEQUENCE_CATALOG  SEQUENCE_SCHEMA  SEQUENCE_NAME  DATA_TYPE  NUMERIC_PRECISION	NUMERIC_PRECISION_RADIX	 NUMERIC_SCALE  START_VALUE                MINIMUM_VALUE         MAXIMUM_VALUE        INCREMENT  CYCLE_OPTION  DECLARED_DATA_TYPE  DECLARED_NUMERIC_PRECISION  DECLARED_NUMERIC_SCALE
> ----------------  ---------------  -------------  ---------  -----------------  -----------------------  -------------  -------------------------  --------------------  -------------------  ---------  ------------  ------------------  --------------------------  ----------------------
> test              TestSchema       CountBy1       bigint     19                 10                       0              1                          -9223372036854775808	 9223372036854775807          1             0              bigint                          19                       0
> ```

However, it gets more interesting when we query the SQL Server internal storage - it did in fact mark this sequence as cached - see below!  So, is the warning correct and there is some hack in SQL Server that re-directs cache_size = 1 to "no cache"?  Who knows.

> ```sql
> SELECT * FROM sys.sequences
> ```
> Outputs:
> ```
> name      object_id  principal_id  schema_id  parent_object_id  type  type_desc  create_date                 modify_date              is_ms_shipped  is_published  is_schema_published  start_value  increment  minimum_value         maximum_value        is_cycling  is_cached  cache_size  system_type_id  user_type_id  precision  scale  current_value  is_exhausted  last_used_value
> --------  ---------  ------------  -------    ----------------  ----  ---------  --------------------------  -----------------------  -------------  ------------  -------------------  -----------  ---------  --------------------  -------------------  ----------  ---------  ----------  --------------  ------------  ---------  -----  -------------  ------------  ---------------
> CountBy1  885578193  NULL          5          0                 SO	SEQUENCE_OBJECT	2019-07-05 10:57:46.007	 2019-07-05 10:57:46.007  0              0             0                    1            1          -9223372036854775808  9223372036854775807  0           1          1           127             127           19         0      1              0             NULL
> ```

And you can see that, internally, if you don't specify a cache value, the size is null:

```sql
CREATE SEQUENCE TestSchema.CountBy1_CacheDefault
	START WITH 1
	INCREMENT BY 1
GO

SELECT is_cached, cache_size FROM sys.sequences WHERE name = 'CountBy1_CacheDefault'
/*
is_cached  cache_size
---------  ---------
1          NULL
*/
```

jzabroski:
I asked on Twitter for a bit more help with this for SQL Server: https://twitter.com/johnzabroski/status/1147195408109723648

jzabroski:
I think the correct patch is:

```csharp
            const long MINIMUM_CACHE_VALUE = 2;
            if (seq.Cache.HasValue)
            {
                if(seq.Cache.Value < MINIMUM_CACHE_VALUE)
                {
                    throw new ArgumentOutOfRangeException(nameof(Cache), "Oracle does not support Cache value equal to 1; if you intended to disable caching, set Cache to null. For information on Oracle limitations, see: https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571__GUID-7E390BE1-2F6C-4E5A-9D5C-5A2567D636FB");
                }
                result.AppendFormat(" CACHE {0}", seq.Cache);
            }
            else
            {
                result.Append(" NOCACHE");
            }
```

jzabroski:
Some other code review thoughts:

1. Should GenericGenerator let users configure the syntax template?  Mark, does the 4.0 branch allow this with Liquid templates?  OracleGenerator uses `NOCACHE` but DB2 + SQLSERVER use `NO CACHE` (note the space between no and cache).

foo-543674:
Thank you for review @jzabroski and @jzabroski.
I modified and pushed OracleGenerator as @jzabroski suggested.
Is there should throw exception when cache size 1 passed to Generator even DB supports "CACHE 1"?

jzabroski:
I approve. @fubar-coder do you agree? Only thing is Foo only did Oracle behavior not the others. But it's straightforward to apply to the others.

jzabroski:
@foo-543674 I think that SQL Server is the only one that allows `Cache 1` ,and I'm highly skeptical about that behavior to the point I'm adding it to my list of sanity checks as a professional SQL Server consultant.

In short, I think every provider should throw if Cache is less than 2. It's just illogical and I fail to see a real world use case. FM should help engineers by making bad code fail hard, before it gets to production. 

Itzik Ben-Gan, who is a top tier SQL Consultant, confirmed for me in the Twitter thread that SQL Server's implicit default is 50 but the fact is that SQL Server docs say they reserve the right to change it.

foo-543674:
I pushed changes for other providers.
Maybe MINIMUM_CACHE_VALUE should be the global scope, but I can't find better practice where to put utilities for Generators so I copy & paste all generators that use Sequence.
Postrgre not support "NO CACHE" syntax and "CACHE 1" behaves as "NO CACHE" according to [doc](https://www.postgresql.org/docs/9.5/sql-createsequence.html).
Please refacter or modify error message as you like.

jzabroski:
This looks great! Thanks for catching the Postgres `Cache 1` docs, too!

jzabroski:
@foo-543674 We have an issue with the build server at the moment, but this will be merged once we figure out why it timed out.

foo-543674:
All right. I hope it will resolve.

jzabroski:
Made small progress today fixing the build, via #1044 - likely won't fix the issue.

Discussed with @fubar-coder - we think we can sidestep the recent Travis CI community changes for C# via installing docker-ce and downloading dotnet-sdk image faster than compiling dotnet sdk from scratch.

PR #1060, opened 2019-09-11
Support renaming columns in MariaDb 10.2

tkallevik:
MariaDb 10.2 returns the string 'NULL' for column default, instead of actual NULL as in previous versions. This was an intentional change by them, to remove some ambiguity between NULL as in "there is no column default" and NULL as in "NULL is the column default". https://jira.mariadb.org/browse/MDEV-13132

To support this change, the column definition when renaming a column should not quote NULL-strings from the column_default, as they actually mean that the default is NULL (not 'NULL').


fubar-coder:
Thank you for your PR. Did you test that the unit tests still work?

tkallevik:
@fubar-coder I ran the unit tests and all pass just fine.

I also tried running the integration tests for MySQL, but there seems to be failing tests there already. I ran the tests before my change and 40 tests failed against MySQL 5.6.43, and after my change "only" 35 tests failed.

I am not sure if the integration tests for MySQL are ran frequently, and if everything is expected to work or not? I followed this https://fluentmigrator.github.io/articles/guides/contribution.html when running the integration tests, only replacing "Postgres" with "MySQL".

fubar-coder:
OK, thank you for the clarification. We still have to find a good way to run the integration tests for various databases. You followed the documentation and thanks to you I'm now aware of problems with the unit tests.

jzabroski:
@tkallevik We're hoping to move toward docker images for running some of these tests, but there are multiple problems with doing so:

* Travis CI build time will be extended by length of time to start docker image as a container, plus run those tests
* Travis CI build time is already bad (get what you pay for - free service)
* Time it takes to create all the docker images and maintain them

ikudjoi:
Just thinking if Snowflake could provide this project free account for integration testing. The project would neither need much storage space nor computing resources.

fubar-coder:
This sounds good. I'm also curious if GitHub actions might be interesting for us.

eloekset:
Azure Pipelines provides [unlimited free use for open source projects](https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/) and 10 parallel builds...
...and it [supports docker images](https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/build-image)...
...and [it's well integrated with GitHub](https://github.com/marketplace/azure-pipelines).

fubar-coder:
I still have bad dreams from when I tried the predecessor that was integrated into "Visual Studio Online". It was a PITA to configure and the build agent was totally underpowered, which caused build times of several minutes for projects that were merely more than a "Hello World". However, it **might** be worth looking at it again.

eloekset:
I'm using it for my hobby projects and I was using it for a lot of projects at my previous job. It works great now. And both GitHub and Azure DevOps are Microsoft services now, so I guess they'll continue to work great together for the foreseeable future. I can help testing it out for FluentMigrator, if you want me to.

fubar-coder:
I created issue #1061 where we should continue discussing the situation.

tkallevik:
Good idea :)

As for this PR I have manually tested the method I changed against both MySQL and MariaDb, and both provide the same result after the change, no matter which version I have tried.

jzabroski:
@fubar-coder Should we merge this?
@tkallevik Sorry for the delay.

My thoughts are: I think we need to do research on which version of MySQL handles INFORMATION_SCHEMA in the way you describe.

10.2.1 - https://jira.mariadb.org/browse/MDEV-10134 - Add full support for DEFAULT
10.2.7 - https://jira.mariadb.org/browse/MDEV-13132 - Information Schema does not show whether column default is expression or literal

So... if you're on MariaDB 10.2.1 - should we warn?  Is there a way to query the Maria DB instance for the current verison number?

jzabroski:
@tkallevik Sorry for the delay giving a more thoughtful review to this.  I had returned from my honeymoon mid-August and was under water and trying to get back into the swing of things, and not thinking about product management-level questions when reading your original PR description.

fubar-coder:
The change looks good to me and I think that it doesn't break anything else, because it only checks for an additional condition (`COLUMN_DEFAULT` being `NULL`). Everything would still work and the query would only produce an SQL fragment, which would be valid in all cases.

PR #1072, opened 2019-10-02
Dotnet 3.0

ikudjoi:
You seem to be busy migrating to new CI environment etc but I'd still like to suggest adding new .NET core version to the target frameworks. With this change we could hopefully use FluentMigrator in Docker container that has only 3.0 sdk installed. Currently we need to install both in order to use FM package from MyGet.

### DIFF
--- a/src/FluentMigrator.DotNet.Cli/FluentMigrator.DotNet.Cli.csproj
+++ b/src/FluentMigrator.DotNet.Cli/FluentMigrator.DotNet.Cli.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFrameworks>netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
     <ToolCommandName>dotnet-fm</ToolCommandName>
     <PackAsTool>true</PackAsTool>
     <OutputType>Exe</OutputType>


jzabroski:
My understanding is targeting packs are decoupled from runtime environments, and you should be able to build from a docker .NET 3.0 SDK container a .netcoreapp2.2 targeting pack.

In other words, I don't think this is strictly necessary for you to do what you want.  Having not tried it myself.... What error are you getting?

ikudjoi:
Sorry, my bad, I don't need to build and don't need SDK 3.0 but I do need 2.2. runtime, 3.0 runtime won't do.

CrispyIce:
I have the same issue, I want to run the cli from a azure pipeline task that only has dotnet core 3 installed - no joy

jzabroski:
Sorry all, I think from the original description I didn't understand the limitation.  @CrispyIce 's last reply made the problem crystal clear.  The DotNet Core CLI paradigm is a bit unfortunate.

jzabroski:
@fubar-coder This is another issue with how the new DotNet Core dependency injection/deps.json/targeting pack/runtime pack model has let us down a bit.  Are you OK with merging this PR to solve this issue?

fubar-coder:
Hmm, I hope that this change works, because I don't know if it works if we support multiple TFMs. I guess that we'll just have to try and see if it works.

jzabroski:
@fubar-coder Can we push an alpha version of 3.2.2 to your myget and have @ikudjoi  and @CrispyIce  test it and let us know?  Seems faster than creating an environment where we only have .net sdk / runtime installed in order to know if installing the global tool works as expected?

fubar-coder:
@jzabroski No, MyGet changed the subscriptions and my open source subscription isn't available anymore, so all my MyGet feeds are read-only. Maybe we should use/create a public prerelease feed on Azure  DevOps?

@eloekset Would this be possible?

eloekset:
@fubar-coder yes, it's possible, but still in preview. I created issue 1101 to continue investigation and discussion about a new NuGet feed for prereleases.

jzabroski:
Very well.  MyGet/Assembla is not a very good company, anyway.  Why MyGet only supports pinning packages vs. package promotion is beyond me.

ikudjoi:
> @fubar-coder Can we push an alpha version of 3.2.2 to your myget and have @ikudjoi and @CrispyIce test it and let us know? Seems faster than creating an environment where we only have .net sdk / runtime installed in order to know if installing the global tool works as expected?

Yes, I can confirm that my pull request would solve our problem, probably for @CrispyIce too.
I compiled FluentMigrator from my branch, copied the netcoreapp3.0 binary folder into my solution, removed .NET 2.2 runtime, removed dotnet tool install command I used to install FluentMigrator and replaced FluentMigrator calls with direct calls to the copied binary and it works like a charm in .NET 3.0.

acac99:
Hey guys,
Are we good for this to be merged? Also in a similar situation.

jzabroski:
Thanks for the nudge - simply fell off the radar.

jzabroski:
The confusing thing for me is why doesn't 2.2 roll-forward to 3.0?  I asked @dagood at Microsoft for help: https://github.com/dotnet/core/issues/3487#issuecomment-568912970

jzabroski:
@fubar-coder, I think we have to merge this and just accept that for every major release we need to update targetframeworks. It is likely sufficient to target  .netcoreapp3.0, due to "minor version roll forward".

https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward

jzabroski:
@CrispyIce @ikudjoi @acac99 Sorry this took so long to research.  I finally figured out the two options, as I cover in #1139 and CC'ed you on that issue.

My preference is we specify `<RollForward>MajorVersion</RollForward>`, for the reasons I list in that issue.  As we begin to roll out end-to-end integration tests with docker, adding more TargetFrameworks will significantly increase the run-time of those integration tests.

jzabroski:
@eloekset Fixed the develop branch build pipeline, so this is now merged and resolved.  However, since the myget repo is currently frozen, I am not sure how we will hand over the artifacts to end users.  If you can give some feedback @ikudjoi on whether you can access the feed from AzureDevOps, please let me know.  I'll try to ping you on gmail so you definitely see this.

ikudjoi:
Thanks! I'm letting my old colleagues know about this change, I've personally just changed employer and no longer working with .NET projects, so ...

Well actually I thought I could just quickly test it out anyway, but I can't locate the Nuget feed address. To me it seems that there are some problems with the build pipeline?

https://dev.azure.com/fluentmigrator/fluentmigrator/_release?_a=releases&view=mine&definitionId=1

Release 21 should be dotnet 3 + Snowflake -friendly, right?

@antru, are you the Antti Ruokomäki that I used to work with?

antru:
> @antru, are you the Antti Ruokomäki that I used to work with?
@ikudjoi yup! I'll follow this discussion

jzabroski:
You guys have the coolest names aside from Icelanders (I have an Iceland t-shirt that says "Come see my Eyjafjallajökull erupt!").  Congratulations on the new job!

The pipeline failure is a bit misleading.  Create a new feed in VIsual Studio called FluentMigrator2 and connect it to https://pkgs.dev.azure.com/fluentmigrator/fluentmigrator/_packaging/fluentmigrator2/nuget/v3/index.json

We will probably re-name this feed in the future from fluentmigrator2 to something more appropriate, so please anticipate it will break at some point or stop being updated. Its named strange because AzureDevOps has weird rules around how long it takes  to permanently delete a misconfigured feed.  I'm not sure if AzureDevOps supports URL redirects for renamed feeds.

Issue #997, opened 2019-03-08
Move from Mono.Options to System.CommandLine


jzabroski:
Main reason is the coolness of dotnet-suggest

fubar-coder:
Having `dotnet-suggest` support would surely be nice. It seems that the first alpha release is out.

Repository: https://github.com/dotnet/command-line-api
We could probably use the `DragonFruit` model for the console runner and the `Experimental` package for `dotnet-fm`?

jzabroski:
@fubar-coder 
> We could probably use the `DragonFruit` model for the console runner 
Yes, if the following permalink is where the Console does its parsing, then its a single-level, and maps well to DragonFruit model: https://github.com/fluentmigrator/fluentmigrator/blob/1333f691be6dfa573c14bd0d46ead762d06d7343/src/FluentMigrator.Console/MigratorConsole.cs#L86

> and the `Experimental` package for `dotnet-fm`?

Yes, we can move to `System.CommandLine.Experimental` https://github.com/dotnet/command-line-api/wiki/Your-first-app-with-System.CommandLine - it would probably be wise to fix the PackageReference so it doesn't auto-update dependencies.

jzabroski:
@dgx Related to your pull request: This is I think where we want to be long-term.  The problem is that System.CommandLine packages (e.g. https://www.nuget.org/packages/System.CommandLine.Experimental) only targets .NET Standard 2.0 right now, and it's still alpha technology.  So it's throwing good money after bad to re-write our options parsing logic using Microsoft's alpha released APIs.  Still, if FluentMigrator were to have a "tech radar", this would be an "incoming" technology we want to on-board in the next 2-3 years.

PhenX:
System.CommandLine seems active again https://www.nuget.org/packages/System.CommandLine#versions-body-tab
And targets net8.0 also, but still in beta

jzabroski:
per my last conversation with them it will ship in net10.0

PR #1082, opened 2019-10-11
update default-schema-name argument to properly accept a value

dgx:
It looks as though the new default-schema-name argument was improperly implemented.
Currently the value being used for the default schema is "default-schema-name" no matter what you pass to the argument.
Convention appears to dictate that in OptionSet, arguments that accept a value must be programmed with an equal sign on the end e.g. "default-schema-name=" for the argument parsing engine to accept a value.


jzabroski:
@dgx good catch! Do you know of any fluent libraries for writing tests for Mono.Options?  Our options parsing stuff is a bit insane right now.  We use Mono.Options, NateMcMaster.CommandLine.Utils.  There is a ticket to unify everything with System.CommandLine. #997

PR #1084, opened 2019-10-15
Postgres10 - auto-increment columns with identity

EduAlbSantos:
Included new (and recommended) standard way to define auto-increment columns for Postgres10: "identity columns". 
Options implemented: "BY DEFAULT" (default) and "ALWAYS"; Data types "serial" and "bigserial" ignored; "ALTER TABLE X ALTER COLUMN y" updated to use the keyword "COLUMN".


src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using FluentMigrator.Builders;
+using FluentMigrator.Builders.Create.Constraint;
+using FluentMigrator.Builders.Create.Table;
+using FluentMigrator.Builders.Insert;
+using FluentMigrator.Infrastructure;
+
+namespace FluentMigrator.Postgres
+{
+    public static class PostgresExtensions
```
fubar-coder: A comment is missing

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        public static string IdentityGeneration => "PostgresIdentityGeneration";
+
```
fubar-coder: A comment is missing here too

src/FluentMigrator.Extensions.Postgres/PostgresGenerationType.cs
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace FluentMigrator
+{
+    public enum PostgresGenerationType
```
fubar-coder: It's self-expanatory, but a comment would be nice

src/FluentMigrator.Runner.Postgres/Generators/Postgres/IPostgresGenerator.cs
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace FluentMigrator.Runner.Generators.Postgres
+{
+    public interface IPostgresGenerator : IMigrationGenerator
```
fubar-coder: It's good that you introduced a new interface for a PostgreSQL-specific generator, but a comment is missing here too.

EduAlbSantos: Comment added, as requested. If you find other issues, please let me know. Thank you!

src/FluentMigrator.Extensions.Postgres/PostgresGenerationType.cs
EduAlbSantos: Comment added, as requested.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.cs
EduAlbSantos: Comment added, as requested.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.cs
EduAlbSantos: Comment added, as requested.

src/FluentMigrator.Runner.Postgres/Generators/Postgres10/Postgres10Column.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Postgres10Column"/> class.
+        /// </summary>
+        /// <param name="quoter">The Postgres quoter.</param>
+        /// <param name="typeMap">The Postgres type map.</param>
+        public Postgres10Column([NotNull] PostgresQuoter quoter, ITypeMap typeMap)
+            : base(quoter, typeMap)
+        {
+        }
```
Sohra: In my branch I added to this file, a private `FormatAlterIdentity` method, which I added to the the base 
class's `AlterClauseOrder` protected list here in the constructor, so I could make use of the base class implementation of `GenerateAlterClauses` without override, to achieve the new extension methods to Add/Set/Drop/Drop If Exists identity generation on existing columns.

src/FluentMigrator.Runner.Postgres/Generators/Postgres10/Postgres10Column.cs
```diff
@@ -0,0 +1,91 @@
...
+                    clauses.Add(string.Format("ALTER COLUMN {0} {1}", Quoter.QuoteColumnName(column.Name), columnClause));
+            }
+
+            return string.Join(", ", clauses.ToArray());
+        }
+
+        /// <inheritdoc />
+        protected override string FormatType(ColumnDefinition column)
+        {
+            if (!column.Type.HasValue)
```
Sohra: For anyone else looking at this PR, this method needs to be overridden since PostgresColumn implements Identity columns by changing the type to `serial` or `bigserial` something that obviously isn't appropriate for a Postgres 10 Identity column definition.  This override here duplicates the definition of `ColumnBase.FormatType` which is skipping the base `PostgresColumn.FormatType` and going up to it's base class definition.

Rather than that duplication, I simply exposed it in PostgresColumn like so:
```
        protected string ColumnBaseFormatType(ColumnDefinition column)
        {
            return base.FormatType(column);
        }
```
So that I could call `base.ColumnBaseformatType` in this class and avoid the code duplication.  I couldn't think of a decent name for it, but I'm a big fan of avoiding code duplication, particularly from an ongoing maintenance point of view.  It's too easy to make a change or bugfix in one place and not be aware of all the other duplicated copies of that code that ought to be updated similarly, so if we can avoid that whenever possible, I think it's good practice to do so.

src/FluentMigrator.Runner.Postgres/Processors/Postgres/PostgresProcessor.cs
```diff
@@ -71,6 +71,23 @@ public class PostgresProcessor : GenericProcessorBase
             _quoter = new PostgresQuoter(pgOptions);
         }
 
+        public PostgresProcessor(
```
Sohra: I never found myself needing to add a constructor such as this, could you please explain to me it's purpose?
One thing I did add but I don't see here, was an addition to `PostgtesrunnerBuilderExtensions.cs` such as:
```
        /// <summary>
        /// Adds Postgres 10.0 support
        /// </summary>
        /// <param name="builder">The builder to add the Postgres-specific services to</param>
        /// <returns>The migration runner builder</returns>
        public static IMigrationRunnerBuilder AddPostgres100(this IMigrationRunnerBuilder builder)
        {
            builder.Services
                .AddScoped<Postgres100Processor>()
                .AddScoped<IMigrationProcessor>(sp => sp.GetRequiredService<Postgres100Processor>())
                .AddScoped<Postgres100Generator>()
                .AddScoped<IMigrationGenerator>(sp => sp.GetRequiredService<Postgres100Generator>());
            return builder.AddCommonPostgresServices();
        }
```

How do you configure Postgres10* components to be used?  Do you not use the extension methods like this and other platforms provide?

jzabroski:
@Sohra Sorry I missed your comment. It sounds like you should submit a PR for https://github.com/Sohra/fluentmigrator/tree/feature/postgres10identity

Thanks also for all your detailed replies.  I think your approach is better, and you bring up a lot of points I've noticed but haven't had a chance to address.  Your help would be greatly appreciated in simplifying some of these choices as you seem to have a good eye for detail.

Issue #1061, opened 2019-09-11
Moving to a different CI/CD service


fubar-coder:
We're currently having problems with Travis-CI and are thinking about moving to a different service.

Some options are:

- GitHub Actions
- Azure Pipelines

What we need is:

- Build
- Test
  - Integration tests with different databases
- Deploy

I'm currently testing GitHub Actions and @eloekset volunteered to test Azure Pipelines.

Comments? Ideas? Alternatives?

eloekset:
I created a new feature branch at my fork to start testing Azure Pipelines CI. I based it on the develop branch, but it doesn't build. Don't we use the develop branch? Should I rebase on master instead, which builds fine locally?

eloekset:
The first build ran successfully in 2m 36s.
![image](https://user-images.githubusercontent.com/791721/64728371-16fb3900-d4db-11e9-919d-fd6e57bb095e.png)

Tests took 18s and 4,285 out of 4,715 succeeded. 430 tests were skipped, and those are the integration tests, I guess. Next up is to test docker images with the tools needed to run integration tests.
![image](https://user-images.githubusercontent.com/791721/64728509-65a8d300-d4db-11e9-9d72-6c85fcc1b3cd.png)

I created a new DevOps organization for this test, so I can invite others if anyone is interested. The project URL is: https://dev.azure.com/githubeloekset/fluentmigrator/

jzabroski:
Wow, this is great.

jzabroski:
So... I like Azure DevOps.  It's only obvious weakness is it doesn't support lifecycles.  But for an open source project, I don't see a need for lifecycles.  Our lifecycle is, quite literally, the build pipeline.

eloekset:
Sorry, but I haven't had time to do much the last two weeks. But hopefully I'll catch up again this weekend. I started copying the step of the Travis yaml file, since I assume that's what's used for NuGet release builds today?

eloekset:
BTW this [blog post](https://dev.to/itnext/the-ultimate-free-ci-cd-for-your-open-source-projects-3bkd) about the ultimate CI/CD for open source popped up in my newsfeed somewhere this week. Also GitHub Actions is mentioned.

BigYellowHammer:
I know it's a different topic, but what do you think guys about connecting some static code analysis tool to FM codebase?
Recently I was playing a bit with it and was able to connect Azure Pipelines with SonarCloud.
Here are the results for latest master
[...]

PR #1105, opened 2019-10-27
Fix triggers of build pipelines

eloekset:
A second attempt at fixing the build pipelines triggers based on the research of fubar-coder in  https://github.com/fluentmigrator/fluentmigrator/issues/1061#issuecomment-546590293


eloekset:
Looks like it works even before the yml files are merge to master. 🤔 but 🎉

Issue #1123, opened 2019-12-03
dotnet fm rollback all does not work


gandarez:
Running the command below gives me an error

```bash
dotnet fm rollback all -p mysql -c "Server=localhost;Database=Db;Uid=root;Pwd=xpto;" -a MyAssmbly.DbMigration\bin\Debug\netcoreapp2.1\MyAssmbly.DbMigration.dll --tag Testing
```

```
Unrecognized option '-p'
```

Am I missing something?

jzabroski:
Looks like a bug

jzabroski:
@gandarez Just for triage, are you on the latest version?

gandarez:
I think so - `Tool 'fluentmigrator.dotnet.cli' (version '3.2.1') was successfully installed.`

gandarez:
Also for other commands like `migrate up`, `migrate down`, `rollback to` and `rollback by`

![Capture](https://user-images.githubusercontent.com/782854/70151931-46ed4f00-167a-11ea-842a-8410e7bf5cb1.PNG)

gandarez:
I asked them about chaining SubCommands and the answer is `Yes`

https://github.com/natemcmaster/CommandLineUtils/issues/310

jzabroski:
I think the way these commands are written probably needs refactoring.  For my taste, it doesn't make a lot of sense that we effectively have three pieces of glue code for the old-school CLI, the dotnet-fm tool, and MSBuild task.  It makes consistency very hard to ensure.

gandarez:
You mean istead of calling `dotnet fm rollback all` we should call `dotnet fm rollback-all` or `dotnet fm rollback-to` and so on... maybe?

jzabroski:
No, I mean we are using Mono Options in one place, Nate's Command Line Utils in another, etc.  It complicates things by having three different ways to enter the Runner

fubar-coder:
A rollback command doesn't support assemblies and/or tags. This seems like an oversight on my side. The `-p` and `-c` options are connection options and come directly after the `rollback` command.

Example:

```text
dotnet fm rollback -p mysql -c "Server=localhost;Database=Db;Uid=root;Pwd=xpto;" all
```

gandarez:
> A rollback command doesn't support assemblies and/or tags. This seems like an oversight on my side. The `-p` and `-c` options are connection options and come directly after the `rollback` command.
> 
> Example:
> 
> ```
[...]

Issue #1033, opened 2019-06-14
MigrateUp vs MigrateDown


chrsin:
Is there a reason for the need to specify the whole migrate up and migrate down?
From a release management standpoint it would be much simpler to just run a "migrate to this specific version". (or just migrate to the latest version
At the moment i need to be aware if what I am releasing is a rollback or just a normal release.

I took a look at the code for the MigrationRunner and as far as i can tell. It should be possible to do the filtering, sorting and decision on which migrations to run and what to run on them.
However does anyone know if it introduces problems if we run down operations and Up operations in the same migration?

chrsin:
I guess i could write my own executable and have a look at if there is anything that is being downgraded with the specified version. Then do a migrateDown if there is. Otherwise do a MigrateUp

fubar-coder:
I can only think of two reasons why this differentiation exists:

- It was easier to implement (practical)
- Preferring explicit expression of intent (migrate-up/down vs migrate-to, philisophical)

BUT: I'm not the original maintainer, so I cannot say for sure why it was done this way.

@jzabroski What do you think? Would this make sense to have a `MigrateTo` function?

jzabroski:
@chrsin I think there is some education awareness to explain the vocabulary and what it means, and how your workflow should conform to the vocabulary.

The MigrateUp/MigrateDown conventions come from Martin Fowler's _Patterns of Enterprise Architecture_ book.  They exist because that is the industry wide vocabulary for thinking about upgrading and rolling back databases.  In practice, I don't write `Down()` and only use `ForwardOnlyMigration`.  The main reason is I get paid to deliver features.  When I worked on algorithmic trading systems, we did have to specify a Down()/Rollback script, but, in those environments, the database administrators always wanted the upgrade and rollback scripts to be pure SQL so that they could peer review it before GoLive.  So, in environments where I have had to write rollback scripts, I also could not use a tool like FluentMigrator.

In terms of expression of intent, I think Up/Down and Version numbers is correct.  If there is an example of a competing project (perhaps in another language like Ruby) where they are doing things differently and innovating, I think we should take a look.  For now, I think we have definitely built a better mouse trap.  For context, I used to contribute a tiny bit to RoundhousE database migration framework, and fully believe FluentMigrator is superior, especially if you know C# - it's not even close.

Let me know if you have any further questions for me.

jzabroski:
@fubar-coder I guess this code snippet is one alternative way to think of things:

```ruby
[...]

PR #1130, opened 2019-12-07
fix subcommands

gandarez:


jzabroski:
It looks like 3 tests may have failed, but it might just be that Tear Down failed for some reason?

gandarez:
The pipeline threw an error

```
X CallingConstraintExistsReturnsFalseIfTableDoesNotExistWithSchema [71ms]
2019-12-07T17:45:35.2665346Z   Error Message:
2019-12-07T17:45:35.2665588Z    TearDown : System.IO.IOException : The directory is not empty.
2019-12-07T17:45:35.2665801Z   Stack Trace:
2019-12-07T17:45:35.2666022Z   --TearDown
2019-12-07T17:45:35.2666237Z    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
2019-12-07T17:45:35.2666466Z    at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound, WIN32_FIND_DATA& data)
2019-12-07T17:45:35.2666709Z    at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
2019-12-07T17:45:35.2666946Z    at FluentMigrator.Tests.Integration.Processors.SqlServerCe.SqlServerCeConstraintTests.TearDown() in d:\a\1\s\test\FluentMigrator.Tests\Integration\Processors\SqlServerCe\SqlServerCeConstraintTests.cs:line 167
```

Possible solution?
https://stackoverflow.com/questions/329355/cannot-delete-directory-with-directory-deletepath-true

gandarez:
https://stackoverflow.com/a/18204128/833531

gandarez:
Locally worked
Screen Shot 2019-12-03 at 21 28 02

gandarez:
@jzabroski do you plan to approve this PR?

fubar-coder:
I guess that it works locally, because you don't run (by default) the SqlServer CE tests? The post-integration cleanup stuff is annoying, which is the reason we want to use docker.

I'd approve this anyway, because this PR doesn't change anything that might've caused this problem.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -25,17 +25,20 @@ namespace FluentMigrator.DotNet.Cli.Commands
     public class ConnectionCommand : MigrationCommand
     {
         [Option("-c|--connection <CONNECTION_STRING_OR_NAME>", Description = "The name of the connection string (falls back to machine name) or the connection string itself to the server and database you want to execute your migrations against.")]
-        public string ConnectionString { get; }
+        public string ConnectionString { get; set; }
+
+        [Option("-m|--transaction-mode <MODE>", Description = "Overrides the transaction behavior of migrations, so that all migrations to be executed will run in one transaction.")]
+        public TransactionMode TransactionMode { get; set; }
```
fubar-coder: This is a good change and I'd like to keep this in the PR.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -25,17 +25,20 @@ namespace FluentMigrator.DotNet.Cli.Commands
     public class ConnectionCommand : MigrationCommand
     {
         [Option("-c|--connection <CONNECTION_STRING_OR_NAME>", Description = "The name of the connection string (falls back to machine name) or the connection string itself to the server and database you want to execute your migrations against.")]
-        public string ConnectionString { get; }
+        public string ConnectionString { get; set; }
```
fubar-coder: The setter isn't needed by the command line parser/processor library.

src/FluentMigrator.DotNet.Cli/Commands/Migrate.cs
```diff
@@ -23,13 +23,9 @@ namespace FluentMigrator.DotNet.Cli.Commands
 {
     [HelpOption]
     [Command("migrate", Description = "Apply migrations")]
-    [Subcommand(typeof(MigrateUp))]
-    [Subcommand(typeof(MigrateDown))]
+    [Subcommand(typeof(MigrateUp), typeof(MigrateDown))]
```
fubar-coder: This change is nice too. Keep it.

src/FluentMigrator.DotNet.Cli/Commands/RollbackAll.cs
```diff
@@ -25,8 +25,8 @@ namespace FluentMigrator.DotNet.Cli.Commands
     [Command("all", Description = "Rollback all migrations")]
     public class RollbackAll : BaseCommand
     {
-        public Rollback Parent { get; }
-
+        public Rollback Parent { get; set; }
+
```
fubar-coder: Trailing whitespace

src/FluentMigrator.DotNet.Cli/Commands/Root.cs
```diff
@@ -20,16 +20,13 @@ namespace FluentMigrator.DotNet.Cli.Commands
...
-    [Subcommand(typeof(Validate))]
-    [Subcommand(typeof(ListCommand))]
+    [Subcommand(typeof(Migrate), typeof(Rollback), typeof(Validate), typeof(ListCommand))]
     public class Root
     {
         protected int OnExecute(CommandLineApplication app, IConsole console)
         {
             console.Error.WriteLine("You must specify a subcommand.");
-            app.ShowHelp();
+            app.ShowHelp();
```
fubar-coder: Trailing whitespace

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
jzabroski: @fubar-coder My only thought is that the "mode" is a change from the switches provided by the Console Runner which uses `--transaction-per-session`.  https://fluentmigrator.github.io/articles/runners/runner-console.html#--transaction-per-session---tps-optional

Perhaps the short-hand should be `-tm`.

To be clear, I think specifying a transaction mode is way better than a switch, as its future proof.  As i've mentioned in the past, I want to add support for monadic transactions.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -25,17 +25,20 @@ namespace FluentMigrator.DotNet.Cli.Commands
...
+        [Option("-m|--transaction-mode <MODE>", Description = "Overrides the transaction behavior of migrations, so that all migrations to be executed will run in one transaction.")]
+        public TransactionMode TransactionMode { get; set; }
 
         [Option("--no-connection", Description = "Indicates that migrations will be generated without consulting a target database. Should only be used when generating an output file.")]
         public bool NoConnection { get; }
 
         [Option("-p|--processor <PROCESSOR_NAME>", Description = "The kind of database you are migrating against. Available choices can be shown with \"list processors\".")]
         [Required]
-        public string ProcessorType { get; }
+        public string ProcessorType { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -25,17 +25,20 @@ namespace FluentMigrator.DotNet.Cli.Commands
...
 
         [Option("-s|--processor-switches <PROCESSOR_SWITCHES>", Description = "Provider specific switches.")]
-        public string ProcessorSwitches { get; }
+        public string ProcessorSwitches { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -44,15 +47,15 @@ public class ConnectionCommand : MigrationCommand
         public bool Verbose { get; }
 
         [Option("--profile <PROFILE>", Description = "The profile to run after executing migrations.")]
-        public string Profile { get; }
+        public string Profile { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -44,15 +47,15 @@ public class ConnectionCommand : MigrationCommand
...
 
         [Option("--context <CONTEXT>", Description = "Set ApplicationContext to the given string.")]
-        public string Context { get; }
+        public string Context { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -44,15 +47,15 @@ public class ConnectionCommand : MigrationCommand
...
 
         [Option("--timeout <TIMEOUT_SEC>", Description = "Overrides the default database command timeout of 30 seconds.")]
-        public int Timeout { get; }
+        public int Timeout { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
```diff
@@ -44,15 +47,15 @@ public class ConnectionCommand : MigrationCommand
...
 
         [Option("-o|--output=<FILENAME>", CommandOptionType.SingleOrNoValue, Description = "Output generated SQL to a file. Default is no output. A filename may be specified, otherwise [targetAssemblyName].sql is the default.")]
-        public (bool, string) Output { get; }
+        public (bool, string) Output { get; set; }
```
jzabroski: Remove set;

src/FluentMigrator.DotNet.Cli/Commands/Migrate.cs
jzabroski: I don't 100% agree.  @gandarez  Please see https://github.com/fluentmigrator/fluentmigrator/issues/1033 for how I envisioned supporting sub-commands for up and down.  The annoying part is MigrationCommand has --start-version.  I'd like to alias --start-version to --minimum-version.  @fubar-coder  Thoughts?

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
gandarez: In the official examples only boolean properties has only `getter`.

https://github.com/natemcmaster/CommandLineUtils/blob/master/docs/samples/subcommands/inheritance/Program.cs

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
gandarez: In the official examples only boolean properties has only `getter`.

https://github.com/natemcmaster/CommandLineUtils/blob/master/docs/samples/subcommands/inheritance/Program.cs

src/FluentMigrator.DotNet.Cli/Commands/Migrate.cs
gandarez: Why this change is related to that issue @jzabroski ?

src/FluentMigrator.DotNet.Cli/Commands/RollbackAll.cs
gandarez: fixed

src/FluentMigrator.DotNet.Cli/Commands/Root.cs
gandarez: fixed

gandarez:
Fixed trailling whitespaces and gave comments in each specific topic

src/FluentMigrator.DotNet.Cli/Commands/Migrate.cs
fubar-coder: Yes, we should discuss the matter of `--minimum-version` vs `--start-version` in a different issue.

BTW: Just a hint for future contributions: Don't introduce style changes that don't directly contribute to the problem the PR tries to solve. This will only complicate the review process and makes it less likely that the PR will get merged.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
fubar-coder: You can see in the [README.md](https://github.com/natemcmaster/CommandLineUtils/blob/1fefc7348f1007d32f5378f180b1cd45926664c6/README.md) that all command line argument properties are missing their setters.

fubar-coder:
@gandarez Just to clarify: Adding setters, combining attributes and moving the `TransactionMode` property up in the class hierarchy does definitely solve issue #1123? I'm puzzled about this...

src/FluentMigrator.DotNet.Cli/Commands/Migrate.cs
jzabroski: Yes -- I started looking into whether Mcmaster.commandlineutils supports aliases.  Even if it doesnt, probably an easy PR to merge into his project. But a bit more yak shaving.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
gandarez: Should I follow the README or the examples? I'm confused and I think even they are.

gandarez:
@fubar-coder that's the way I solved the problem and also I applied some standards following the guideline provided by CommandLineUtils.

jzabroski:
I think exposing only a getter is ideal here as it doesn't make sense to
set these values.

src/FluentMigrator.DotNet.Cli/Commands/ConnectionCommand.cs
jzabroski: I think the README is the right way. The bottom line is it doesnt make sense to expose a setter unless you need to for testing purposes.  Even so, you probably are testing things incorrectly if you need a setter to configure your objects (even though I love doing this).

jzabroski:
Hi @gandarez Thanks again for your help.  Things are quieting down at work into the Holidays, so I hope to look more at the docker compose approach.  I have also been learning more about docker and didn't realize Docker supported building directly from a GitHub repo (and even a GitHub PR).  Curious if you've ever given that a shot?  I'm also curious what's better:

### Traditional CI approach
a) downloading git source into build server
b) building dockerfile with a volume mounted for the downloaded source directory, which contains WORKDIR and COPY commands to hoist the mounted source code into the container for manipulation by the build tools

### Completely Dockerized approach
a) building dockerfile from GitHub tag/pr/branch
b) dockerfile in turn contains both runtime pack and targeting pack layers

paulmarshall:
Can this get completed soon - rollback all doesn't work !

PR #1137, opened 2019-12-24
Ensures VersionLoader disposes the migration processor

thibault-reigner:
### Changes
Makes IMigrationRunner and IVersionLoader inherits IDisposable
 
### Explanation
Fixes an issue in VersionLoader, which was not disposing the migration processor.
As a result, the connection contained in GenericProcessorBase was not properly disposed, leading to a connection leak.
NB : the migration runner, containing both a migration processor and a version loader has to dispose them too.

### DIFF
--- a/src/FluentMigrator.Runner/ConnectionlessVersionLoader.cs
+++ b/src/FluentMigrator.Runner/ConnectionlessVersionLoader.cs
@@ -203,5 +203,10 @@ namespace FluentMigrator.Runner
                 new KeyValuePair<string, object>(VersionTableMetaData.DescriptionColumnName, description)
             };
         }
+
+        public void Dispose()
+        {
+            _processor.Dispose();
+        }
     }
 }
--- a/src/FluentMigrator.Runner/IMigrationRunner.cs
+++ b/src/FluentMigrator.Runner/IMigrationRunner.cs
@@ -26,7 +26,7 @@ namespace FluentMigrator.Runner
     /// <summary>
     /// The migration runner
     /// </summary>
-    public interface IMigrationRunner : IMigrationScopeStarter
+    public interface IMigrationRunner : IMigrationScopeStarter, IDisposable
     {
         /// <summary>
         /// Gets the migration processor used by this runner
--- a/src/FluentMigrator.Runner/IVersionLoader.cs
+++ b/src/FluentMigrator.Runner/IVersionLoader.cs
@@ -16,12 +16,14 @@
 //
 #endregion
 
+using System;
+
 namespace FluentMigrator.Runner
 {
     /// <summary>
     /// Manages the version table and the stored versions
     /// </summary>
-    public interface IVersionLoader
+    public interface IVersionLoader : IDisposable
     {
         /// <summary>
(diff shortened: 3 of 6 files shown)


jzabroski:
@thibault-reigner Do you have a failing test to demonstrate the issue?  Can you also provide details like what version of FluentMigrator you're using, which database system+version?  Are you using anything fancy like connection impersonation?  Are you using FluentMigrator.Console, FluentMigrator.MSBuild, FluentMigrator.DotNet.Cli, or the In-Process Runner?  If you're using the In-Process Runner, what's your DI configuration like?  Why can't the problem be solved a different way?

While I can believe the problem you are running into is that you are leaking connections, I don't understand how the fix is to sprinkle `IDisposable` everywhere.  `VersionLoader` itself does not manage any external resources.  When it gets finalized, its object graph will also get finalized, which will cause `Dispose` to be called and let go of the IDbConnection.

Note, also, that `IDbConnection` is not your typical `Dispose`-able resource: It doesn't actually dispose of anything, necessarily.  If you're using connection pooling (which you almost certainly are), it _signals_ the Connection Pool that it no longer needs the remote resource.  The connection itself is not necessarily disposed of.  On SQL Server, especially SQL Server 2014 and later, [it causes `sp_reset_connection` to be called internal to the pool](https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/using-integrated-authentication?view=sql-server-ver15#tracking-access-to-a-database).  It's also better to not `Dispose` of the object if it's going to be re-used again, as that round-trip to the connection pool can [cause contention calling sp_reset_connection](https://blogs.msdn.microsoft.com/psssql/2014/03/03/sp_reset_connection-rate-usage-dont-fight-over-the-grapes/).

thibault-reigner:
@jzabroski I agree adding `IDisposable` everywhere is not a good solution...
Regarding my use case, (very twisted one I confess) it is in a unit testing framework, where I need to ensure at some point there is no active connections to the database, and to do so, I am clearing the pool.
Of course, as the connection in the processor may still be opened, as despite of the lifetime of `VersionLoader` being over, there is no guarantee the GC (hence the finalizer) has ran. The GC being non-deterministic, this issue happens from time to time.
I guess it is a matter of convention here, is the process injecting the process in `VersionLoader` still responsible of disposal of this processor, or does `VersionLoader`, by getting full ownership, also has to ensure disposal ?
I will review in detail how the `VersionLoader` is instantiated in my code to check how easy it is to ensure the disposal of the processor at a higher level.

jzabroski:
> I guess it is a matter of convention here, is the process injecting the process in `VersionLoader` still responsible of disposal of this processor, or does `VersionLoader`, by getting full ownership, also has to ensure disposal ?

There is a third option (and possibly a fourth option here):

Option 3: Don't focus on disposal, focus on closing the connection when it's no longer needed.  (I didn't say this in my initial reply, because it's a lot of code reading for even me and it's holiday season and I was hoping to fix bugs, not just read through code.)

Option 4: Implement first-class connection management through `Func<T>` and tunnel the connection accessor into whatever T needs a connection.

It's Christmas eve, though, so I'll punt on any more brainpower to this problem for today.  Sorry :)  Happy Holidays.

Issue #1139, opened 2019-12-26
Champion: <RollForward>MajorVersion</RollForward>


jzabroski:
Rather than continuously update TargetFrameworks in our csproj, we can specify `<RollForward>MajorVersion</RollForward>` in the csproj, instead.  This will override the default configuration of the .NET Runtime for FluentMigrator users.

The advantages are:
1. Faster build times, as we won't be building/linking against multiple target frameworks, just net461 and .netcoreapp2.2
2. We only add target frameworks to the  `<TargetFrameworks>` property in the csproj if a third party dependency forces our hand.  (_I'm not sure why this would happen, but Murphy is the enemy of any CI/CD build/deploy process._)
3. As we begin to roll out end-to-end integration tests with docker, adding more TargetFrameworks will significantly increase the run-time of those integration tests.
4. Smaller .nupkg and .s.nupkg uploads to nuget.org - friendlier to AzureDevOps hosting our dlls as well, as we will consume less space potentially.

The disadvantages are:
* Not easily discoverable.  We will get Issues about how to run this on .netcoreapp3.0
* .NET Runtime doesn't guarantee (regression test) MajorVersion rollforward will work in all scenarios, but it should work.
    * Biggest risk might be if you're loading a dll which has `TypeForwardTo` behaviors.  Does ReSharper's "Go To Implementation" (and similar tools) correctly resolve this and provide spy of the rollforward dll implementation?
        * If a Profiling tool is not using the Profiling API, they might not handle `TypeForwardTo` behaviors correctly.  But this has always been a risk.

For testing, we would:
* Test with the latest framework, as-is reasonable, and just assume that the underlying TargetFrameworks work as opposed to testing against the older runtime.  This makes the job of FluentMigrator's maintainers easier, as all they have to do is make sure the CI/CD builds pass on newer frameworks, and aren't responsible for pushing newer packages with each runtime MajorVersion update.

#### Action Items
- [ ] Update [FluentMigrator.DotNet.Cli csproj](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.DotNet.Cli/FluentMigrator.DotNet.Cli.csproj) to add `<RollForward>Major</RollFoward>`
- [ ] Update [Sample application csproj](https://github.com/fluentmigrator/fluentmigrator/blob/master/samples/FluentMigrator.Example.Migrator/FluentMigrator.Example.Migrator.csproj) to add `<RollForward>Major</RollFoward>`
[...]

Issue #1016, opened 2019-04-17
External runners are unreliable


fubar-coder:
The external runners (Migrate.exe, dotnet fm, msbuild) are unreliable due to several problems around different target frameworks (net461, netstandard1.3, netcoreapp2.0, etc.) and .NET implementations (Framework, Core, Mono).

The current implementation is very naive and just tries to load the assemblies and its dependencies into the current context.

# Situations

The following situations can be found:

## Executed from project directory

This would probably require us to use the same method, which requires the user to specify the startup project and the target project.

The following things are of interest for us:

- `Microsoft.EntityFrameworkCore.Design` (required by `dotnet-ef`)
- `dotnet-ef`

## Executed from `bin/` directory

This will only work when all dependencies are available, so I suggest that we don't support this use case.

## Executed on published output

Maybe we can use [`DotNetCorePlugins`](https://github.com/natemcmaster/DotNetCorePlugins)?

# Tools

Every tool should support executing the migrations from the published output and only `dotnet-fm` and `msbuild` tasks should support executing migrations from source code.

jzabroski:
Thinking out loud: Is there a way we can use PrivateAssets part of the new MSBuild fluent PackageReference project element to somehow force-include the references into the output path?

jzabroski:
Solution should also address https://github.com/fluentmigrator/fluentmigrator/issues/989 , in that, today, if you have an Execute.WithConnection command that references SqlConnection instead of DbConnection, it will fail to resolve the System.Data.SqlClient dll reference due to not being in the output path.

fubar-coder:
I guess that we have to (somehow) utilize msbuild, probably together with the [C#/VB.NET/F# project system](https://github.com/dotnet/project-system).

jzabroski:
Couple of thoughts here based on some fast reading today, that I'd like to share.  This might not solve our problems, but is in general good to know as it is part of the overall "knobs we can turn":

1. Forcing all `PackageReference`s in a .NET SDK project file to be included in the bin directory:
    - Example:
    ```xml
    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
    ```
    - Known issues: https://github.com/dotnet/sdk/issues/933 (fixed but not released yet I think)
[...]

PR #1141, opened 2019-12-31
Removed database driver implementations from Runner

ydemetriades:
Removed the following dependencies from FluentMigrator.Runner
- Db2
- Firebird
- MySql
- Oracle
- Postgres
- Redshift
- SqlAnywhere
- SQLite
- SqlServer
- SqlServerCe
- Hana

Removed obsolete dependencies classes and tests

When installing FluentMigrator.Runner, it downloads all database drivers implementation nuget packages and thus polluting the resulting published application with dll's that not being used.

If you would like to keep the classes and tests I am willing to create another library that contains the obsolete classes just to keep the backwards compatibility.


jzabroski:
@ydemetriades Don't you think a change like this should go in the develop branch which targets 4.0?  When we release 3.2.2 in the next few days/weeks, this would break everyone who is dependent on this packaging behavior.  If you want to curtail the published dlls, you can do so by customizing your pack command.

That said, we're also talking about saving only 421Kb here.

samples/FluentMigrator.Example.Migrator/FluentMigrator.Example.Migrator.csproj
```diff
@@ -25,5 +25,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\FluentMigrator.Runner\FluentMigrator.Runner.csproj" />
     <ProjectReference Include="..\FluentMigrator.Example.Migrations\FluentMigrator.Example.Migrations.csproj" />
+    <ProjectReference Include="..\..\src\FluentMigrator.Runner.SQLite\FluentMigrator.Runner.SQLite.csproj" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
```
jzabroski: I'm not sure what you are trying to do here with the example project?

jzabroski: To clarify, the example Migrator csproj demonstrates the ideal future end-state for FLuentMigrator that we're inching towards in v4.  We ran into various technical challenges in v3, mostly due to under-documented APIs and design decisions in .NET Core SDK:

* It wasn't clear from a project management perspective whether .netstandard was like a framework or merely an API specification
* It wasn't clear how to run old assemblies on newer runtimes (See #1139 )
* There is magic with how `.deps.json` determines what assemblies get loaded into the default AssemblyLoadContext.  See #1016 - Using an in-process runner pretty much guarantees a correct `.deps.json` file gets generated.

ydemetriades:
Sorry for not responding earlier. So great news! Fluent Migrator version 3 is live! Great work! 
It seems that the main purpose of my Pull Request, which was to decouple the Implementations (eg. SqlServer, Db2) from the FluentMigrator.Runner was not solved. Have a look [here](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj#L15).

Please allow me some time to catch up with the changes and commit some more changes!
@jzabroski Leave me a comment if you need any further clarifications! 😎

jzabroski:
> Sorry for not responding earlier. So great news! Fluent Migrator version 3 is live! Great work!
> It seems that the main purpose of my Pull Request, which was to decouple the Implementations (eg. SqlServer, Db2) from the FluentMigrator.Runner was not solved. Have a look [here](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Runner/FluentMigrator.Runner.csproj#L15).
> 
> Please allow me some time to catch up with the changes and commit some more changes!
> @jzabroski Leave me a comment if you need any further clarifications! 😎

I guess I'm not clear on what problem you want to solve for FM v3.  As I mentioned, the additional runner packaging costs only 421 Kb.  Is there a problem you're running into?  Otherwise, let's push this change for FM v4, as it's a breaking change for consumers.  The `develop` branch is where you want to push to.

Issue #1152, opened 2020-01-17
Problem with create table with custom type column


slad08:
Hello.My name is Denis. 
I've have some problem.I'll need create table via fluentmigrator.Below show sql script,for create this table.
I use SQL Server 2014, FluentMigrator v3.2 . 
In code  i also use  .AddSqlServer2014()
All information with scripts and code i attached 
[Problem with create table via fluentMigrator all scripts.txt](https://github.com/fluentmigrator/fluentmigrator/files/4077077/Problem.with.create.table.via.fluentMigrator.all.scripts.txt)
Could you help me?

P.S. 
If i deleted word NULL manually in script , what generated via fluentMigrator, when i describe customer types(in attache item 3 i mean DurationSecs and StartTimeDays columns) and run in MSSQL2014 management studion , this scrtipt working good.
How can i point in code, what don't need generated word NULL or NOT NULL(i mean item 3 in attached columns DurationSecs and StartTimeDays).
Not work:
[DurationSecs]  AS (DATEDIFF(SECOND, [StartTime], [EndTime])) PERSISTED NULL, 
[StartTimeDays]  AS (DATEDIFF(DAY,'19900101', [StartTime])) NULL)
if deleted null, work
[DurationSecs]  AS (DATEDIFF(SECOND, [StartTime], [EndTime])) PERSISTED, 
[StartTimeDays]  AS (DATEDIFF(DAY,'19900101', [StartTime])) )

jzabroski:
I think the issue might be with the `NULL` keyword, not the custom type column, per say.

That being said, as a code review, I suggest you not make computed columns that depend on your server's datefmt settings, as '19900101' does.  I'm surprised SQL Server allows this , but it will definitely fail if you add PERSISTED to the end of this.

@fubar-coder  Do we have any support for computed columns?  It looks like we don't.

@slad08  As a workaround, first create the table without the computed columns, and then use an extension method to add the computed columns to the table:

```c#
public static class ColumnExtensions {
  public static void ComputedColumn(this Migration up, string columnName, string tableName, string schemaName, string computedColumnExpression, bool persisted)
  {
    this.Execute.Sql($@"ALTER TABLE [{schemaName}].[{tableName}] ADD [{columnName}] AS ({computedColumnExpression}){persisted ? " PERSISTED;" : ";"}");
  }
}
```

slad08:
Thank's @jzabroski for you help. You answer helped me.

slad08:
[...]

PR #1155, opened 2020-01-21
1152 Issue Support Computed Columns

slad08:
I proposed decision problem from Issue 1152
https://github.com/fluentmigrator/fluentmigrator/issues/1152

### DIFF
--- a/src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2014Column.cs
+++ b/src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2014Column.cs
@@ -1,0 +1,31 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using FluentMigrator.Model;
+
+namespace FluentMigrator.Runner.Generators.SqlServer
+{
+    internal class SqlServer2014Column : SqlServer2008Column
+    {
+        public SqlServer2014Column(ITypeMap typeMap, IQuoter quoter)
+             : base(typeMap, quoter)
+        {
+            ClauseOrder.Add(FormatSparse);
+        }
+        protected override string FormatNullable(ColumnDefinition column)
+        {
+            if (column.IsNullable.HasValue == false && column.Type == null && !string.IsNullOrEmpty(column.CustomType))
+            {
+                return string.Empty;
+            }
+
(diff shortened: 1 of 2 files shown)


jzabroski:
I'm pretty sure even SQL Server 2008 supports computed columns, so why is this being introduced here as a Sql Server 2014 feature?

In terms of returning string.Empty, is that the right change?  I think it might be, but it feels like we're co-opting two different meanings:

1. No null behavior mentioned (in which case, it should _probably_ default to `NULL`)
2. If user uses AsCustom and doesn't specify null behavior, we're assuming we shouldn't display `NULL`.

What if instead of using column.CustomType, we instead introduced a new dimension, ComputedColumn?  This would have the nice effect of, depending on the database, automatically generated the boilerplate for the computed column.  For example, in Firebird, the syntax is:

```sql
alter table nums alter b generated always as (4*a + 7);
```

so you don't need "` generated always as (`" and the closing "`);`"

jzabroski:
Thanks for your help, by the way.

slad08:
I answering for you questions:
   "In terms of returning string.Empty, is that the right change? I think it might be, but it feels like we're    co-opting two different meanings:
    1.No null behavior mentioned (in which case, it should probably default to NULL)
    2.If user uses AsCustom and doesn't specify null behavior, we're assuming we shouldn't display NULL."
My answeres:
SqlServer2014Column are overriding method FormatNullable for SqlServer2005Column and control Nullable property for this Server2014. By default if no null behavior mentioned  and column Customer then nothing show, if NULL mentioned and column Customer - written NULL, or NOT NULL if mentioned Not Null. Such algorithm i meant.
I use now SQL server 2014 and i haven't opportunity for testing previous versions SQL server.
Maybe you have some own opinion about this?

jzabroski:
I need to re-review this this week, but it probably won't make 3.2.7 release.  I left a comment on #1152:

> [`SET ANSI_NULL_DFLT_ON`](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-ansi-null-dflt-on-transact-sql?view=sql-server-ver15) Modifies the behavior of the session to override default nullability of new columns when the ANSI null default option for the database is false.

@slad08 

>Maybe you have some own opinion about this?

I should have asked for tests on this but I also think I need to re-think what the right way to do this is.

PhenX:
Closed by #1896

Issue #1171, opened 2020-02-17
On Insert : "ORA-01704: string literal too long"


PhenX:
When inserting string via `Insert.Into` on Oracle, we are facing this issue when a string value is over 4000 chars : 
https://stackoverflow.com/questions/13945710/error-ora-01704-string-literal-too-long

jzabroski:
@PhenX Thanks - I think this StackOverflow post ([oracle10g - How to insert/update larger size of data in Oracle tables](https://stackoverflow.com/a/8802397/1040437)) answered by Sathyajith Bhat better explains how we'll need to update the generated SQL.

In your link, I found these notes very helpful, too:

# Workaround 1: [concatenate sequences of 4000 byte literals using || concatenation operator](https://stackoverflow.com/a/30219281/1040437)
> ```sql
> Insert into table (clob_column) values ( to_clob( 'chunk 1' ) || to_clob( 'chunk 2' ) );
> ```

# Gotcha 1: [Handling Special Characters](https://stackoverflow.com/a/31794976/1040437)

> The split work until 4000 chars depending on the characters that you are inserting. If you are inserting special characters it can fail. The only secure way is to declare a variable.

My Note: I believe what this person is saying is that Workaround #1 works until you deal with special characters.

# Proposed Solution - [Combines Workaround 1 and Gotcha 1](https://stackoverflow.com/a/59469966/1040437)

> ```sql
> DECLARE
>   chunk1 CLOB; chunk2 CLOB; chunk3 CLOB;
> BEGIN
>   chunk1 := 'very long literal part 1';
>   chunk2 := 'very long literal part 2';
>   chunk3 := 'very long literal part 3';
> 
>   INSERT INTO table (MY_CLOB)
>   SELECT ( chunk1 || chunk2 || chunk3 ) FROM dual;
> END;
> ```

fairking:
I have the same issie too. Waiting for the release. Appreciate your hard work guys.

jzabroski:
@fairking I merged the fix , release coming this week.

Issue #1129, opened 2019-12-05
Docker images for various databases


jzabroski:
* *Azure Cosmos DB*: https://github.com/Azure/azure-cosmos-db-emulator-docker
* *SQL Server Linux*: https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-configure-docker?view=sql-server-ver15
* Oracle
    * *Oracle Database Express Edition 18.4+*: https://github.com/oracle/docker-images/tree/master/OracleDatabase/SingleInstance - according to Avi Miller, Product Owner of Oracle, Oracle XE 18.4 supports docker.
    * *Oracle Database Express Edition 11g*: https://github.com/wnameless/docker-oracle-xe-11g - community provided, ⚠️ read the README.md! The package was renamed. This dockerfile is used by [quartznet](https://github.com/quartznet/quartznet/blob/139aafa23728892b0a5ebf845ce28c3bfdb0bfe8/docker/oracle/Dockerfile), a popular job scheduling framework⚠️ 
* *MySQL*: https://hub.docker.com/_/mysql
* *MariaDB*: https://hub.docker.com/_/mariadb
* *Postgres*: https://hub.docker.com/_/postgres
* *DB2*: https://hub.docker.com/_/db2-developer-c-edition

In addition, JetBrains' DataGrip team has open sourced a bunch of their continuous integration infrastructure.

jzabroski:
Updated original post to add links to various images or emulators.  Slow progress, but it's happening.

PhenX:
Hello, just in case you are still looking for images for various DBMS, I just found JetBrains has its own set of docker images here for their Datagrip product, certainly for testing purpose : 
https://registry.hub.docker.com/u/datagrip
I didn't try them yet though

jzabroski:
Amazing find! Thanks, Fabien!

PhenX:
For your information I started working on it with Rider here : https://github.com/fluentmigrator/fluentmigrator/compare/master...PhenX:feature/docker

![image](https://user-images.githubusercontent.com/42170/94198050-f23fec80-feb6-11ea-8bb4-f2974e84292f.png)

jzabroski:
@PhenX  Wow, great job.  It looks like you didn't mean to `git add .idea`, based on your `.gitignore` file.  Maybe you updated the gitignore file after?  If you can, maybe squash those commits locally so its easier to follow the changes you made to get all this working?

[...]

PR #1196, opened 2020-03-23
Fix large string literals oracle

PhenX:
Fixes issue #1171


src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
```diff
@@ -15,13 +15,40 @@
...
 
 using FluentMigrator.Runner.Generators.Generic;
 
 namespace FluentMigrator.Runner.Generators.Oracle
 {
     public class OracleQuoterBase : GenericQuoter
     {
+        private const int MaxStringLength = 3900;
+
+        private static IEnumerable<string> SplitBy(string str, int chunkLength)
```
jzabroski, line 29: Is there no .NET library function that does this?

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
```diff
@@ -15,13 +15,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
 
 using FluentMigrator.Runner.Generators.Generic;
 
 namespace FluentMigrator.Runner.Generators.Oracle
 {
     public class OracleQuoterBase : GenericQuoter
     {
+        private const int MaxStringLength = 3900;
```
jzabroski, line 27: Consider adding a comment explaining where 3900 comes from.

test/FluentMigrator.Tests/Unit/Generators/Oracle/OracleDataTests.cs
```diff
@@ -11,6 +14,94 @@ public class OracleDataTests : BaseDataTests
...
...ssssssssssssss. Mouse is good you understand your place in my world or hide head under blanket so no one can see. Meow in empty rooms i is not fat, i is fluffy or howl on top of tall thing eat a rug and furry furry hairs everywhere oh no human coming lie on counter don't get off counter, so meow meow so pet me pet me don't pet me, and if human is on laptop sit on the keyboard. I will ruin the couch with my claws gnaw the corn cob and plan steps for world domination. Sit on human get video posted to internet for chasing red dot always ensure to lay down in such a manner that tail can lightly brush human's nose relentlessly pursues moth and being gorgeous with belly side up, meow meow we are 3 small kittens sleeping most of our time, we are around 15 weeks old i think, i don’t know i can’t count. Good now the other hand, too find empty spot in cupboard and sleep all day yet chase the pig around the house and eat prawns daintily with a claw then lick paws clean wash down prawns with a lap of carnation milk then retire to the warmest spot on the couch to claw at the fabric before taking a catnap leave dead animals as gifts. Eat all the power cords poop in litter box, scratch the walls.
```
jzabroski, line 20: This is impressive... @fubar-coder did you ever expect a PR with this in it? :)

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
jzabroski: @PhenX Actually, I'm wondering why this value isn't 4000, based on our discussion on this issue and comments from StackOverflow indicating 4000?

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
jzabroski: I think we should use a version of string.Substring that works for Unicode.  See my research here: https://stackoverflow.com/a/61125710/1040437 and see: https://stackoverflow.com/a/8959444/1040437 (I've edited this answer so you might need to wait 24 hours for moderator approval to fix a minor nit with the answer).

PhenX: I didn't find any, but you seem to have found one, though, related to another review you made, I think we must split the string so that each chunk does not exceed 4000 bytes. Do you think it's possible to achieve this with `SubstringByTextElements` ?

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
PhenX: Yep, that's 4000, I chose to keep a margin of 100 for multibyte characters 😄

PhenX: But I think I might have made a mistake : I can't seem to find any report of the error "PLS-00172: string literal too long" for NCLOB columns, I'll have to check, I'm currently making a Docker image with Oracle 12C.
By the way, is everything OK at your side ? In France, it could be worse ...

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
jzabroski: Ah!  I see the issue now.  I was reading MaxStringLength as characters, forgetting that it's a byte length limitation.  Shoot!  Good, good point.

I think it is possible to achieve this with SubstrinByTextElements - we just need a rolling accumulator variable that tracks how many bytes we've been passed from `StringInfo.ParseCombinigCharacters`.  Do you follow my explanation?

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
jzabroski: Following up from other PR comment: Let's rename `MaxStringLength` to `MaxStringLengthInBytes`. When I first read this, after not thinking about the problem for a month, I forgot the issue is related to byte length, not character length.  My brain for some reason just assumed we were only thinking about characters here.  My comments about string splitting were actually split-brain: I was thinking about splitting BMP characters by accident, not thinking about the definition of 4000 being bytes.

jzabroski, line 27:
```c#
        private const int MaxStringLengthInBytes = 4000;
```

jzabroski: > By the way, is everything OK at your side ? In France, it could be worse ...

I have been WFH from end of February / early March (had to make two trips into the office).  I live in the suburbs so I am much less at risk for community spread disease.  But it sucks not being able to see friends.  I am through the hard part of adjusting.  I bought Scrabble this week to play with my wife so we don't lose our minds.  If you have any suggestions on how to keep sane, you can find me on gitter.im in the [FluentMigrator chat](https://gitter.im/FluentMigrator/fluentmigrator) :)

jzabroski: @PhenX One last thing - I updated #1129 to include a link to the semi-official Oracle 11xe docker image, if that makes your testing easier.  (I don't really understand g, c, i suffixes on Oracle releases but I imagine if you get something working on 11xe, it's likely to work on 19xe)

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoterBase.cs
PhenX: Hello, I've been able to test this pull request on a Oracle 12.1 server, and fixed a few things : 
- MaxStringLength is now 4000
- As quotes are doubled during the escape sequence, we have to take this into account
- Oracle is OK with multibyte strings, so we don't have to worry about couting bytes or not (for CLOB and NCLOB ...)

Also, I've implemented the base code for EndToEnd integration tests on OracleManaged and a test for this pull request.

I'll tell you when it's ready, because I have to update the other unit tests as the test string changed to include Unicode chars

jzabroski: Thanks, @PhenX - I'll try to merge it into 3.2.7 as opposed to 3.3.0, since its likely your tests won't impact my other changes I'm working on.

jzabroski:
@PhenX I will release this this week as part of 3.2.7.  Just waiting on the previous build to complete before i click merge :)

Issue #1194, opened 2020-03-20
Infrastructure Index (such as Unique Index on VersionInfo table) is not using IConventionSet


rodrigoramos:
I have customized my FluentMigrator project using IConventionSet but it doesn't seems to be considered when creating the infrastructure object (e.g. VersionInfo table's indexes).

**Example**

I have the following class correctly configured in my project:
``` C#
public class ProjectConstraintNameConvention : IConstraintConvention
{
    // Rest of code was omitted

    private static string GetConstraintName(ConstraintDefinition expression)
    {
        var sb = new StringBuilder();
        sb.Append("MYProject_");
        sb.Append(expression.IsPrimaryKeyConstraint ? "PK_" : "UC_");


        sb.Append(expression.TableName);
        foreach (var column in expression.Columns)
        {
            sb.Append('_').Append(column);
        }

        return sb.ToString();
    }
}
``` 

Running "Up" database, the following SQL is create:
``` SQL
CREATE UNIQUE INDEX UC_Version ON VersionInfo (Version ASC);
```

Instead of:

``` SQL
CREATE UNIQUE INDEX MYProject_UC_VersionInfo_Version ON VersionInfo (Version ASC);
```

I'm using version FluentMigrator/3.2.1

tgharold:
Is this a .NET Core project where you are using `AddFluentMigratorCore()`?  

If so, you need to register the `IConventionSet` *after* making the  `AddFluentMigratorCore()` call in `Startup.ConfigureServices()`.  See [this sample for an example](https://github.com/tgharold/OptionsPatternValidation/blob/c4fcc6c5a1aaebcfac3c642c7d7c939bc5c7e408/examples/OptionsPatternMvc.Dapper/Extensions/ServiceCollectionExtensions.cs#L33-L44).

The reason for this (I think) is a line inside of `AddFluentMigratorCore()`.  Because the IServicesCollection is LIFO (last in first out), if `AddFluentMigratorCore()` registers something to handle `IConventionSet` after you register it... the FM registration will trump your registration.  The workaround is to register your `IConventionSet` after the FM registration.

Source:

https://github.com/fluentmigrator/fluentmigrator/blob/27eb0110ff14b4d32aed7d316ee5caf4dda20716/src/FluentMigrator.Runner/FluentMigratorServiceCollectionExtensions.cs#L101-L102

jzabroski:
[...]

PR #1210, opened 2020-04-09
AddFluentMigratorCore() - Only register default IConventionSet if none registered yet

tgharold:
#1194

Use `TryAddScoped()` instead of `AddScoped()`.  (Unfortunately, the method is not chainable.)

Assumptions:

1. The `AddFluentMigratorCore()` call should not register an additional `IConventionSet` if one has already been registered by the calling program in the `IServicesCollection`.
2. The FluentMigrator code uses `GetService/GetRequiredService` and is pulling only the last registered `IConventionSet` out of the DI/IoC `ServiceProvider`.

Note: If the caller registers their `IConventionSet` *after* the call to `AddFluentMigratorCore()`, there will be two copies of `IConventionSet` registered; but the caller's `IConventionSet` is what will be returned by  `GetService/GetRequiredService`. This is to be expected and is normal.

Refs:
- [TryAddScoped in the MS documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#service-registration-methods)
- [stackoverflow answer](https://stackoverflow.com/a/48187842)

### DIFF
--- a/test/FluentMigrator.Tests/Unit/Initialization/ConventionSetTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Initialization/ConventionSetTests.cs
@@ -67,12 +67,8 @@ namespace FluentMigrator.Tests.Unit.Initialization
             Assert.AreEqual(null, expr.SchemaName);
         }
 
-        /// <summary>Demonstrate that AddFluentMigratorCore() always registers a class to handle
-        /// IConventionSet; even when one has already been registered in the services collection.
-        /// Because the ServiceCollection is LIFO, the last one registered is the one that gets
-        /// used.</summary>
         [Test]
-        public void RegisterOnlyOneConventionSet()
+        public void RegisterOnlyOneConventionSetIfRegisteredBeforeAddCall()
         {
             const string schemaName = "RegTestSchemaA";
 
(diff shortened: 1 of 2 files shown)


test/FluentMigrator.Tests/Unit/Initialization/ConventionSetTests.cs
```diff
@@ -67,12 +67,8 @@ public void NotLoadingConventionSetFromAssemblyWithScanForMigrations()
             Assert.AreEqual(null, expr.SchemaName);
         }
 
-        /// <summary>Demonstrate that AddFluentMigratorCore() always registers a class to handle
-        /// IConventionSet; even when one has already been registered in the services collection.
-        /// Because the ServiceCollection is LIFO, the last one registered is the one that gets
-        /// used.</summary>
```
jzabroski: @tgharold This change looks good but did you mean to delete this comment?

tgharold: Yes, the first commit was to demonstrate failure.  Later commits made that comment obsolete.

Issue #1184, opened 2020-02-26
Analyzer that warns of duplicate MigrationAttribute version values


benmccallum:
Let me know if you think there's value in this. I know it'll blow up when it is attempted to run, and most people will hit that in development before committing their changes, but an analyzer could prevent the build entirely so they have to put in a new version.

I'd be happy to implement this as I'd be keen to write an analyzer as I never have before.

jzabroski:
Hi @benmccallum 

I'm always pro people learning and trying new things.  I think its a welcome improvement.

One thing I never learned how to do with Roslyn analyzers that might be a good "V2" version of such an analyzer would be to point it towards a target database.  For example, we use Red Gate SQL Clone to create a DatabaseName_DEV continuous integration database shared amongst all developers.  It effectively gives developers a sneak preview that somebody else has "taken" the migration number. It would be nice if analyzers could be configured to take into account some per-solution or per-project settings file, like:

```json
{
  "Server": "mydb.db.company.com",
  "Database": "mydb_DEV",
}
```

In terms of analyzers on attributes, I think Johan Larsson has written a couple for Enum Flags and so forth.  He and I discussed those ideas ~2 years ago.  He is pretty active on Gitter in the Reflection Analyzers gitter channel.

Good luck!  Slow and steady!  Keep learning!

benmccallum:
Cool, good to see there's some interest and I also like your ideas about checking against the target DB as a v2 if possible.

Can't promise I'll get to this any time soon, but will have a crack when I get time.

jzabroski:
In addition to an analyzer, it would be nicer for there to be a Fixer that checks if there is a gap to move the duplicate "up" or "down" one version number.

For example:

```c#
[Migration(1)]
public class M1_AddTable : ForwardOnlyMigration
{
  public void Up()
  {
  }
}

[Migration(1)]
public class M1_DuplicateMigrationVersion : ForwardOnlyMigration
{
  public void Up()
  {
  }
}
```

The Fixer would pop up with suggestions to:
1. Move M1_AddTable Migration 1 to 0.
2. Move M1_AddTable Migration 1 to 2.
3. Move M1_DuplicateMigrationVersion 1 to 0.
4. Move M1_DuplicateMigrationVersion 1 to 2.

I've never written a Fixer so I'm not sure how complicated the UI can get and what the right requirements here are, but just brainstorming in the hopes others improvise on this idea.

benmccallum:
[...]

PR #1240, opened 2020-05-02
feat: Analyzers

benmccallum:
fixes: #1184


jzabroski:
I need to add a nuget.config file to the solution to get this to pull in Microsoft.CodeAnalysis.CSharp.Analyzer.Testing as a dependency. @schambers You mentioned wanting some low hanging fruit - @benmccallum wrote this PR awhile back and it's pretty cool what it does.  Can you add a nuget.config with the Roslyn package source?

See: https://github.com/dotnet/roslyn-sdk/tree/master/src/Microsoft.CodeAnalysis.Testing#azure-packages-feed-for-prerelease-packages for the updated nuget.config instructions - in September, Microsoft killed off MyGet and moved to Azure Dev Ops package hosting.

benmccallum:
This was unfortunately left pretty raw as I was stuck trying to get some form of test runs happening to TDD it but I'm sure with a bit of massaging and newer versions of things it'd be fixable and then you could start actually writing an analyser :p

I think the general core analyser helpers that I wrote are valuable; I was using xunit's analyser stuff as heavy inspiration.

jzabroski:
I hope to give it a shot soon.  I think the key thing is to add a nuget.config so that whoever wants to build on top of what you did, can.  I have had some more ideas for analyzers since you wrote this PR, so it was a good start to get us thinking in the direction of cleaner code and more static analysis.

jzabroski:
I made some progress editing your PR.

I have not pushed any changes yet.  There are some things about Roslyn that I don't quite understand, like this `todo [jaz]` note I created:

```c#
        internal override void AnalyzeCompilation(CompilationStartAnalysisContext compilationStartContext, FluentMigratorContext fluentMigratorContext)
        {
            compilationStartContext.RegisterSymbolAction(symbolContext =>
            {
                var typeSymbol = (ITypeSymbol)symbolContext.Symbol;
                var migrationVersion = GetMigrationAttributeVersion(fluentMigratorContext, typeSymbol.GetAttributes());
                if (migrationVersion == null)
                {
                    return;
                }

                var otherTypeNames =
                    symbolContext.Compilation.Assembly.TypeNames.Where(typeName => typeName != typeSymbol.MetadataName).ToList();
                foreach (var otherTypeName in otherTypeNames)
                {
                    var ns = symbolContext.Symbol.ContainingNamespace.Name;
                    // todo [jaz]: Fix me. I don't think this is the right way to get the namespace, but without this, the tests don't pass at all right now,
                    // because GetTypeByMetadataName requires a fully qualified type name.
                    var otherTypeSymbol = symbolContext.Compilation.GetTypeByMetadataName($"{ns}.{otherTypeName}");
```

jzabroski:
Made some more progress discussing with the maintainer of GuAnalyzers some best practices.
Still not clear about a few things, but have enough info to hopefully push forward with something I can commit soon.

benmccallum:
Awesome, sorry I've been AWOL, took a couple of weeks off. Happy to chip in when/where I can :)

jzabroski:
I haven't been able to make progress over the holidays due to trying to wrap up things.  My current contract ends in February so its top priority to finish strong.  If you're interested I can push what I have to a branch on jzabroski/FluentMigrator.

Issue #1242, opened 2020-05-06
How to remove sensitive information (ex: passwords) from the output


andrecarlucci:
Hi guys,

I have one migration to create a user on Postgres:

```
var pass = Environment.GetEnvironmentVariable("GENERATED_PASSWORD");
Execute.Sql($"CREATE USER bob WITH PASSWORD '{pass}';");
```
The output of this migration would be:
```
-------------------------------------------------------------------------------
1: _001_Change_MasterUser migrating
-------------------------------------------------------------------------------
Beginning Transaction
ExecuteSqlStatement CREATE USER bob WITH PASSWORD 'jw8s0F4dkjskfjkjsdfahu38**3e;;d';
Committing Transaction
1: _001_Change_MasterUser migrated
```
We use this for our pipelines, we let the developers create databases on-demand in lower environments but we take control of the password for the production ones.

For that reason, we don't want the new password to be exposed in the log files.

Is there a way to achieve this?

**Expected benefit**
Sensitive information is not exposed in log files.

Tks!

pergardebrink:
I have a related question on this. I have some migrations I generate from our legacy sql files and run them with Execute.Sql() and the output becomes cluttered with all SQL statements.

As I can see, the issue is that ExecuteSqlStatementExpression overrides ToString() that concatenates the statement, which is then logged. https://github.com/fluentmigrator/fluentmigrator/blob/e82aafa20e6dbe3cefa221303fe23cf8bf59fffd/src/FluentMigrator.Abstractions/Expressions/ExecuteSqlStatementExpression.cs#L45
I would prefer if it would be possible to display an optional custom more descriptive message. Something like this?

```
public override void Up() 
{
    Exeucte.Sql("UPDATE Something SET x=1", "Update Something");
}
```

jzabroski:
@andrecarlucci First, Great to see you decided to join FluentMigrator as a community member.  I'm glad I reached out.  Sorry for the delays in responding - COVID-19 lockdown really slows down life with things like buying grocercies, etc.  

Second, I think you bring a good point.  I have not had any GDPR requirements but this is definitely huge for GDPR customers.  And, on my project the last 3 years, we use SQL Server so we just manage users as Windows AD Groups, so I have not hit password creation issues.

@pergardebrink So, here's my thoughts on adding a Description field to override outputting SQL:

1.  Technically it shouldn't break anyone API wise unless they're using some weird reflection and expecting only a single method called Sql.
[...]

PR #1243, opened 2020-05-09
Add description to sql statement expression

pergardebrink:
This is a suggested change to solve https://github.com/fluentmigrator/fluentmigrator/issues/1242

Technically I think it might be a breaking change on assembly level, haven't verified that but can retarget to develop branch in case the proposed solution is ok.

### DIFF
--- a/src/FluentMigrator.Abstractions/Builders/Execute/IExecuteExpressionRoot.cs
+++ b/src/FluentMigrator.Abstractions/Builders/Execute/IExecuteExpressionRoot.cs
@@ -35,6 +35,13 @@ namespace FluentMigrator.Builders.Execute
         void Sql(string sqlStatement);
 
         /// <summary>
+        /// Executes an SQL statement
+        /// </summary>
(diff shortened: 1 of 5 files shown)


jzabroski:
As discussed in the other thread, please create an integration test that demonstrates this doesn't break [`--no-connection`](https://fluentmigrator.github.io/articles/runners/dotnet-fm.html#--no-connection) and [`--preview`](https://fluentmigrator.github.io/articles/runners/dotnet-fm.html#--preview) used in tandem to allow developers to save FM scripts out to a file for DBAs to review, and then I am all for this.

pergardebrink:
I created an integration test which I think is what you asked for. I have only just recently started using fluent migrator, but think I understand what the options you mention does. 

The SqlScriptFluentMigratorLoggerProvider with ShowSql = true and PreviewOnly setting is what I found is the equivalent of the settings you mentioned?

jzabroski:
Off the top of my head, I think that's correct.  Will review it this weekend.  There are a couple of PRs including one for Oracle I need to merge.

PR #1261, opened 2020-06-02
add schemaName to sql for create index expressions

emejibka:
I add schemaName to create index and create constraint expressions.
Our case:
1. we run migration in oracle with system user
2. we create index or constraint in migration in user schema
bug №1 - index created in system schema
bug №2 - IndexExists query does not work because index located in system schema

### DIFF
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleGenerator.cs
@@ -271,6 +271,26 @@ namespace FluentMigrator.Runner.Generators.Oracle
             return WrapInBlock(alterColumnWithDescriptionBuilder.ToString());
         }
 
+        public override string Generate(CreateIndexExpression expression)
+        {
+            var indexColumns = new string[expression.Index.Columns.Count];
+
+            for (var i = 0; i < expression.Index.Columns.Count; i++)
+            {
+                var columnDef = expression.Index.Columns.ElementAt(i);
+
+                var direction = columnDef.Direction == Direction.Ascending ? "ASC" : "DESC";
+                indexColumns[i] = $"{Quoter.QuoteColumnName(columnDef.Name)} {direction}";
+            }
+
+            return string.Format(CreateIndex
+                , GetUniqueString(expression)
+                , GetClusterTypeString(expression)
+                , Quoter.QuoteIndexName(expression.Index.Name, expression.Index.SchemaName)
+                , Quoter.QuoteTableName(expression.Index.TableName, expression.Index.SchemaName)
+                , string.Join(", ", indexColumns));
+        }
+
         public override string Generate(InsertDataExpression expression)
         {
             var columnNames = new List<string>();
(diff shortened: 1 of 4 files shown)


src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2005Generator.cs
```diff
@@ -228,7 +228,7 @@ public override string Generate(CreateIndexExpression expression)
                 CreateIndex,
                 GetUniqueString(expression),
                 GetClusterTypeString(expression),
-                Quoter.QuoteIndexName(expression.Index.Name),
+                Quoter.QuoteIndexName(expression.Index.Name, expression.Index.SchemaName),
```
jzabroski, line 231: @emejibka Are you sure this is correct?  In SQL Server, the schema name of the index is inferred by the table's schema.  Indexes don't exist outside of a table schema, so there is no concept of an index name consisting of a name + schema.  I can't speak to Postgres or Oracle or the other DBs, but this is likely wrong for SQL Server.

jzabroski:
@emejibka I am pretty confused here by this PR.  What bug are you running into?

See: https://www.postgresql.org/docs/9.1/sql-createindex.html

Quoting the documentation for index name (emphasis mine):
> `name`
> 
> The name of the index to be created. **No schema name can be included here; the index is always created in the same schema as its parent table.** If the name is omitted, PostgreSQL chooses a suitable name based on the parent table's name and the indexed column name(s).

See: https://firebirdsql.org/file/documentation/reference_manuals/fblangref25-en/html/fblangref25-ddl-idx.html

No mention of schema.  Only indexname.

See: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver15#syntax-for-sql-server-and-azure-sql-database

Microsoft publishes a Backus-Naur Form grammar for each statement, making it very unambiguous that you can't put a schema in the `index_name` slot.

Quote:
> `index_name`
> 
> Is the name of the index. Index names must be unique within a table or view, but do not have to be unique within a database. Index names must follow the rules of identifiers.

⚠️  Oracle, however, does appear to support assigning a schema to an index.

See: https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_5010.htm

> image

emejibka:
We have some issue with oracle. We run migration with system user and create index expression create index in system schema but we need index in table owner schema.
My mistake - i don`t check postgres documentation.

How we can add schema name in create index expression only in oracle ?

jzabroski:
I would cancel this PR, remove the changes to Postgres and SQL Server, and squash/rebase your commit tree so that it doesnt include the history related to changing Postgres as it will confuse readers looking at history in the future.

Alternatively, if you're not a git pro, and dont feel comfortable squash/rebase, I'd just copy-paste your changes into a fresh branch off master and don't copy the Postgres and SQL Server stuff, or anything non-Oracle related.

We're getting closer to having live integration tests for each db but we're not there yet.

jzabroski:
@PhenX Any chance you can review this change? It looks good to me, but I don't have Oracle installed. Thanks.

emejibka:
@jzabroski 
I can give you access to test database.
Can you accept request?

jzabroski:
@emejibka Thanks for the offer, but I should probably try to set-up Docker for Oracle18xe .  The following Oracle blog post looks promising: https://blogs.oracle.com/oraclemagazine/deliver-oracle-database-18c-express-edition-in-containers

emejibka:
Hello.
How are you?
Do you install oracle?

jzabroski:
@emejibka Hi - I did not install Oracle yet.  I will prioritize trying to get docker up and running.

Issue #1265, opened 2020-06-09
OracleException ORA-01704 string literal too long


fairking:
**Describe the bug**
Still got the same error after upgrading to v 3.2.7

**To Reproduce**
Table:
![image](https://user-images.githubusercontent.com/13495631/84169877-0b8c6c80-aa71-11ea-8bde-c18cb4c61f3d.png)
Migration:
![image](https://user-images.githubusercontent.com/13495631/84170294-6de56d00-aa71-11ea-9776-c75caff532aa.png)
File content (file size 12,739 bytes):
![image](https://user-images.githubusercontent.com/13495631/84170419-979e9400-aa71-11ea-87e0-4504b3498149.png)
Chunks (total 4 chunks):
![image](https://user-images.githubusercontent.com/13495631/84171687-06c8b800-aa73-11ea-9509-dde64f77416b.png)

**Expected behavior**
I assume it must split the text into chunks with up to 4000 characters each. Based on the size it must be at least 5 chunks but I can see only 4.

**Information (please complete the following information):**
 - OS: Windows 10 Pro
 - Platform: .NET Core 3.1
 - FluentMigrator version: 3.2.7
 - FluentMigrator runner: None
 - Database Management System: Oracle
 - Database Management System Version: Oracle 18xe

**Additional context**
I tried to get the actual query from the terminal logged by migrator. Just pasted one of the chuncks into the Notepad++ and it says the length is 4729.
![image](https://user-images.githubusercontent.com/13495631/84170981-2e6b5080-aa72-11ea-9f09-44e887bb2e34.png)

Is it an issue with how the length is calculated or it is my fault?

fairking:
Files to debug: [http://www.filedropper.com/completedticketsbypriority](http://www.filedropper.com/completedticketsbypriority)

fairking:
Update: If I remove `Encoding.UTF8` from `File.ReadAllText` it runs the script with no issues.

jzabroski:
@PhenX Any idea why UTF encoding would change the behavior here?  I'm sure there is a logical explanation but I am blanking as to why.  I did find this issue on Stack overflow: https://stackoverflow.com/questions/59851455/utf-8-oracle-entity-framework-core

jzabroski:
@fairking If you use less than 4000 characters in the file, does the problem still happen?  I wonder if this is not at all related to Fabien's change but some low-level issue in the Oracle driver.

fairking:
[...]

PR #1266, opened 2020-06-09
Fix of #1265 OracleException ORA-01704 string literal too long

fairking:
Fix of #1265 
The issue was with end line characters. I have also added TABs just in case. Please feel free to review or edit if necessery.


jzabroski:
So these are Oracle escape Characters, not C# escape Character, right? In other words the code was doing:
```sql
'Foo\' +
'nBar'
```

fairking:
I am not sure. I did a quick fix and it works, but now I am starting to realize more:
It looks like it wasn't escaping end lines but adding extra 2 spaces to the left.
Why it was working with current changes?
Let see we got:
ABC{\r}{\n} (total 5 characters) based on my pr we must reduce the number of characters to 2
Result:
{space}{space}ABC{\r}{\n} (total 7 characters) (so two extra spaces are reduced by {\r}{\n} we subtracted.

Do you remeber anything we do with the left padding in the project?

Before I was thinking it was something related to TABS (replacing with spaces) but I looked at my file and didn't find any.

jzabroski:
Take your time.  My point is more to do with whether the EscapeCharacters collection should include all characters. https://docs.oracle.com/cd/B28359_01/appdev.111/b28370/literal.htm#LNPLS01326

> A character literal is an individual character enclosed by single quotes (apostrophes). Character literals include all the printable characters in the PL/SQL character set: letters, numerals, spaces, and special symbols. PL/SQL is case sensitive within character literals. For example, PL/SQL considers the literals 'Q' and 'q' to be different. For more information, see Character Literals.
> 
> A string literal is a sequence of zero or more characters enclosed by single quotes. The null string ('') contains zero characters. A string literal can hold up to 32,767 characters. PL/SQL is case sensitive within string literals. For example, PL/SQL considers the literals 'white' and 'White' to be different.
> 
> To represent an apostrophe within a string, enter two single quotes instead of one. For literals where doubling the quotes is inconvenient or hard to read, you can designate an escape character using the notation q'esc_char ... esc_char'. This escape character must not occur anywhere else inside the string.

Other stuff: It appears Oracle supports changing the quote character.  https://livesql.oracle.com/apex/livesql/file/content_CIREYU9EA54EOKQ7LAMZKRF6P.html

fairking:
Got some news
I digged into the rabbit hole and found that the issue is caused by `Oracle.ManagedDataAccess.Client.OracleCommand`. If I copy a query from the command it runs in the sql manager with no issues, but the step into the command execution throws an exception `string literal too long`:
image

fairking:
I am giving up. I cannot figure out how to trace the failed query in the profiler. Those spaces I was mentioned in the above comment could be related to the terminal. But `.replace(" ", "")` definitely worked. So I have no idea what is going on.

jzabroski:
@fairking Thanks, I will definitely pick this up at some point this week.  Initially Fabien did have 3900 characters.  I think you've provided enough info for me to figure out the issue.

In the mean time, is your UTF8 workaround a viable workaround?  Thanks

fairking:
What do you mean UTF8 ? My current workaround is `.Replace(" ", "")` but it changes the content of the original document so I assume it is temporary.

jzabroski:
I'm going to try to review this today - yesterday was a super long day at work.

jzabroski:
I haven't forgotten about this PR.  Close to accepting it. I started a code spike using ParseCombiningCharacters to understand how to future proof this using standard APIs.

fairking:
I would do one extra test (not sure it is possible with unit tests but can be done manually). To save the string (eg. my example) and then pull it back from the db and check the string is the same.

fairking:
I did the manual test. Seems it is working. I have included tab and single quotes to make sure everything comes back with the same shape. Feel free to merge it down if you have no complains.
image

fairking:
Hi @jzabroski 
Are you going to include this fix into 3.2.8 ? Cannot see any milestones tagged.
One more thing you can use Squash Commit to make your master branch more clear:
image

jzabroski:
@fairking I don't squash.  Not a believer.

I added the milestone - it's in the non-public draft release notes.  Thank you for catching that.  I'm pretty close to release 3.2.8.  Just refactoring some logic I wrote last week to make it cleaner, then releasing it.

PR #1272, opened 2020-06-10
Add support for Postgres 10 Identity fields

### LINKED ISSUE #1256: PostgreSQL IsIdentity is creating the PK the old way?
**Describe the question**
When I'm creating a new table for my PostgreSQL v12.2 and I set the **IsIdentity** property to **true** for the PK the column is created as `serial NOT NULL`. This is the old way of creating PK in PostgreSQL (before v10) and the new way is `int4 NOT NULL GENERATED BY DEFAULT AS IDENTITY` 

Do you know how can I set this option so the PK can be created by the new way?

 - OS: Windows 10
 - Platform: .NET Core 3.1
 - FluentMigrator version: 3.2.6
 - FluentMigrator runner: "in-process runner"
 - Database Management System: Postgres v12.2

Sohra:
Fixes #1256
I wish to call out #1084 which beat me to it - well done @EduAlbSantos
Unfortunately that PR has not been merged in the comments I was asked to raise this PR.

I apologise this is raised rather hurridly - it's been quite a few months since the two of us initially did our respective work but I just noticed the request for me to raise this PR and I will endeavour to bring this to the state required to accept into the FluentMigrator codebase as and when I have time.

I added my own review to the original PR calling out some of differences in approach I took in this branch so, at least initially, I'll avoid self-reviewing this PR but happy to do so to provide explanatory comments if I am asked.

If accepted, https://fluentmigrator.github.io/#supported-databases should be updated with a new entry for identifier `Postgres100` referring to this [mere beginning of full] Postgres 10 support

If this were my repo, I'd squash the PR into the base branch when merging to keep a clean history, however I don't know what is done in here... let me know if it'd help you if I rebased this work on the latest master to give you some cleaner history if you prefer to fast-forward thee working commits when merge this...

### DIFF
--- a/src/FluentMigrator.Runner.Postgres/Processors/Postgres/Postgres100Processor.cs
+++ b/src/FluentMigrator.Runner.Postgres/Processors/Postgres/Postgres100Processor.cs
@@ -1,0 +1,40 @@
+#region License
+// Copyright (c) 2018, FluentMigrator Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using FluentMigrator.Runner.Generators.Postgres;
+using FluentMigrator.Runner.Initialization;
+
+using JetBrains.Annotations;
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace FluentMigrator.Runner.Processors.Postgres
+{
+    class Postgres100Processor : PostgresProcessor
+    {
+        public Postgres100Processor(
+            [NotNull] PostgresDbFactory factory,
+            [NotNull] Postgres100Generator generator,
+            [NotNull] ILogger<PostgresProcessor> logger,
+            [NotNull] IOptionsSnapshot<ProcessorOptions> options,
+            [NotNull] IConnectionStringAccessor connectionStringAccessor,
+            [NotNull] PostgresOptions pgOptions)
+            : base(factory, generator, logger, options, connectionStringAccessor, pgOptions)
+        {
+        }
+    }
+}
--- a/FluentMigrator.sln
+++ b/FluentMigrator.sln
@@ -77,7 +77,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "packaging", "packaging", "{
 		PackageTool.props = PackageTool.props
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FluentMigrator.Extensions.Oracle", "src\FluentMigrator.Extensions.Oracle\FluentMigrator.Extensions.Oracle.csproj", "{56103516-EDEA-494F-A3FA-0CCDB5348453}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FluentMigrator.Extensions.Oracle", "src\FluentMigrator.Extensions.Oracle\FluentMigrator.Extensions.Oracle.csproj", "{56103516-EDEA-494F-A3FA-0CCDB5348453}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FluentMigrator.Extensions.Postgres", "src\FluentMigrator.Extensions.Postgres\FluentMigrator.Extensions.Postgres.csproj", "{68CA06A0-6BB8-464D-9D7C-2B09B4FAF768}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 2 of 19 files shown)


Sohra:
I am not particularly happy with `Postgres100` myself either, however the reason I opted for that was for consistency with the earlier `Postgres92` - see for example https://github.com/fluentmigrator/fluentmigrator/blob/bfb074b3dd8158b721081950d44b5677f07500a6/src/FluentMigrator.Runner.Postgres/PostgresRunnerBuilderExtensions.cs#L53

That particular provider appears to deviate from other provider patterns, such as `SqlServer` by separating its classes into its own `Postgres92` namespace instead of using just `Postgres` (a pattern I did not copy with my work, I have the `Postgress100*` classes in the `Postgres` namespace like `SqlServer` does with all its different versions.

To change the Postgres92 classes to conform, and perhaps rename to a `Postgres9_2` even would obviously be a breaking change which I can't imagine you'd want to do...

Anyway, I'm happy to go either way... if you want I can push another commit to rename all the `Postgres100*` references to `Postgres10_0*` - just say the word and I'll get to it.

Sohra:
In saying that it looks like the author of the Postgres92 work was following a pattern a bit like the SqlServer database compatibility levels, where SQL Server 2005 (i.e. v9) was DB Compatibility level 90, etc
https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-compatibility-level?view=sql-server-ver15#arguments

But yeah, it's not a pattern followed throughout the FluentMigrator codebase anywhere else that I noticed...

src/FluentMigrator.Runner.Postgres/Processors/Postgres/Postgres100Processor.cs
```diff
@@ -0,0 +1,40 @@
...
+using JetBrains.Annotations;
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace FluentMigrator.Runner.Processors.Postgres
+{
+    class Postgres100Processor : PostgresProcessor
+    {
+        public Postgres100Processor(
```
Sohra, line 29: Oops, I just noticed now all other processor classes override the DatabaseType and DatabaseAliases properties... while this appears to work without them, and I haven't looked into what those properties are used for yet (does anyone else know of the top of their heads?), perhaps I ought to similarly override them in this class?
https://github.com/fluentmigrator/fluentmigrator/blob/bfb074b3dd8158b721081950d44b5677f07500a6/src/FluentMigrator.Runner.Postgres/Processors/Postgres92/Postgres92Processor.cs#L34
https://github.com/fluentmigrator/fluentmigrator/blob/bfb074b3dd8158b721081950d44b5677f07500a6/src/FluentMigrator.Runner.Postgres/Processors/Postgres92/Postgres92Processor.cs#L36

jzabroski:
I think we should rename it to Postgres10_0.  Yes, it's inconsistent, but eventually Postgres 9.2 will fade from mainstream usage.  We should plan for an architecture where we can let people flexibly target the exact database version.

Sorry for the delay.  I was legitimately thinking about the pros and cons.  Below is a breakdown.

# Pros

1. Clear distinction between major and minor versions
2. Avoids possible collisions in the future, e.g., if Postgres v92.0 is released.

# Cons

1. Resharper warnings about names containing underscores. Would require adding the phrase Postgres10_0 to the team-shared dictionary.
2. Contrary to .NET conventions people already know.
    1. .NET seems to have adopted the "agglutinated" MajorMinor numbering convention, e.g, net452, net461, net471, net472
    2. I think this was a mistake by .NET that doesn't get talked about much. It's probably part of why we went from ".NET Core 3.1" to ".NET 5".
2. Breaks pre-existing convention someone chose for Postgres 9.2.
    1. However, as you note, in some places there is already inconsistency due to this being an open source project and people choosing their own personal style.   It would be nice to start to "herd the cats", especially as we start to plan to modularize FluentMigrator to be per-runner.

# Neither Here Nor There
1. Should we change Postgres92 to be Postgres9_2?
    1. For now, no.  We can consider it as part of FluentMigrator v4.0 breaking change.

src/FluentMigrator.Runner.Postgres/Processors/Postgres/Postgres100Processor.cs
Sohra: Upon closer inspection, these are what drive the `MigrationBase.IfDatabase` member.  For that reason I'll override them in the new `Postgres10_0Processor`.  I will specify PostgresSQL10_0 and PostgresSQL as alternate identifiers to the database tyoe Postgres10_0.

I also considered updating the existing `Postgres92Processor` to similarly specify `PostgreSQL` as an _additional_ alternate identifier (much like all the versions of SqlServer use plain `SqlServer` as an alternate so you can target any version by referencing this string without a version), so that it will match expressions that do `IfDatabase("PostgresSQL")` as v9.2 is still PostgresSQL.

Technically would be a breaking change though.  Up until now people could target only the first version of the processor by using `Postgres` or `PostgresSQL` without having 9.2 match... If we made this change then to avoid matching 9.2 they could only use `Postgres`.  Maybe it would be worth saving that change also for a 4.0 release where Postgres92 is renamed to Postgres9_2...?

Sohra:
Sorry for the force pushes there, had a few typos PostgresSQL instead of PostgreSQL which I was fixing in the comment messages and string identifiers.

Sohra:
@jzabroski, I think I've sorted that now, is there anything else I should do on this PR?

jzabroski:
@Sohra I'm re-running the tests now.  We had a problem with the DTAExecutionHost process that I was not able to figure out / didn't have time to investigate.  Let's see if it's working now and was just a goof in ADO.

KrisPetkov:
Hi @jzabroski  will this feature be available in the next release (3.2.8) or it will be postponed for a later one?

jzabroski:
@KrisPetkov My goal is to merge it with a fix for the MSSQL Filtered indexes and Oracle blob issue, and release some time this week (most likely the weekend).

Sohra:
This is great news, and thanks for your help @jzabroski !!

Issue #1290, opened 2020-06-30
SQLite Rename Column


Chandler-Davidson:
SQLite previously did not support column renaming; however, since 2018-09-15 (v. 3.25.0), SQLite has supported it via the ALTER command. See: https://www.sqlite.org/changes.html

We should update the existing RenameColumn generator for SQLite to use the command. I'm unsure how we support versioning of databases, please let me know!

jzabroski:
The way we support versioning is not super-modular, so it makes sense you wouldn't know how we do it.  The most common way is to recurse on the existing Generator by subclassing it and then overriding the virtual method that adds the incremental functionality.  Each new Generator in turn needs a new Processor and ProcessorId (see fluentmigrator.github.io for list of ids).

It would, in many ways, be ideal to allow metadata to be passed around and each individual piece to dispatch on the available metadata.  The downside to doing the metadata approach is then people have to step through more lines of code.

southernprogrammer:
👍

fairking:
[dotnet](https://www.nuget.org/packages/Microsoft.Data.Sqlite.Core/6.0.0-preview.3.21201.2) is considering to update the version of sqlite: https://github.com/dotnet/efcore/issues/24834
Also the [System.Data.SQLite.Core](https://www.nuget.org/packages/System.Data.SQLite.Core/) will be updated very soon: https://system.data.sqlite.org/index.html/doc/trunk/www/news.wiki

PR #1291, opened 2020-06-30
Sqlite rename column

Chandler-Davidson:
Related to #1290 

Let me know if we should add versioning to the SQLite generator.

### DIFF
--- a/samples/FluentMigrator.Example.Migrations/FluentMigrator.Example.Migrations.csproj
+++ b/samples/FluentMigrator.Example.Migrations/FluentMigrator.Example.Migrations.csproj
@@ -18,4 +18,7 @@
   <ItemGroup>
     <EmbeddedResource Include="**/*.sql" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Data.Sqlite" Version="3.1.5" />
+  </ItemGroup>
 </Project>
--- a/samples/FluentMigrator.Example.Migrator/FluentMigrator.Example.Migrator.csproj
+++ b/samples/FluentMigrator.Example.Migrator/FluentMigrator.Example.Migrator.csproj
@@ -18,7 +18,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.3.4" />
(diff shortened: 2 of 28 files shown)


jzabroski:
I think we should just merge this and see if anyone complains about using it on an older version of SQLite, rather than do the version check.  I suspect if you're running SQLite you're upgrading frequently, but I could be wrong.

jzabroski:
@Chandler-Davidson can you fix this test? https://github.com/fluentmigrator/fluentmigrator/pull/1291/files#annotation_275527805

Chandler-Davidson:
@jzabroski, sorry, just fixed the test! I agree about the use of SQLite. If you're moving to use FluentMigrator, you're probably in a position to also upgrade the version of SQLite.

igitur:
This is closed, but I noticed that this PR *adds* the `Microsoft.Data.Sqlite` dependency to some projects that did not have the dependency before. Example : https://github.com/fluentmigrator/fluentmigrator/blob/4bacf0cd2a225d6f36c41931c774c7a52d2cdce6/src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj#L45

Was this intentional? 

I noticed this now when I upgrade my Nuget package and got a whole lot of SQlite packages referenced that don't think I need at all.

jzabroski:
@igitur No, that was not intentional.  I have no idea how I missed that.  I must not have reviewed the last commit.  If you want to submit a PR to remove those obviously wrong dependencies, I'll merge it and release 3.2.9 tonight.

igitur:
Usually I have no problem submitting PRs, but I'll give the author who introduced the error first chance to correct it. Hope that's OK?

jzabroski:
Sure.  @Chandler-Davidson - any interest in fixing this?  The problem seems to be this commit: https://github.com/fluentmigrator/fluentmigrator/pull/1291/commits/12bf2f165b6eca6b469ef8ec2012e4110289513d - A bunch Sqlite packages got added to various projects.

Chandler-Davidson:
@igitur, I'm out of the office for a bit, so go ahead! Sorry for the extra work.

igitur:
Ok, I'll do a PR. Do I base it on `master` or `develop` ?

jzabroski:
`master`.  Thank you.

PR #1322, opened 2020-08-22
fix query in SequenceExists

emejibka:
ALL_SEQUENCES view replaced with USER_SEQUENCES if schema is empty


jzabroski:
@emejibka Thanks for all these awesome Oracle patches.  You have @PhenX have been crushing it.

PR #1324, opened 2020-08-25
Partially revert "Upgrade sqlite version", which added unnecessary Microsoft.Data.Sqlite dependencies to a lot of the projects.

igitur:
This partially reverts commit 12bf2f165b6eca6b469ef8ec2012e4110289513d.

As discussed, this reverts the accidental addition of `Microsoft.Data.Sqlite` dependencies to some projects.

I couldn't figure out how you build the `.nupkg` files in order to verify that the dependencies are dropped in those files too. Will you please double-check that when you release the next version?


jzabroski:
@igitur Thanks, will push 3.2.9 soon, probably after/during lunch break.

PR #1325, opened 2020-08-25
Enhance description to disable flag that has default value = true

igitur:
It's impossible to know how to disable a flag with a default value of `true` in `FluentMigrator.Console` without digging through the source code comments in `Options.cs`. 

This PR adds some text to enlighten the user as to how to disable it. 

At the moment, this is relevant only to the stripping of comments. Currently `--strip` or `--strip-comments` defaults to `true`. In order to disable the stripping of comments, the user will have to pass `--strip-` or `--strip-comments-`.


jzabroski:
@igitur  Thank you

PR #1334, opened 2020-09-08
Fix SQLite message about incorrect double quotes

fvaillancourt:
When migrating a SQLite DB I was getting the following warning out of SQLite
Message: double-quoted string literal: "VersionInfo"
Error code: 28

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Processors/SQLite/SQLiteProcessor.cs
+++ b/src/FluentMigrator.Runner.SQLite/Processors/SQLite/SQLiteProcessor.cs
@@ -44,6 +44,8 @@ namespace FluentMigrator.Runner.Processors.SQLite
     {
         [CanBeNull]
         private readonly IServiceProvider _serviceProvider;
+        [NotNull]
+        private readonly SQLiteQuoter _quoter;
 
         public override string DatabaseType
         {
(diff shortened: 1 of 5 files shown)


fvaillancourt:
Sorry about single quote, show have run tests

jzabroski:
@fvaillancourt I think the SqliteProcesssor should use the SqliteQuoter.

1. Update SqliteProcessorFactory to inject the same instance of the quoter in the generator and the processor, here: https://github.com/fluentmigrator/fluentmigrator/blob/e82aafa20e6dbe3cefa221303fe23cf8bf59fffd/src/FluentMigrator.Runner.SQLite/Processors/SQLite/SQLiteProcessorFactory.cs
2. Update SqliteProcessor to take a SqliteQuoter
3. Remove the string.Replace nastiness with the SqliteQuoter
4. Ensure tests pass
5. Maybe run some integration tests locally to ensure end-to-end tests work for your scenario (no more warnings to the console)

fvaillancourt:
@jzabroski How about that? I think it is close to what you had in mind? I did test it in our app and it does remove the warning SQLite was outputing.

jzabroski:
@fvaillancourt yes, perfect.  Thank you so much! Love when people quickly follow up on reviews.  Puts a pep in my step.

PR #1362, opened 2020-10-19
Add include for index in postgres

lillo42:
Add include for index in postgres
Add support for postgres 11

### DIFF
--- a/src/FluentMigrator.Extensions.Postgres/Builder/Create/Index/CreateIndexExpressionNonKeyBuilder.cs
+++ b/src/FluentMigrator.Extensions.Postgres/Builder/Create/Index/CreateIndexExpressionNonKeyBuilder.cs
@@ -1,0 +1,41 @@
+#region License
+// Copyright (c) 2020, FluentMigrator Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using FluentMigrator.Builders.Create.Index;
+using FluentMigrator.Infrastructure;
+using FluentMigrator.Postgres;
+
+namespace FluentMigrator.Builder.Create.Index
+{
+    internal class CreateIndexExpressionNonKeyBuilder : ICreateIndexNonKeyColumnSyntax
+    {
+        public CreateIndexExpressionNonKeyBuilder(ICreateIndexOnColumnSyntax expression, ISupportAdditionalFeatures supportAdditionalFeatures)
+        {
+            Expression = expression;
+            SupportAdditionalFeatures = supportAdditionalFeatures;
+        }
+
+        public ICreateIndexOnColumnSyntax Expression { get; }
+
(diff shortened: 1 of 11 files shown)


jzabroski:
Thanks for your submission, @lillo42 .

For review, things I will look at:

- How closely does this code resemble similar feature implemented for SQL Server generator.  Ideally, code should be similar.
- Postgres is a bit different from SQL Server in that it also supports a `USING method` where `method` is one of `[btree, hash, gist, spgist, gin, and brin]` (although, custom methods are allowed, nobody really does this except PhD lab workers). Should we support that additional feature as well?  If not specified, assume Postgres default value `btree`.  Ideally, the `method` would be a C# enum `PostgresIndexMethod`.
- Should we handle the negative test case where someone is using a 9_2 or 10_0 generator with this code?  Right now it will silently fail.

lillo42:
@jzabroski 

>     * How closely does this code resemble similar feature implemented for SQL Server generator.  Ideally, code should be similar.

I implement based on SQL Server 😄 
>     * Postgres is a bit different from SQL Server in that it also supports a `USING method` where `method` is one of `[btree, hash, gist, spgist, gin, and brin]` (although, custom methods are allowed, nobody really does this except PhD lab workers). Should we support that additional feature as well?  If not specified, assume Postgres default value `btree`.  Ideally, the `method` would be a C# enum `PostgresIndexMethod`.

That is something I want to implement, also NULL {First | Last}, CONCURRENTLY and maybe ONLY/WITH/TABLESPACE/WHERE. 😄 
 
>     * Should we handle the negative test case where someone is using a 9_2 or 10_0 generator with this code?  Right now it will silently fail.
What I should to do when user try to use include a column in Postgres  10 or lower? Throw a exception

jzabroski:
Yes, I would throw an exception for older generators. Or at least logsay a warning (in this case, that may be more conservative, as it guarantees non semantic changes that are physical layout issues don't block migration unexpectedly.  While ideally people know their target deployment, some people work in the wild west and need a reminder to get things under control). 

As for SQL Server vs. Postgres, do you see potential for code reuse. I plan to look myself but have not had time yet.

Otherwise this looks good and I'm optimistic to include this soon as part of 3.2.10

jzabroski:
Table space is a huge feature that should work across all processors as most databases except Sqlite support partitioning. There is a ticket in the backlog for this.

jzabroski:
Chances are remaining tasks here are to update FluentMigrator Documentation repo to document the new processor id.

lillo42:
foreach feature (NULL Last/FIRST, using index algorithm and etc) should I open a new PR or do every thing in this one?


> As for SQL Server vs. Postgres, do you see potential for code reuse. I plan to look myself but have not had time yet.

Yes, the code are very similiar.

jzabroski:
> foreach feature (NULL Last/FIRST, using index algorithm and etc) should I open a new PR or do every thing in this one?

Separate PRs are better.  Easier to review and more self-contained.  If, prior to opening the PR, you can create a small issue with related Postgres 11 documentation (or earlier if feature existed before then), that would be great as well.  I get some people just want to go straight to the PR but it helps with google search indexing when we create issues.

jzabroski:
@lillo42 Thanks for the PR and the last addition!

jzabroski:
@lillo42 I should release 3.2.10 by this weekend, and that will likely be the last .NET Core 3.0 release leading up to .NET 5 RTM.  I am trying to rush to get the 4.0 branch merged with 3.2.9+changes so that we can easily/smoothly take in .NET 5.

jzabroski:
@lillo42 If interested I created the associated doc task in the documentation repo for FluentMigrator.  Ideally it would combine SQL Server and Postgres into the same page.  Something like an example with approximate generated SQL would be useful - anything to help land on a Google search for something like "fluentmigrator include column index"

PR #1366, opened 2020-10-29
Add support for using index algorithm

lillo42:
Add support for using index algorithm:

- BTree
- Hash
- Gist
- Spgist
- Gin
- Brin

### DIFF
--- a/src/FluentMigrator.Extensions.Postgres/Model/PostgresIndexAlgorithmDefinition.cs
+++ b/src/FluentMigrator.Extensions.Postgres/Model/PostgresIndexAlgorithmDefinition.cs
@@ -29,7 +29,7 @@ namespace FluentMigrator.Model
 #pragma warning restore 618
     {
         [Required(ErrorMessageResourceType = typeof(ErrorMessages), ErrorMessageResourceName = nameof(ErrorMessages.ColumnNameCannotBeNullOrEmpty))]
-        public virtual PostgresIndexAlgorithm Algorithm { get; set; }
+        public virtual Algorithm Algorithm { get; set; }
 
         [Obsolete("Use the System.ComponentModel.DataAnnotations.Validator instead")]
         public virtual void CollectValidationErrors(ICollection<string> errors)
@@ -43,7 +43,7 @@ namespace FluentMigrator.Model
         }
     }
 
-    public enum PostgresIndexAlgorithm
+    public enum Algorithm
     {
         BTree,
         Hash,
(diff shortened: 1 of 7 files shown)


jzabroski:
This looks great, but I don't see (and was expecting to see) a PostgresExtensions with a signature that let's you pass in the Enum?  Am I just being blind looking at the PR on GitHub and I don't see that interface?

Basically, I see `.UsingBtreeAlgorithm()`, etc, but I dont see `.Using(Algorithm.Btree)`.  The rationale for letting people pass in an enum is it is easier to parameterize.  One use case would be to demo the differences in performances between two index methodologies.  Another use case would be to allow people to use FluentMigrator as a dynamic tooling API to generate scripts from within a GUI.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.UsingIndexAlgorithm.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public static ICreateIndexNonKeyColumnSyntax UsingBrinAlgorithm(this ICreateIndexOnColumnSyntax expression)
+        {
+            var additionalFeatures = expression as ISupportAdditionalFeatures;
+            additionalFeatures.UsingIndexAlgorithm(PostgresIndexAlgorithm.Brin);
+            return new CreateIndexExpressionNonKeyBuilder(expression, additionalFeatures);
+        }
+
+        internal static void UsingIndexAlgorithm(this ISupportAdditionalFeatures additionalFeatures, PostgresIndexAlgorithm algorithm)
```
jzabroski, line 114: Please create a `public` extension method variant of this that:

1. Is named `Using` (so that it auto-completes as `Using(PostgresIndexAlgorithm.Btree)`
2. hangs off `this ICreateIndexOptionsSyntax expression`
3. hangs off `this ICreateIndexOnColumnSyntax expression`

That would be perfect, in terms of what I had in my mind.

jzabroski (changes requested):

src/FluentMigrator.Extensions.Postgres/Model/PostgresIndexAlgorithmDefinition.cs
```diff
@@ -0,0 +1,55 @@
...
+            this.CollectErrors(errors);
+        }
+
+        public object Clone()
+        {
+            return MemberwiseClone();
+        }
+    }
+
+    public enum Algorithm
```
lillo42, line 46: I rename this enum, should I rollback to other name or it's fine calling just as Algorithm?

jzabroski:
This pr is fine. I will merge this weekend

PR #1385, opened 2020-12-02
Add support for concurrently and only index

lillo42:


test/FluentMigrator.Tests/Integration/Migrations/Postgres/IndexMigration.cs
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using FluentMigrator.Postgres;
+
+namespace FluentMigrator.Tests.Integration.Migrations.Postgres
+{
+    [Migration(20201127080600)]
+    public class IndexMigration : Migration
```
lillo42, line 22: How can I run/test this migration?

fubar-coder: You can activate the integration tests for Postgres by configuring a connection string in the `appsettings.json` for the test project and enabling it.

lillo42:
Is it possible run a command without transaction? Because the concurrency index couldn't be create inside of transaction

fubar-coder:
@lillo42 You can disable the transaction for the whole migration (property of the `MigrationAttribute`).

lillo42:
For concurrently feature will be necessary to not use transaction, do you see some problem with it?

jzabroski:
@lillo42 I don't see a problem.  Add a `<remarks></remarks>` section about this restriction so that it shows up in IntelliSense mouse-over

As of v16.6.5, Visual Studio enables by default **"Text Editor > C# > Advanced > Editor Help", option: "Show remarks in Quick Info"**

So, if you add  this note in remarks, people should be able to see it.  We can also add a ticket in the backlog for a static analyzer to warn about this.  But I'd say the percentage of people using Concurrently might be rather small.

fubar-coder:
@jzabroski Maybe another candidate for an analyzer?

jzabroski:
@fubar-coder Can you explain what that analyzer would do?  Is it "If using Postgres Concurrently in a Transaction context, add warning that it won't work, and to set the Transaction context to none"?  If so, I agree.  There are also some SQL Server extensions I wrote for a client with similar behavior.  I have a Migration extension method called KillAllConnectionsExceptMine which calls another Migration extension method, `RequireTransactionBehaviorOfNone(nameof(KillAllConnectionsExceptMine))`

jzabroski:
@fubar-coder @lillo42 Let's move the discussion about transaction behavior to the [Roslyn Analyzers for FluentMigrator](https://github.com/fluentmigrator/fluentmigrator/discussions/1390) discussion thread.  I think this PR is fine for release.

@fubar-coder One last thought: Is there a way for a ISupportAdditionalFeatures to know its migration parent?  I think not, which is likely a good thing.  But I also think one person complained about this awhile ago and I think I mentioned it in the UnitOfWork ADR as part of improving overall modularity.  This would be one scenario where it might be helpful to give access to more info?  On the other hand, the more modular the system is, the harder it might be to refactor and/or write analyzers for.

fubar-coder:
@jzabroski Yes, this is what I meant when I said that it's a candidate for an analyzer. `ISupportAdditionalFeatures` was meant to be lightweight and I fear the additional complexity that might creep into the expression builders and extension methods when they're aware of the context they're called in.

jzabroski:
@fubar-coder I thought about it, and I think that the most we should do is allow a narrow interface to query the context.  Not expose the IMigrationContext itself, but a narrower object that can't even be cast to an IMigrationContext.  But that is nowhere near the top of my list of things to fix or add.

PR #1404, opened 2021-01-14
Postgres - Index - Add support for Index Storage parameters

lillo42:
Add support for Index Storage parameters.

- Some of index parameters depended on Index Method(B-Tree/Hash/GIN/GiST/BRIN) Should I valid it? Or let the postgres decided it?

- Should I throw an exception when I try to use some option that isn't avaliable on Postgres verion that the user is using?


jzabroski:
We can add invalid combinations to the GitHub Discussions thread on Analyzers. https://github.com/fluentmigrator/fluentmigrator/discussions/1390

- Yes, I think we should continue to fail if the option isn't available.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <param name="fillfactor">The fillfactor value from 10 to 100 can be selected</param>
+        /// <returns>The next step</returns>
+        public static ICreateIndexOptionsSyntax WithFillfactor(this ICreateIndexOptionsSyntax expression, int fillfactor)
+        {
+            var additionalFeatures = expression as ISupportAdditionalFeatures;
+            additionalFeatures.SetAdditionalFeature(IndexFillFactor, fillfactor);
+            return expression;
+        }
+
+        #region B-Tree
```
jzabroski, line 44: I wonder if these can be refactored into sub-syntax so that you can say something like `.WithGin().FillFactor().Buffering()`. I am not that familiar with all these knobs in Postgres.  I mainly use the GIN for storing json data and that is it - what do you see as the common way people would write these migrations?

jzabroski (changes requested):
Take a look at my feedback and let me know your thoughts on the stylistic questions.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
```diff
@@ -0,0 +1,171 @@
...
+using FluentMigrator.Infrastructure.Extensions;
+
+namespace FluentMigrator.Postgres
+{
+    public static partial class PostgresExtensions
+    {
+        public const string IndexFillFactor = "PostgresFillFactor";
+
+        /// <summary>
+        /// The fillfactor for an index is a percentage that determines how full the index method will try to pack index pages. For B-trees, leaf pages are filled to this percentage during initial index build,
```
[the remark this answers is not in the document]
lillo42: I understood, this comment I copy from postgres documentation :sweat_smile: 

https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-STORAGE-PARAMETERS

I will move to remarks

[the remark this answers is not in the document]
jzabroski: I think the key part we want to convey in the Summary:

> Postgres B-trees use a default fillfactor of 90, but any integer value from 10 to 100 can be selected.

It's good to know what the default value is in Postgres.  Bonus points if you can find out if the default has changed over time.  SQL Server default is 100% FYI.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
jzabroski: @lillo42 Just making sure you saw this other comment too.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.UsingIndexAlgorithm.cs
```diff
@@ -27,53 +26,99 @@ namespace FluentMigrator.Postgres
...
+        public static ICreateBrinIndexOptionsSyntax UsingBrin(this ICreateIndexOptionsSyntax expression)
         {
             var additionalFeatures = expression as ISupportAdditionalFeatures;
             additionalFeatures.Using(Algorithm.Brin);
-            return expression;
+            return new CreateBrinIndexOptionsSyntax(expression);
         }
 
-        public static ICreateIndexOptionsSyntax Using(this ICreateIndexOptionsSyntax expression, Algorithm algorithm)
+        public static ICreateIndexMethodOptionsSyntax Using(this ICreateIndexOptionsSyntax expression, Algorithm algorithm)
```
jzabroski, line 96: This latest PR looks great.  You are really fast.

I'm going to play around with the syntax changes to make sure the tree works the right way and then merge it if I find no issues.

PR #1407, opened 2021-01-25
Add columns with expression

andrebtoe:
#### What does this implement/fix? Explain your changes.

I added a method to add a column via expression func, to get errors at run time.  I added the ability for columns to be able to have the ColumnAttribute attribute of "System.ComponentModel.DataAnnotations"

```csharp
Create.Table("Project")
           .InSchema("Tenant")
           .WithColumn<ProjectDataModel>(x => x.Id).AsInt32().PrimaryKey().NotNullable()
           .WithColumn<ProjectDataModel>(x => x.Nome).AsString(10).NotNullable();
```

### DIFF
--- a/src/FluentMigrator.Abstractions/Infrastructure/Extensions/LambdaExpressionExtension.cs
+++ b/src/FluentMigrator.Abstractions/Infrastructure/Extensions/LambdaExpressionExtension.cs
@@ -1,0 +1,49 @@
+using System;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace FluentMigrator.Infrastructure.Extensions
+{
+    public static class LambdaExpressionExtension
+    {
+        public static string GetMemberName(this LambdaExpression memberSelector)
+        {
+            var currentExpression = memberSelector.Body;
+
+            while (true)
+            {
+                switch (currentExpression.NodeType)
+                {
+                    case ExpressionType.Parameter:
+                        return ((ParameterExpression)currentExpression).Name;
+                    case ExpressionType.MemberAccess:
+                        return GetMeberNameWithColumnAttribute(((MemberExpression)currentExpression).Member);
+                    case ExpressionType.Call:
+                        return ((MethodCallExpression)currentExpression).Method.Name;
+                    case ExpressionType.Convert:
+                    case ExpressionType.ConvertChecked:
+                        currentExpression = ((UnaryExpression)currentExpression).Operand;
+                        break;
+                    case ExpressionType.Invoke:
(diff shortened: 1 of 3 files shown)


src/FluentMigrator.Abstractions/Infrastructure/Extensions/LambdaExpressionExtension.cs
```diff
@@ -0,0 +1,49 @@
...
+            var currentExpression = memberSelector.Body;
+
+            while (true)
+            {
+                switch (currentExpression.NodeType)
+                {
+                    case ExpressionType.Parameter:
+                        return ((ParameterExpression)currentExpression).Name;
+                    case ExpressionType.MemberAccess:
+                        return GetMeberNameWithColumnAttribute(((MemberExpression)currentExpression).Member);
```
jzabroski, line 21: Typo

src/FluentMigrator.Abstractions/Infrastructure/Extensions/LambdaExpressionExtension.cs
```diff
@@ -0,0 +1,49 @@
...
+                        break;
+                    case ExpressionType.ArrayLength:
+                        return "Length";
+                    default:
+                        throw new Exception("not a proper member selector");
+                }
+            }
+        }
+
+        public static string GetMeberNameWithColumnAttribute(MemberInfo memberInfo)
```
jzabroski, line 39: Typo

jzabroski (changes requested):
- [ ] There are some typos.
- [ ] Should this be in its own package?
- [ ] I think the summary isn't very clear to me.  It doesn't surface that this can use ColumnAttribute, etc.   Consider that this summary will also appear in the online documentation on fluentmigrator.github.io under the API explorer
- [ ] It seems like the best approach would be to have a `WithColumnsFromClass<T>` method that calls this in a loop.  Why would you want to type this over and over again if you already have the info on your class via ?
- [ ] I think you should update the FluentMigrator samples code to include this as fourth migration.
- [ ] Update fluentmigrator/documentation to include this.

I like your creativity, and I think @mikependon (maintainer of RepoDB) would love this.

That said, I wonder if this should be in a different namespace and/or package?  There used to be a similar project, EasyMigrator, but the owner took everything down off bitbucket.  I tried to contact him but he never replied.

jzabroski:
Hi @andrebtoe - Just checking if you saw my review notes. Thank you.

jzabroski:
As an update, Quentin Starin and Michael Camara Pendon and I are having a three-way chat about resurrecting EasyMigrator.FluentMigrator package, which does this and more.

jzabroski:
Quentin did resurrect EasyMigrator. However, it only supports .NET Framework https://github.com/qstarin/EasyMigrator

PR #1411, opened 2021-02-01
Oracle: Log SQL on error

RobertWilliams2000:
Currently SQL Server and PostgreSQL runners log SQL when an error occurs.
This change will make Oracle follow the same pattern.

### DIFF
--- a/src/FluentMigrator.Runner.Oracle/Processors/DotConnectOracle/DotConnectOracleProcessor.cs
+++ b/src/FluentMigrator.Runner.Oracle/Processors/DotConnectOracle/DotConnectOracleProcessor.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.IO;
 
 using FluentMigrator.Expressions;
 using FluentMigrator.Runner.Generators.Oracle;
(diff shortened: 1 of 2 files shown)


src/FluentMigrator.Runner.Oracle/Processors/DotConnectOracle/DotConnectOracleProcessor.cs
```diff
@@ -208,7 +209,24 @@ protected override void Process(string sql)
...
+            {
+                try
+                {
+                    command.ExecuteNonQuery();
+                }
+                catch (Exception ex)
+                {
+                    using (var message = new StringWriter())
+                    {
+                        message.WriteLine("An error occured executing the following sql:");
```
jzabroski, line 221: This looks great.  Can you consider consolidating the code across the three Processors so that there is a standard way to catch these exceptions and log the error?  Even if SQL Server and Postgres are a bit different, I think a SqlExceptionUtility class would be useful to avoid repeating logic everywhere.

jzabroski (changes requested):
Please see my review comments.  Thanks for contributing!

src/FluentMigrator.Runner.Oracle/Processors/DotConnectOracle/DotConnectOracleProcessor.cs
jzabroski: This is what I have in mind:

```c#
public static class SqlExceptionUtility
{
  public static void Throw(string sql, Exception ex)
  {
    using (var message = new StringWriter())
    {
       message.WriteLine("An error occured executing the following sql:");
       message.WriteLine(sql);
       message.WriteLine("The error was {0}", ex.Message);
       throw new Exception(message.ToString(), ex);
    }
  }
}
```

jzabroski: It should go in the Runner.csproj

RobertWilliams2000: @jzabroski , I added it as a virtual method to ProcessorBase.cs.
This seemed closer to the classes that would utilize it.

jzabroski:
Cool.  Will try to release this tonight.

jzabroski (approved):

Issue #967, opened 2018-11-26
Create table with tablespace?


CQJonnyLin:
Can you tell me how to create table with tablespace.I thy create table with inschema,but schema is user.
`Create.Table("TEST_XML").InSchema("ZL9BASEITEM")
                .WithColumn("input").AsString(2000).Nullable()
                .WithColumn("data").AsXml().Nullable();`
I want to create table name is TEST_XML with tablespace name is ZL9BASEITEM.

fubar-coder:
It's not supported yet. I'll leave this open as a reminder for a new feature for Oracle. You may use the following workaround in the meantime:

```cs
Execute.Sql("ALTER DATABASE DEFAULT TABLESPACE ZL9BASEITEM");
Create.Table("TEST_XML")
    .WithColumn("input").AsString(2000).Nullable()
    .WithColumn("data").AsXml().Nullable();
```

It's correct, that Oracle interprets the schema part of a fully qualified table name as user name.

I guess that it would be nice to have an oracle extension that would allow writing something like:

```cs
Create.Table("TEST_XML")
    .InTableSpace("ZL9BASEITEM")
    .WithColumn("input").AsString(2000).Nullable()
    .WithColumn("data").AsXml().Nullable();
```

This would be an Oracle-only extension.

CQJonnyLin:
> It's not supported yet. I'll leave this open as a reminder for a new feature for Oracle. You may use the following workaround in the meantime:
> 
> ```cs
> Execute.Sql("ALTER DATABASE DEFAULT TABLESPACE ZL9BASEITEM");
> Create.Table("TEST_XML")
>     .WithColumn("input").AsString(2000).Nullable()
>     .WithColumn("data").AsXml().Nullable();
> ```
> 
> I found this on this web-site: https://www.techonthenet.com/oracle/tablespaces/set_default.php
> 
> It's correct, that Oracle interprets the schema part of a fully qualified table name as user name.
> 
> I guess that it would be nice to have an oracle extension that would allow writing something like:
> 
> ```cs
> Create.Table("TEST_XML")
>     .InTableSpace("ZL9BASEITEM")
>     .WithColumn("input").AsString(2000).Nullable()
>     .WithColumn("data").AsXml().Nullable();
> ```
> 
> This would be an Oracle-only extension.

Your guess perfect.My now solution is excute sql

jzabroski:
Discussed in FluentMigrator gitter channel today with @hayer

Notes:

[...]

PR #1414, opened 2021-02-08
Postgres add support for tablespace

lillo42:
Add support for tablespace in index

### DIFF
--- a/src/FluentMigrator.Extensions.Postgres/Builder/Create/Index/IndexMethod/CreateBTreeIndexOptionsSyntax.cs
+++ b/src/FluentMigrator.Extensions.Postgres/Builder/Create/Index/IndexMethod/CreateBTreeIndexOptionsSyntax.cs
@@ -34,7 +34,7 @@ namespace FluentMigrator.Builder.Create.Index
         /// <inheritdoc />
         public ICreateBTreeIndexOptionsSyntax Fillfactor(int fillfactor)
         {
-            Fillfactor(fillfactor);
+            base.Fillfactor(fillfactor);
             return this;
         }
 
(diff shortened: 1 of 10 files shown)


jzabroski (changes requested):
@lillo42 Would you have an appetite to tackle this for all DBMSes that support tablespace and/or filegroup(s)?  There are a lot of them outlined in #967

lillo42:
@jzabroski Yes, I would like to tackle that in the future, now I'm focus on PostgreSQL missing feature(like ~~index~~/create extensions/improve table feature)

jzabroski:
OK, So is the plan that at some point, we would do a major release where we move the namespace for Tablespace from Postgres into the global namespace?  Unless I missed it, your current proposal doesnt have  a Create.Tablespace root item.  So, would people just call Execute.Sql to create the tablespace?

lillo42:
> OK, So is the plan that at some point, we would do a major release where we move the namespace for Tablespace from Postgres into the global namespace? 

I don't know if is necessary to be major release to do it, but we can move it to from Postgres into the global namespace or create direct in the global namespace.

> Unless I missed it, your current proposal doesnt have a Create.Tablespace root item. So, would people just call Execute.Sql to create the tablespace? If so, I suggest we update the summary docs to explain that is the approach for now

Thinking better now, it's feature isn't so necessary, maybe more for DBA, so I believe this feature can we for tablespace implementation.

jzabroski:
I will try to merge this tomorrow.  Got super busy with deadlines at work.

jzabroski:
@lillo42 Sorry I've been so tied up with work.  Will get to this by this weekend. This is my last week before vacation

PR #1433, opened 2021-02-24
Improve docs on PostgreSQL Extensions

lillo42:


src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
```diff
@@ -44,38 +43,5 @@ public static ICreateBTreeIndexOptionsSyntax Fillfactor(this ICreateIndexOptions
             return expression.UsingBTree()
                 .Fillfactor(fillfactor) as ICreateBTreeIndexOptionsSyntax;
         }
-
```
jzabroski, line 47: @lillo42 Why is this PR about improving the docs but you're deleting code?  Everything else looks good to me.  Just making sure you did this on purpose.

jzabroski (changes requested):

fubar-coder (changes requested):

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
jzabroski: @lillo42  Any answer here?  Thanks.

src/FluentMigrator.Extensions.Postgres/Builder/Create/Index/IndexMethod/GistBuffering.cs
```diff
@@ -16,10 +16,24 @@
...
+        /// </summary>
+        Auto,
+
+        /// <summary>
+        /// Enabled buffering
+        /// </summary>
         On,
-        Off,
-        Auto
+
```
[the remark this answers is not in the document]
jzabroski: @fubar-coder I think in this case we should accept it.  On and Auto are very similar conceptually.  And I believe Postgres's default is Auto.  I should've caught that in the initial review but was not thinking about that as something with a default value.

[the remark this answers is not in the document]
fubar-coder: TBH I don't think that it'll be a problem in practice, just a potential one.

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
fubar-coder: I'm curious about this change too. However, I do like that the constants are moved to the `PostgresExtensions` file even though that it'll be a breaking change too (at source level).

fubar-coder (changes requested):

src/FluentMigrator.Extensions.Postgres/Postgres/PostgresExtensions.IndexStorage.cs
lillo42: Sorry for delay, I was moving last week, a lot of thing to do, about the delete code I realize that I have forgot to delete this code after move the storage parameter to the interfaces, maybe it's better rollback it and add Obsolete attribute, what do you think?

fubar-coder: If there is a replacement for those functions and it didn't exist for too long, then I agree that the breaking change might not be too disruptive for the FM users out there.

fubar-coder (approved):
I think that the risk of this breaking change is low, so I guess that it'll be fine to merge.

Issue #1441, opened 2021-03-22
Migration.Insert.Row where a value is Enum


MihailsKuzmins:
I haven't found a similar issue, but please redirect me if it actually exists. (I apologise if I've created a duplicate by accident)
I have a problem with this code
```cs
public enum myEnum
{
    Value1 = 12334
} 

var values = new Dictionary<string, object>
{
    { "intColumn", myEnum.Value1 }
};

Insert.IntoTable("myTable")
    .Row(values);
```

The error I get: **!!! Incorrect integer value: 'Value' for column 'intColumn' at row 1**
So it means that the runner transforms enums into their string values instead of their underlying numeric type which is not the right behaviour.
Would it be possible to take a look at this case? I can possibly try to do it as well

jzabroski:
```c#
public enum myEnum
{
    Value1 = 12334
} 

var values = new Dictionary<string, object>
{
    { "intColumn", (int)myEnum.Value1 }
};

Insert.IntoTable("myTable")
    .Row(values);
```

MihailsKuzmins:
@jzabroski, thank for the reply.
Of course it does, but it makes my code a bit clumsy and unnatural.
Do you think that there might be problems with enum handling?

jzabroski:
@MihailsKuzmins There is no enum handling .  The code does not do any coercion.

If you find it unnatural, I would suggest creating a wrapper syntax, where you traverse the dictionary and update object values from enum values to integral storage values.  However, you might want to also handle the underlying storage of the enum type, when you perform your cast.

```c#
public enum myEnum : short
{
    Value1 = 12
} 
```

I would think of this transform as similar to using Linq and calling Cast on an Array.

jzabroski:
If you want to submit a PR for handling enums, you can, but you need to handle all possible integral storage types in your unit tests.  Thank you,

MihailsKuzmins:
@jzabroski 
I'm thinking about creating a PR for enum handling. There is one question I would like to ask about a possible implementation.
Of course the easiest way is to do it in `InsertDataExpressionBuilder`, but it will result for an unnecessary mapping from Dictionary to Dictionary for `Row(IDictionary<string, object> data)`. Therefore, in order to avoid it, enum handling could be implemented where rows are enumerated over, but it requires changes in more than one classes (of course handling will be an extracted extension method).

So, due to the fact that I don't see a good solution, could you please advise, what could be a better approach? One of the aforecited or a completely different one?

[...]

PR #1475, opened 2021-05-09
handle enum values in Quoters

MihailsKuzmins:
PR for the issue #1441.

- created `QuoterOptions` which is passed to all Quoters
- handle enum values by `EnumAsUnderlyingType` (false - stringm true - numeric). Default value: false, i.e. string (as it is now)
- added a configuration method `WithQuoterConventions()` in `MigrationRunnerBuilderExtensions`
- added tests for enum handling


test/FluentMigrator.Tests/Unit/Generators/GenericGenerator/QuoterTests.cs
```diff
@@ -74,74 +74,74 @@ public override string ToString()
         [Test]
```
jzabroski, line 75: I think this would be a lot cleaner with an NUnit Theory, so that you can parameterize the options values and outcomes.

In other words, while it might be a bit silly to set EnumAsString in the below CanEscapeAString test, it would then be possible to test not just a string, but an enum-as-string.

Thoughts?

test/FluentMigrator.Tests/Unit/Generators/GenericGenerator/QuoterTests.cs
```diff
@@ -187,6 +186,18 @@ public void EnumIsFormattedAsString()
                 .ShouldBe("'Bar'");
         }
 
+        [Test]
+        public void EnumIsFormatterAsUnterlyingType()
```
jzabroski, line 192: Typo. Please find+replace `EnumIsFormatterAsUnterlyingType` with `EnumIsFormatterAsUnderlyingType`

src/FluentMigrator.Runner.Core/Generators/Generic/GenericQuoter.cs
```diff
@@ -26,11 +26,11 @@ namespace FluentMigrator.Runner.Generators.Generic
 
     public class GenericQuoter : IQuoter
     {
-        private readonly QuoterOptions _quoterOptions;
+        private readonly QuoterOptions _options;
 
-        public GenericQuoter(QuoterOptions quoterOptions = null)
+        public GenericQuoter([CanBeNull] QuoterOptions options)
```
jzabroski: Please see @fubar-coder remarks about having the default constructor still available.  I would not use [CanBeNull] - instead, the default constructor should just initialize the quoter to a default value.  Adding nulls will just complicate future code.

src/FluentMigrator.Runner.Firebird/Generators/Firebird/FirebirdGenerator.cs
```diff
@@ -39,23 +40,26 @@ public class FirebirdGenerator : GenericGenerator
...
         {
         }
 
         public FirebirdGenerator(
             [NotNull] FirebirdOptions fbOptions,
+            [NotNull] IOptions<QuoterOptions> quoterOptions,
             [NotNull] IOptions<GeneratorOptions> generatorOptions)
-            : this(new FirebirdQuoter(fbOptions.ForceQuote), fbOptions, generatorOptions)
+            : this(new FirebirdQuoter(fbOptions, quoterOptions), fbOptions, quoterOptions, generatorOptions)
         {
```
jzabroski: It seems like you moved the ForceQuote flag off of FirebirdOptions and onto QuoterOptions - but I don't see where?  Was it in a previous commit?

src/FluentMigrator.Runner.Firebird/Generators/Firebird/FirebirdGenerator.cs
```diff
@@ -39,23 +40,26 @@ public class FirebirdGenerator : GenericGenerator
...
         }
 
         public FirebirdGenerator(
```
jzabroski, line 55: We need an obsolete overload for this constructor unfortunately.  (Unless I am missing something.)

test/FluentMigrator.Tests/Unit/Generators/SQLite/SQLiteDataTests.cs
```diff
@@ -28,21 +29,16 @@ namespace FluentMigrator.Tests.Unit.Generators.SQLite
...
     public class SQLiteDataTests : BaseDataTests
     {
-        protected SQLiteGenerator Generator;
-
-        [SetUp]
-        public void Setup()
-        {
-            Generator = new SQLiteGenerator();
-        }
+        private static SQLiteGenerator CreateFixture(QuoterOptions options = null) =>
```
jzabroski, line 34: I wonder if BaseDataTests should be generic and have a `BaseDataTests<T> where T : IGenerator` so that CreateFixture is protected static:

```c#
protected static T CreateFixture(QuoterOptions options = null) =>
```

@fubar-coder It sounded like you didn't like the CreateFixture() thing, vs. using NUnit's well-defined SetUp - but I kind of like it because the style is more in-line with XUnit which I am used to.  I don't really love a distinct SetUp phase.  The one use case that SetUp provides is when working with cancellable Tasks, and needing to memoize the cancellationtoken for the duration of the test.  Just my opinion.

src/FluentMigrator.Runner.Firebird/Generators/Firebird/FirebirdQuoter.cs
```diff
@@ -75,7 +75,9 @@ public class FirebirdQuoter : GenericQuoter
         {
         }
 
-        public FirebirdQuoter(bool forceQuote, IOptions<QuoterOptions> options)
```
jzabroski: N.B.D. (No Big Deal).  Reviewed

src/FluentMigrator.Runner.Firebird/Generators/Firebird/FirebirdGenerator.cs
```diff
@@ -41,7 +41,7 @@ public class FirebirdGenerator : GenericGenerator
...
             [NotNull] FirebirdOptions fbOptions)
-            : this(fbOptions,new OptionsWrapper<GeneratorOptions>(new GeneratorOptions()))
```
jzabroski: N.B.D. (No Big Deal).  Reviewed

src/FluentMigrator.Runner.Redshift/Generators/Redshift/RedshiftDescriptionGenerator.cs
```diff
@@ -26,7 +26,7 @@ namespace FluentMigrator.Runner.Generators.Redshift
     /// </summary>
     public class RedshiftDescriptionGenerator : GenericDescriptionGenerator
     {
-        private readonly RedshiftQuoter _quoter;
+        private readonly IQuoter _quoter;
```
jzabroski, line 29: N.B.D.  (No Big Deal), Reviewed.

src/FluentMigrator.Runner.SqlServer/Processors/SqlServer/SqlServerProcessorFactory.cs
```diff
@@ -18,10 +18,7 @@ public SqlServerProcessorFactory()
         }
 
         [Obsolete]
-        public override IMigrationProcessor Create(
-            string connectionString,
-            IAnnouncer announcer,
-            IMigrationProcessorOptions options)
+        public override IMigrationProcessor Create(string connectionString, IAnnouncer announcer, IMigrationProcessorOptions options)
```
jzabroski: N.B.D. (No Big Deal), Reviewed

src/FluentMigrator.Runner.Core/Generators/Generic/GenericQuoter.cs
```diff
@@ -30,9 +30,9 @@ public class GenericQuoter : IQuoter
     {
         private readonly QuoterOptions _options;
 
-        public GenericQuoter([CanBeNull] IOptions<QuoterOptions> options = null)
+        public GenericQuoter(IOptions<QuoterOptions> options)
```
jzabroski, line 33: I see you did my comments in the previous commit.  Agree, using [CanBeNull] and assigning to null adds messy three-valued logic that we don't need.  **We just need a default constructor for GenericQuoter so that we don't break ABI with existing consumers.**

jzabroski (changes requested):
@MihailsKuzminsDG Thanks for this contribution.  Most of my remarks are fairly minor - typos, adjusting application binary interface to ensure no breaking changes.  It looks like you took care of one of my earlier remarks about the QuoterOptions not being three-valued logic but we still need a default constructor.  Also, the FirebirdGenerator constructor appears to have an ABI breaking change - can you please restore the old constructor and mark it as Obsolete so that consumers are aware.

src/FluentMigrator.Runner.Core/Initialization/QuoterOptions.cs
```diff
@@ -1,14 +1,11 @@
...
 {
-    /// <summary>
-    /// Conventions of how C# values are converted to SQL
-    /// </summary>
-    public sealed class QuoterOptions
+    /// <inheritdoc cref="IQuoterOptions"/>
+    public sealed class QuoterOptions : IQuoterOptions
     {
+        /// <inheritdoc />
         /// <summary>
```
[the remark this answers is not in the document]
MihailsKuzmins: sorry, my bad. I created an interface for it where `<remarks>` were specified, so the `<summary>` was taken from the actual class and the `<remarks>` were taken from the interface.
Along the way I deleted the interface and probably forgot to remove the `<inferitdoc />` tag 😔

test/FluentMigrator.Tests/Unit/Generators/GenericGenerator/QuoterTests.cs
MihailsKuzmins: Unlike xUnit, NUnit has global Theory parameters (it might become painful to have multiple Theories in one class). Do you think tests for `QuoteColumnName` should be extracted to another class?
I am sorry, but I don't quite follow you with regards to `CanEscapeAString`. The property `EnumAsString` (now EnumAsUnderlyingType) does not affect `QuoteColumnName` which is tested by `CanEscapeAString`

jzabroski: That's the point- the theory would verify it doesn't affect QuoteColumnName.  It might be needless complexity since an enum field name can't have a common escape in it - but Oracle and DB2 allow for arbitrary escapes, so it gets interesting fast.

src/FluentMigrator.Runner.Db2/Processors/Db2/Db2Processor.cs
```diff
@@ -36,10 +36,10 @@ namespace FluentMigrator.Runner.Processors.DB2
     public class Db2Processor : GenericProcessorBase
     {
         [Obsolete]
-        public Db2Processor(IDbConnection connection, IMigrationGenerator generator, IAnnouncer announcer, IMigrationProcessorOptions options, IDbFactory factory)
+        public Db2Processor(IDbConnection connection, IMigrationGenerator generator, IAnnouncer announcer, IMigrationProcessorOptions options, IDbFactory factory, IOptions<QuoterOptions> quoterOptions)
```
MihailsKuzmins, line 39: remove this parameter

src/FluentMigrator.Runner.Redshift/Generators/Redshift/RedshiftDescriptionGenerator.cs
```diff
@@ -28,9 +28,9 @@ public class RedshiftDescriptionGenerator : GenericDescriptionGenerator
     {
         private readonly IQuoter _quoter;
 
-        public RedshiftDescriptionGenerator()
+        public RedshiftDescriptionGenerator(RedshiftQuoter quoter)
```
MihailsKuzmins, line 39: add an [obsolete] overload

src/FluentMigrator.Runner.Core/Generators/Generic/GenericQuoter.cs
MihailsKuzmins: `[CanBeNull]` for just for me in order that I don't have billions of compilation errors when I add this new parameter. But you're right, I should've added a default obsolete constructor.

src/FluentMigrator.Runner.Firebird/Generators/Firebird/FirebirdGenerator.cs
MihailsKuzmins: No, I haven't moved `_forceQuote`, is it in the `FirebirdQuoter` were it had been before I pushed my changes.
The reason for this change is that I simply wanted to use another constructor where `FirebirdOptions` and `IOptions<QuoterOptions>` are provided.
I use this constructor. I does exactly the same - pass `ForceQuote` from `ForceQuote` to `_forceQuote`.
In theory, it can still be rolled back to `new FirebirdGenerator(fbOptions.ForceQuote, new OptionsWrapper ...)`
image

test/FluentMigrator.Tests/Unit/Generators/SQLite/SQLiteDataTests.cs
MihailsKuzmins: I prefer xUnit as well. I don't think that having `CreateFixture` in `BaseDataTests` will result in a maintainable method because CreateFixture might accept other parameters - not only `QuoterOptions`.
Usually, for each class a create an abstract test class where I define the `CreateFixture` method. It this case it would mean adding another abstract class between SQLiteDataTests and BaseDataTests which is an overkill (isn't it?).
Well, therefore, I think having `CreateFixture` is the same class is OK. But please let me know the final decision. I will do it as it suits the FluentMigrator :)

schambers:
This PR is outdated by several years and is stale @jzabroski. Any concerns about closing it? I can ping the original author to clean it up and open a new PR if needed. Thoughts?

jzabroski:
We can close it. I might reimplement it now that I'm done making type map injectable

PR #1478, opened 2021-05-18
Fix not working password mask when connection string contains whitespace

wiciok:
When connection string contained whitespace around equality character (ex. 'Password = examplepassword'), password masking didn't work properly, thus plain text password was appearing in the logs.


wiciok:
This PR fixes #1479

fubar-coder (approved):
LGTM :shipit:

PR #1543, opened 2021-11-27
Adding additional description properties

azerios:
PR to add any additional description properties

This will allow any user to setup any additional description properties in order to store additional informations or usage of any columns inside the database

### DIFF
--- a/src/FluentMigrator.Abstractions/Model/ColumnDefinition.cs
+++ b/src/FluentMigrator.Abstractions/Model/ColumnDefinition.cs
@@ -110,9 +110,9 @@ namespace FluentMigrator.Model
         public virtual ColumnModificationType ModificationType { get; set; }
 
         /// <summary>
-        /// Gets or sets the column description
+        /// Gets or sets the column descriptions
         /// </summary>
-        public virtual string ColumnDescription { get; set; }
+        public virtual Dictionary<string, string> ColumnDescriptions { get; set; } = new Dictionary<string, string>();
 
         /// <summary>
         /// Gets or sets the collation name if the column has a string or ansi string type
--- a/src/FluentMigrator.Abstractions/Builders/IColumnOptionSyntax.cs
+++ b/src/FluentMigrator.Abstractions/Builders/IColumnOptionSyntax.cs
@@ -51,6 +51,14 @@ namespace FluentMigrator.Builders
         TNext WithColumnDescription(string description);
 
         /// <summary>
+        /// Sets any columns description
+        /// </summary>
+        /// <param name="descriptionName">The descriptionName</param>
+        /// <param name="description">The description</param>
+        /// <returns>The next step</returns>
+        TNext WithColumnAdditionalDescription(string descriptionName, string description);
+
+        /// <summary>
         /// Sets the columns identity configuration
         /// </summary>
         /// <returns>The next step</returns>
--- a/src/FluentMigrator.Runner.Core/Generators/Generic/GenericDescriptionGenerator.cs
+++ b/src/FluentMigrator.Runner.Core/Generators/Generic/GenericDescriptionGenerator.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 //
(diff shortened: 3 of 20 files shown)


src/FluentMigrator.Abstractions/Model/ColumnDefinition.cs
```diff
@@ -110,9 +110,9 @@ public class ColumnDefinition
         public virtual ColumnModificationType ModificationType { get; set; }
 
         /// <summary>
-        /// Gets or sets the column description
+        /// Gets or sets the column descriptions
         /// </summary>
-        public virtual string ColumnDescription { get; set; }
+        public virtual Dictionary<string, string> ColumnDescriptions { get; set; } = new Dictionary<string, string>();
```
jzabroski, line 115: @azerios This part would still be a breaking change, so we would need to put it in a 4.0 branch.

jzabroski (changes requested):

src/FluentMigrator.Abstractions/Model/ColumnDefinition.cs
azerios: @jzabroski I need this feature to store GPDR details of each field of my database, like who,when,why asked for this field and what its his purpose; In which repo or branch do I need to merge ?

fubar-coder: A IMHO better solution would've been to add a new property called "AdditionalProperties" or something like that and keep the "ColumnDescription" property untouched.

jzabroski: @fubar-coder  I was thinking the same thing but was struggling with what to describe the new property name as.  "AdditionalProperties" makes sense.

@azerios Can I ask why you wouldn't just track that metadata using a database table schema?

e.g.

```
Create.Table("Schema").InSchema("Metadata")
.AddColumn("SchemaId").AsInt64().NotNullable().Identity(1,1).PrimaryKey()
.AddColumn("Name").AsString(255).NotNullable();

Create.Table("Table").InSchema("Metadata")
.AddColumn("TableId").AsInt64().NotNullable().Identity(1,1).PrimaryKey()
.AddColumn("SchemaId").AsInt64().NotNullable().Identity(1,1).ForeignKey("Metadata", "Schema", "SchemaId")
.AddColumn("Name").AsString(255).NotNullable();

Create.Table("Column").InSchema("Metadata")
.AddColumn("ColumnId").AsInt64().NotNullable().Identity(1,1).PrimaryKey()
.AddColumn("TableId").AsInt64().NotNullable().ForeignKey("Metadata", "Table", "TableId")
.AddColumn("Name").AsString(255).NotNullable();

Create.Table("ColumnProperty").InSchema("Metadata")
.AddColumn("ColumnPropertyId").AsInt64().NotNullable().Identity(1,1).PrimaryKey()
.AddColumn("ColumnId").AsInt64().NotNullable().ForeignKey("Metadata", "Column", "ColumnId")
.AddColumn("Description").AsString(int.MaxValue).NotNullable();
```

You can then write maintenance migrations that verify, for every row in your INFORMATION_SCHEMA.COLUMNS, you are tracking what you want to track.  You can also use tools like STRING_CONCAT to then easily write SQL reports that not only track the GDPR properties but what their contents are (assuming it is not encrypted / data at rest storage requirement).

This is how I would solve your problem.  It is also portable across every database vendor.

Let me know your thoughts.

jzabroski: @azerios @fubar-coder  actually, I think the recommended way to solve this with SQL Server is via SENSITIVITY CLASSIFICATION feature.  See this other user's question: https://github.com/fluentmigrator/fluentmigrator/discussions/1537

jzabroski:
@azerios I believe your other pull request (that was merged into 4.0) solves the issues with this PR.  Please correct me if I am wrong, but I am closing this PR for now.

Issue #1492, opened 2021-07-07
Sqlite Creating Unique Constraint not working


fairking:
**Describe the bug**
As we just found out the following line:
```
public override void Up()
        {
            Create.UniqueConstraint("uq_my_table").OnTable("my_table").Columns(new[] { "col1", "col2" });
        }
```
is not working with sqlite database. After the migration executed the constraint is missing in sqlite db and the duplicates are allowed.
However when I added extra line as a workaround, it created the expected constraint:
```
public override void Up()
        {
            Create.UniqueConstraint("uq_my_table").OnTable("my_table").Columns(new[] { "col1", "col2" });
            IfDatabase("sqlite").Execute.Sql("CREATE UNIQUE INDEX uq_my_table ON my_table(col1, col2);");
        }
```

**To Reproduce**
See description

**Expected behavior**
It must create an unique constraint

**Information (please complete the following information):**
 - OS: Windows 10
 - Platform .NET Core 3.1
 - FluentMigrator version 3.2.17
 - FluentMigrator runner
 - Database Management System SQLite
 - Database Management System Version

**Additional context**
None

jzabroski:
Hi @fairking ,

Hmm... That's pretty weird.  I went to the changes log for SQLite and it looks like this feature has been around awhile:

> ### 2001-09-27 (2.0-alpha-3)
> SQLite now honors the UNIQUE keyword in CREATE UNIQUE INDEX. Primary keys are required to be unique.

I think the problem is here:
https://github.com/fluentmigrator/fluentmigrator/blob/9c87d698b161b3a1bb596a5624c02891a64459c5/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs#L108-L111

And the fact that SQLite is not set to Strict but Loose compatibility mode by default cf https://github.com/fluentmigrator/fluentmigrator/issues/812#issuecomment-383417163

I think the SQLiteGenerator code probably needs to be thrown out and replaced.

Fungusware:
Yes, this seem to be happening to me also.

Unique Constraint is not created and also Foreign Key create as

`            Alter.Table("Usage").AddColumn("LicenseId").AsInt32().Nullable().ForeignKey("License", "Id");`

Is not being applied.

zelenij:
I experience the same issue with unique constraints in Sqlite

mattbrailsford:
[...]

Issue #1490, opened 2021-06-16
Microsoft.Data.SqlClient epic


jzabroski:
It seems like the 3.2.16 release decision to move to Microsoft.Data.SqlClient may have been a regretful one.  Initially it was only Mono users impacted by this problem, but now it appears to be more general.

#1483  -- @arhughes
#1486  -- @m4rkyx 

> It works but there seems to be some kind of locking of the dll for Microsoft.Data.SqlClient.SNI.
> Example. I have an ASP.NET MVC 5 site which also runs migrations using FluentMigrator, hosting it in IIS and rebuilding the solution fails because the dll is locked by the IIS process.
> `using (var scope = ServiceProvider.CreateScope())` doesn't seem to help with releasing resources.
> 
> -- @YanerTavuz  https://github.com/fluentmigrator/fluentmigrator/issues/1023#issuecomment-862367436

PeterBenko:
Do the regrets have something to do with [this issue](https://github.com/dotnet/SqlClient/issues/727) or are they unrelated?

jzabroski:
> Do the regrets have something to do with [this issue](https://github.com/dotnet/SqlClient/issues/727) or are they unrelated?

Interesting. @mattbrailsford I wonder if the SNI issue with SQLite is related to this issue?

@PeterBenko Matt had to manually include the SQLite native interface package in his recent PR.

jzabroski:
@PeterBenko  I think I found the root cause, which wasn't really articulated in that issue you linked, but hinted at:

It looks like this document explains it: https://docs.microsoft.com/en-us/sql/connect/ado-net/sqlclient-troubleshooting-guide?view=sql-server-ver15

> # Unable to load native SNI (Server Name Indication) library
> 
> SNI is the native C++ library that SqlClient depends on for various network operations when running on Windows. **In .NET Framework applications that are built with the MSBuild Project SDK, native DLLs aren't managed with restore commands. So a ".targets" file is included in the "Microsoft.Data.SqlClient.SNI" NuGet package that defines the necessary "Copy" operations.**

Our tests run on net461:

https://github.com/fluentmigrator/fluentmigrator/blob/master/test/FluentMigrator.Tests/FluentMigrator.Tests.csproj#L3

 I believe @mattbrailsford ran into a similar issue, and his fix was to force a PackageReference to the package containing the missing dll.  

[...]

Issue #1456, opened 2021-04-19
Sqlite Identity column is required to be int32


zelenij:
**Describe the bug**

Sqlite code generator seems to require that Identity column must be Int32 - in which case it will be defined as autoincrement. I can't find any reason for this in the documentation.  In fact, the sqlite doc says that primary autoincrement key will be mapped to the hidden ROWID column, which is 64 bit.  So there is no reason that the code generator shouldn't accept Int64 at least.  The relevant code line:

https://github.com/fluentmigrator/fluentmigrator/blob/9c87d698b161b3a1bb596a5624c02891a64459c5/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs#L34

Sqlite doc:

https://sqlite.org/autoinc.html

**To Reproduce**
In a migration define a column like this:

```
        this.Create.Table("TestTable")
            .WithColumn("Id").AsInt64().PrimaryKey().Identity()
```

And run a migration for Sqlite

**Expected behavior**

New table successfully generated.

**Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Platform .net5
 - FluentMigrator version latest
 - FluentMigrator runner [  FluentMigrator.Console ]
 - Database Management System [ SQLite ]
 - Database Management System Version [ N/A ]

fubar-coder:
This is somewhat confusing, and might indeed be a problem when you want to use the same migration for multiple databases and you expect more than 2 billion rows in a table.

If you're using the migration only with SQLite, then this is not a problem at all, because SQLite doesn't distinguish between multiple integer types of different widths. In other words: SQLite only knows "INTEGER", "REAL", etc..., but other types (e.g. "INT8", "DOUBLE") are mapped to the 5 basic types SQLite knows about: [INTEGER, TEXT, BLOB, REAL, NUMERIC](https://www.sqlite.org/datatype3.html)

zelenij:
This is exactly my scenario - I expect a lot of record in production, where I'm targeting PostgreSQL, but for local testing I'm fine with Sqlite for now.  Ideally I'd want to use identical migrations as far as possible.

zelenij:
I'll be happy to create a PR to address this problem.

jzabroski:
Please create.

jzabroski:
@zelenij Just wondering if you do plan to create a PR here.  I am looking at overhauling the SqliteGenerator

zelenij:
[...]

Issue #1063, opened 2019-09-24
Make TypeMap injectable, with fluent TypeMap interface


jayharris:
Is there any way to add or override a specific TypeMap? I would like to make a one-off mapping to map a specific DbType to a specific Template without having to recreate the entire TypeMap class.

For example, SQLite does not support 'DateTimeOffset', but EFCore automatically serializes `DateTimeOffset` to `TEXT` within SQLite. When creating my Migrations, I know that I can use `IfDatabase` to test for SQLite and another to test if not SQLite, creating the appropriate column type for each database engine, but this is a lot of repeated code for each of my DateTimeOffset columns.

Instead, it would be much easier to update the SQLiteTypeMap to somehow fire off `SetTypeMap(DbType.DateTimeOffset, "TEXT");`. This would eliminate all of the `IfDatabase` calls. However, with a lot of internal classes and instantiation within constructors, this seems difficult to do without duplicating most of `FluentMigrator.Runner.SQLite`.

What am I missing? I can't find any documentation for a one-off TypeMap override.

jzabroski:
> What am I missing? I can't find any documentation for a one-off TypeMap override.

In my opinion, it is not a good idea to change the meta-model for how FluentMigrator works across systems.  "EFCore automatically serializes `DateTimeOffset` to `TEXT` within SQLite" isn't sufficient enough justification to break the meta-model.

That said, FluentMigrator 4.0 is customizing a lot more stuff, and it is conceivable this could be on the table to make extensible.  However, breaking a common Meta-Model pretty much defeats the purpose of using FluentMigrator to abstract away syntactic differences among SQL providers.  We try to avoid abstracting away semantic differences.  If SQLite doesn't support it, we'd rather throw an error and have you fix it by providing a custom mapping.

tl;dr: use AsCustom extension method to specify an arbitrary data type. 
 https://fluentmigrator.github.io/api/v3.x/FluentMigrator.Builders.IColumnTypeSyntax-1.html#FluentMigrator_Builders_IColumnTypeSyntax_1_AsCustom_System_String_

jzabroski:
@jayharris  I'll close this - please open a new ticket if you think this should be a feature.  I've given you some context around how we think about it, but I'm always open to being wrong about a point of view/perspective.  True innovation is when customers say, "It doesn't have to be that way."

jzabroski:
[...]

PR #1581, opened 2022-04-12
Limit SQLite type mappings to the 4 core data types

mattbrailsford:
SQLite only really supports 4 core data types `INTEGER, REAL, TEXT, and BLOB` (See https://www.sqlite.org/datatype3.html#storage_classes_and_datatypes) but uses type affinity to allow a form of `alising` (See https://www.sqlite.org/datatype3.html#type_affinity)

Unfortunately `Microsoft.Data.Sqlite` only really supports the 4 core ones (See documentation here https://docs.microsoft.com/en-us/dotnet/standard/data/sqlite/types)

> We recommend only using the four primitive SQLite type names: INTEGER, REAL, TEXT, and BLOB.

...so I think for broader compatibility this type list should be limited to the 4 core supported types. I've updated these definitions as per the MS link above.

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteTypeMap.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteTypeMap.cs
@@ -44,22 +44,22 @@ namespace FluentMigrator.Runner.Generators.SQLite
             SetTypeMap(DbType.UInt16, "INTEGER");
             SetTypeMap(DbType.UInt32, "INTEGER");
             SetTypeMap(DbType.UInt64, "INTEGER");
-            SetTypeMap(DbType.Currency, "NUMERIC");
-            SetTypeMap(DbType.Decimal, "NUMERIC");
-            SetTypeMap(DbType.Double, "NUMERIC");
-            SetTypeMap(DbType.Single, "NUMERIC");
-            SetTypeMap(DbType.VarNumeric, "NUMERIC");
+            SetTypeMap(DbType.Currency, "TEXT");
+            SetTypeMap(DbType.Decimal, "TEXT");
+            SetTypeMap(DbType.Double, "REAL");
+            SetTypeMap(DbType.Single, "REAL");
+            SetTypeMap(DbType.VarNumeric, "TEXT");
             SetTypeMap(DbType.AnsiString, "TEXT");
             SetTypeMap(DbType.String, "TEXT");
             SetTypeMap(DbType.AnsiStringFixedLength, "TEXT");
             SetTypeMap(DbType.StringFixedLength, "TEXT");
 
-            SetTypeMap(DbType.Date, "DATETIME");
-            SetTypeMap(DbType.DateTime, "DATETIME");
-            SetTypeMap(DbType.DateTime2, "DATETIME");
-            SetTypeMap(DbType.Time, "DATETIME");
+            SetTypeMap(DbType.Date, "TEXT");
+            SetTypeMap(DbType.DateTime, "TEXT");
+            SetTypeMap(DbType.DateTime2, "TEXT");
+            SetTypeMap(DbType.Time, "TEXT");
             SetTypeMap(DbType.Boolean, "INTEGER");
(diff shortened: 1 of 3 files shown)


mattbrailsford:
I submitted this PR straight from GitHub so I'll have to pull this down tomorrow and try and resolve the breaking unit tests

jzabroski:
Yikes - Moving to Microsoft.Data.Sqlite was probably a mistake in this regard.  At least, doing so via minor version bump.

jzabroski:
@mattbrailsford Do you have any interest in helping to fill this out? https://docs.google.com/spreadsheets/d/1CWgsOhae-bGYt8fEKkDM7KsUkjN1BqHaRKN4am5HJsc/edit#gid=0

The end goal is to better support aliases, as well as custom mappings viz a viz #1063

mattbrailsford:
Yea, in my local solution I've actually created an `MsSqlite` runner which I'm wondering whether that might actually be a good idea as then people can choose and you aren't attempting to make one runner that works for both.

I'll take a look at the google doc and see what I can add 👍

jzabroski:
The processor tags generally combine a processor (including db driver) and generator (assumes basic things about database version level generally). I don't think creating extra processors is very scalable, although it worked well for the first 15 years of this project. I don't think we anticipated Microsoft writing a SQLite driver, people using that for Entity Framework as that is what that team tests against, etc.

However, in this case, I think simply making type maps injectable would be a lot simpler workaround than creating an MsSqlite runner. It would also then be a lot easier to sunset legacy type maps through major version releases of FluentMigrator.

mattbrailsford:
Sounds fair, so long as Microsoft don’t break other things like the other PR I submitted 😁

For those closer to MS I think the MS provider is probably going to be the go to. Whilst it’s part of EF Core on the MS docs they are touting it as “more .NET Core focused” than the other. I believe there is also something about it running on MacOS too without the need of a rebuild (I may be mistaken here though)

jzabroski:
Thanks.  I will ship this PR along with logic to deprecate the old SQLite behavior for backwards compatibility. Will this address #1456 / any interest in writing tests to make it work?

Also, see: #1492 - If we can upgrade SqliteGenerator to be CompatibilityMode.Strict vs. Loose, I think that will be a safer world to live in going forward. A related topic came up in https://github.com/fluentmigrator/fluentmigrator/discussions/1485 where users are unhappy we set the compatibility mode to loose.

Also, w.r.t. running on Mac OS X, please see: #1490 - I'm not sure if this is still an active problem or has been addressed via Microsoft.Data.Sqlite hotfix.

Think if we can tackle all the above we can release this as 4.0 and people should be very happy.

mattbrailsford:
Sure. I think I can take a look at some of these. 

Re CompatabiltyMode is there more info on this. I did read that issue before but I didn’t really understand it.

jzabroski:
@mattbrailsford To be honest, Compatibility Mode is a crappy concept. I am not sure what it's original goal was, but I think the idea was that FluentMigrator should support a set of verbs and nouns, but not all databases are robust enough to support all those verb-noun relationships. Also, the matrix of what verb-noun relationships are supported changes over time for each provider/generator, so hard-coding it is a bad idea.  I think a lot of the existing Sqlite code was written back for Sqlite 1.6?

nickeeex:
After updating to FluentMigrator 5.0 I got hit with this issue.

This feels like a breaking change?

We've used dapper with system.data.sqlite and there it's worked perfectly fine to do something like 
`connection.Query<(Guid Id, int Version, string Data)>`
Where the Id was of type "UNIQUEIDENTIFIER". Now as it's type TEXT and it throws an exception when you try to cast string to guid.

jzabroski:
@nickeeex Did you consider opening an issue Rather than commenting on a PR that is closed. Thank you for following our process in advance.

nickeeex:
> @nickeeex Did you consider opening an issue Rather than commenting on a PR that is closed. Thank you for following our process in advance.

Wasn't really sure if this was a question or bug or more of a general heads up. But i'll add a issue :)

PR #1583, opened 2022-04-15
Fixes SQLite FormatIdentity bug 1456

### LINKED ISSUE #1456: Sqlite Identity column is required to be int32

mattbrailsford:
Fixes #1456

The previous implementation of `FormatIdentity` incorrectly limited identity support to `DbType.Int32` when in reality, it should be allowed for anything that maps to "INTEGER".

I've updated it so it checks the DbType against the TypeMap and have also separated it from the `FormatPrimaryKey` function as you should be able to have a primary key without identity too.

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs
@@ -29,14 +29,19 @@ namespace FluentMigrator.Runner.Generators.SQLite
         /// <inheritdoc />
         protected override string FormatIdentity(ColumnDefinition column)
         {
-            // SQLite only supports the concept of Identity in combination with a single integer primary key
-            // see: http://www.sqlite.org/syntaxdiagrams.html#column-constraint syntax details
-            if (column.IsIdentity && !column.IsPrimaryKey && (!column.Type.HasValue || GetTypeMap(column.Type.Value, null, null) != "INTEGER"))
+            if (column.IsIdentity)
             {
-                throw new ArgumentException("SQLite only supports identity on a single integer, primary key coulmns");
+                // SQLite only supports the concept of Identity in combination with a single integer primary key
+                // see: http://www.sqlite.org/syntaxdiagrams.html#column-constraint syntax details
+                if (!column.IsPrimaryKey && (!column.Type.HasValue || GetTypeMap(column.Type.Value, null, null) != "INTEGER"))
+                {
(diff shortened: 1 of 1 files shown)


volkanceylan:
Unfortunately, this pull request breaks existing migrations, now it produces something like:

`CREATE TABLE "SomeTable" ("Id" INTEGER NOT NULL AUTOINCREMENT....`

which results in a syntax error. I can't update to 5.x because of this reason. I'm using Sqlite primarily in tests, and migrations that work in SqlServer/Postgres/MySql etc. now fails in Sqlite.

I think the original author tried to apply a workaround for Sqlite to make ordinary migrations with an Identity column (not primary key) and other composite etc. primary keys.

I understand you tried to support types other than Int32, but did you test this change against set of existing migrations etc? I want to be sure before sending a pull req to revert this one.

volkanceylan:
Hmm I might be wrong, seems like the file is changed several times after this pull request. 

Anyway, here is a sample migration that is failing in 5.0.0 (only with Sqlite) but not in 3.3.2:

```cs
        Create.Table("UserExternalLogins")
            .WithColumn("UserExternalLoginId").AsInt32().Identity().NotNullable()
            .WithColumn("LoginProvider").AsString(100).NotNullable().PrimaryKey()
            .WithColumn("ProviderKey").AsString(200).NotNullable().PrimaryKey()
```

jzabroski:
Thanks you for the reproduction. Can you please open a new ticket

Issue #1218, opened 2020-04-12
SapAnywhere tests need separate project


jzabroski:
SapAnywhere is not support on .NET Core by SAP, and the test runner cannot run on .NET Core due to this limitation.

As part of #1178 , I've littered the code with `#if NET461` checks.  I think Phase 2 should be to move the FluentMigrator.Tests related to SapAnywhere into their own csproj, and only run them on net461.  This will likely make it more readable, and easier to delete if we eventually drop support for SapAnywhere.

@eloekset Do you have any insight into whether SAP is still supporting SapAnywhere?  My understanding is its mostly used in German and Scandinavian countries, so I thought you might be more an expert than me, as I've never used SAP-anything-anywhere-any-time :)

eloekset:
I'm afraid it has been dead technology for many years. The previous company I was working for still uses it for the on-prem software, so that's why I added support for SQL Anywhere to FluentMigrator. However, I don't work there any more, and I doubt that I'll ever get into this DB again.

I had contact with some internals of SQL Anywhere, and I mentioned this project and that I had built support for their technology, but they didn't show any interest. I doubt that they will ever support .NET Core, as there was hardly support for the original Entity Framework.

BTW, here's a request to have Microsoft build EfCore drivers for SQL Anywhere, something that they of course won't spend resources on: https://github.com/dotnet/efcore/issues/12153 I suppose it's up to those who still depends on SQL Anywhere to build a community driver.

I support your suggestion to move the tests into a separate csproj.

eloekset:
If we restructure dependencies, [as described in this thread](https://github.com/fluentmigrator/fluentmigrator/issues/1164), we'll some time in the future be able to get insight into the usage per provider from the download count at nuget.org.

![image](https://user-images.githubusercontent.com/791721/79107031-f0082c80-7d73-11ea-9439-048ae9e60980.png)

If some providers stay at 0 downloads for a few version, we could consider dropping the support for that provider.

eloekset:
This is the latest SQL Anywhere stuff of NuGet. Version 17 of the driver is built for .NET Framework, and it may or may not work for .NET Core 3.X apps. It depends on the internal API calls, whether they are available in .NET Core 3.X. Pre .NET Standard 2.0, it would not even be [possible to install a package like this into a .NET Core application](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-framework-compatibility-mode).

[...]

Issue #1211, opened 2020-04-09
Tags not working correctly with runner in 3.2.4


mmiller678:
The runner is complaining that no migrations exist when a tag is specified.  I think is because its only picking up migrations with that tag, rather than migrations with the tag + migrations without any tags.

Exception
```
While executing migrations the following error was encountered: No migrations found,    at FluentMigrator.Runner.DefaultMigrationInformationLoader.FindMigrations(IMigrationSource source, IMigrationRunnerConventions conventions, String namespace, Boolean loadNestedNamespaces, String[] tagsToMatch, Boolean includeUntagged) in /home/travis/build/fluentmigrator/fluentmigra
tor/src/FluentMigrator.Runner/DefaultMigrationInformationLoader.cs:line 185 [D:\Source\Panda\trunk\build.targets]
D:\Source\Panda\trunk\Tools\packages\Weiss.MSBuild.1.2.54.1137\content\Weiss.Core.targets(91,5): error :    at FluentMigrator.Runner.DefaultMigrationInformationLoader.LoadMigrations() in /home/travis/build/fluentmigrator/fluentmigrator/src/FluentMigrator.Runner/DefaultMigrationInformationLoader.cs:line 126 [D:\Source\Panda\trunk\build.targets]
```

jzabroski:
I think this is a false positive.  This can happen is you run FluentMigrator.MSBuild 3.2.1 task with an assembly referencing FluentMigrator 3.2.4 nuget package.

I played around with reproducing the issue here: https://github.com/fluentmigrator/fluentmigrator/issues/1211

The only way, in my judgment, this can happen, is if the wrong assembly version is loaded.  Why it fails this way is a bit beyond me at the moment (midnight).  I will follow up with more, tomorrow maybe.

mmiller678:
Agreed - this was a mistake on my part.  We can close this.

Issue #1208, opened 2020-04-07
Run migrations in-process without DI


taprm:
Hello!
Is there any chance to create MigrationRunner directly without using DI to apply migrations?
Thanks

jzabroski:
I'm not sure exactly what problem you're trying to solve?

taprm:
> I'm not sure exactly what problem you're trying to solve?

I have an asp.net core application, with autofac as a DI. So I want my application to apply all migrations during start. How can I create necessary instances of migration runners in my case? I thought that there must be a way to use migration runners without dependency injection at all

jzabroski:
I think the path forward here is to pull the example out of the examples folder and turn it into an FluentMigratorApplication abstract base class.  The Application class would hook into whatever DI you inject.  In this way, we could simplify building in-process toolchains which is the current recommended guidance, anyway, given #1016 

Your code would then look like:

```c#
public class AspNetFluentMigratorApp : FluentMigratorApplication
{
  public AspNetFluentMigratorApp(IMigrationRunner migrationRunner) : (migrationRunner) {
    this.Start();
  }
}
```

@fubar-coder What do you think?  It's just pseudo-code at this point and missing the hard battle of getting it working.

jzabroski:
Related (old) question on StackOverflow: https://stackoverflow.com/questions/7574417/is-it-possible-to-use-fluent-migrator-in-application-start

jzabroski:
Technically, some of this infrastructure already pretty much exists, viz a viz:

https://github.com/fluentmigrator/fluentmigrator/blob/782e12cda168c9b94ba5647e3edf713d7a738e03/src/FluentMigrator.DotNet.Cli/Setup.cs#L37-L43

What would need to be cleaned up is the bits where this logic calls BuildServiceProvider, as the [ASP.NET Runtime Host needs to be in charge of doing that correctly](https://twitter.com/davidfowl/status/1038521947007594498), or  the ServiceCollection will get transformed into a ServiceProvider multiple times.  When that happens, each ServiceProvider can have a reference to a different universe of types and some of the "services" may be disposed of by the time FluentMigrator (or the ASP.NET web page) get a chance to use those services.

In other words, the following line needs to change from returning an IServiceProvider to an IServiceCollection, so that the caller can decide whether they need to kick the job of calling `BuildServiceProvider()` to a runtime or handle it themselves:

[...]

Issue #1006, opened 2019-03-26
Why is in-process runner preferred? [question] 


fredrik-lundin:
Hi,

I'm curious to know more about that reasoning behind the comment in the _Migration runners docs_ saying ([https://fluentmigrator.github.io/articles/migration-runners.html?tabs=vs-pkg-manager-console](https://fluentmigrator.github.io/articles/migration-runners.html?tabs=vs-pkg-manager-console)):

> Please use the in-process runner if possible.

What are the pros and cons with using the in-process instead of for example the _dotnet tool_? 

Think would usually be a question for stack overflow or similar, but since the docs so clearly recommends an approach over another, it would be nice with some comments on why 😊

fubar-coder:
TL/DR: DLL dependencies are a real problem when using the external runners.

Long version:

The main reason why I recommend the in-process runner is, that running the migrations from your own application works better than using the external migration runner. The external runners use a simplistic approach to find all migrations (DLL dependencies) and sometimes fail when they're unable to find all dependencies.

Advantages of the in-process runner:

- Works (no problems with dependent DLLs)
- Fully customizable (external runners may not expose all features)
- No external tools required, everything works out-of-the-box

Advantages of the external runners:

- DB migrations can be run during the installation process
- Migrations can be kept separate from the main application

The advantages of one are the disadvantages of the other 😁 .

jzabroski:
@fredrik-lundin  I was wondering the same thing. Thanks for asking.

@fubar-coder Would it be far to say that Nate McMaster's [blog post](https://natemcmaster.com/blog/2018/07/25/netcore-plugins/) about what's broken with .NET Assembly loading in older versions of .NET applies to MSBuild (aka "external runners")?  In essence, the issue is not external runners vs. in-process runners, and everything to do with how bad the "assembly loading experience" is with external runners like MSBuild, since:

1. External Runners like MSBuild don't support bindingRedirect(s)
2. There are two distinct problems to solve with assembly binding: locating the assembly, and loading it.  Due to how the CLR is architected, the assembly has to be loaded or an expression will raise a TypeLoadException if the assembly hasn't been loaded in time.

fubar-coder:
@jzabroski Interesting. I haven't seen this blog post before. Maybe it can solve some of the problems I encountered. I'm willing to give this a shot. This would be only a partial fix, because this would only work with the `dotnet-fm` tool and not with the msbuild runner.

jzabroski:
[...]

PR #1600, opened 2022-05-05
Move all FluentMigration.Runner code to FluentMigrator.Runner.Core

mattbrailsford:
This is a re-implementation of PR #1198 but does not remove any legacy / obsolete code. This should result in the same outcome, that FluentMigrator.Runner is now just a meta package with a single extensions method for adding all database types to the `IMigrationRunnerBuilder` but all other code has moved to `FluentMigrator.Runner.Core` thus meaning folks shouldn't need to install all database providers.

The only real changes I've had to make are

1. The `AddAllDatabases` extension method in `FluentMigrator.Runner` has changed from being a `IServiceProvider` extension method to an `IFluentMigratorRunnerBuilder` extension.
2. The `FluentMigratorServiceCollectionExtensions.CreateServices` method now no longer calls `AddAllDatabases` and instead accepts a `Action<IMigrationRunnerBuilder> configureRunner` to allow external configuration and to call the `AddAllDatabases()` extension
3. `TaskExecutor` now accepts an option `Action<IMigrationRunnerBuilder> configureRunner` that is passed through to the `CreateServices` method above.
4. All `AddDbType` methods have been updated to register the provider factory with the `MigrationProcessorFactoryProvider` rather than the `MigrationProcessorFactoryProvider` being responsible for adding them all.


mattbrailsford:
Hmm, not entirely sure why the tests are failing. They passed on the first commit and the other commits are minor changes that shouldn't affect this 🤔

jzabroski:
This seems nice, based on your comments/description of the changes. I have not looked at the code changes yet.

I'm in the process of updating the example Runner.

Ideally, we also tackle #1006 , in particular #1208 - See my comments here: https://github.com/fluentmigrator/fluentmigrator/issues/1208#issuecomment-611071082

Fully closing #1006 would probably entail smoothing out how we package FluentMigrator.MSBuild, as the way it is bundled today is incomplete and there should probably be a second "batteries included" version of FluentMigrator.MSBuild package.

I think we may want to obsolete the legacy Console as well for either 4.0 or 5.0, but that would currently affect FluentMigrator.MSBuild I think (sorry for cryptic comment - just thinking out loud notes for me to check on).

jzabroski:
I think we should update:

https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Console/Program.cs#L32

to include a warning message (in yellow ANSI Color code) at the entry point:

```c#
System.Console.WriteLine($"WARN: FluentMigrator.Console is obsolete, and will be going away in 5.0.0 release.{System.Environment.NewLine}Please check the upgrade guide http://fluentmigrator.github.io/....");
```
Something like that.

jzabroski:
Additional Notes to self: Pull over my comments from #1198 

> The economic value it adds is:

> 
> 1. Ability to write Roslyn Analyzers that suggest database-specific best practices based on which Runners the users import.
> 2. Ability to track via nuget.org the popularity of database-specific runners and therefore where to focus efforts of FluentMigrator improvements by popularity (not that popularity is the only consideration, as die hard/passionate users are often the best ways to build community and improve projects - @fubar-coder , @eloekset and myself fall in the die hard camp and that's how we all got involved - we just couldn't stand to see the project not be actively maintained)
> 3. Ability to split out unit tests per runner - I started doing this as part of https://github.com/fluentmigrator/fluentmigrator/issues/1218 - with the major advantage being that if some vendors/projects fall behind in supporting the .NET releases, the core branch isn't over-complicated with #if NETFRAMEWORK and other pre-processor directives and so on.
>     1. My thought is that I could then generate a report for which runners cover which "test drivers", and have an "architecture verification test" for saying things like "SQLite doesn't support partitions, so ignore tests related to partitions and partitioned indexes" and "SQL Server seems to be missing tests for partitioned indexes but has create/alter partition tests."

> 
> Previously, nobody could give me great reason to do this other than "I hate files in my directory".

> 
> Remaining proof-of-concept is re-bundling FluentMigrator.Console and verifying it ships all runners still. I am assuming you expect FluentMigrator.DotNet.Cli to also ship everything, and that this change is mainly to support using the In-Process Runner?

and

> what happens if you upgrade to FluentMigrator 3.2.8 locally but your build server is on 3.2.1. I believe the behavior will depend on RollForward behavior, and that any runtime dependency mismatches will not get caught until you go to production. Do you have any suggestions on how we can detect such mismatches and avoid them? See https://github.com/fluentmigrator/fluentmigrator/issues/1211 for an example where a customer was bitten by this.

along with:

> The downside I see to hot loading types in RunWithServices is that you move away from deps.json auto-resolving the dependency, so you can end up in situations where a developer has upgraded to a newer version of fluentmigrator, but the build ci/deploy cd servers are on older versions, and thus you get a runtime error due to not resolving dependencies.

I gave this a lot of thought over the years.  Perhaps for 4.0 we not address it.  Ideally we can analyze the deps. I believe .NET 6.0 has APIs to do this.

mattbrailsford:
Not sure if there is anything you want me to action in there, or if it's just notes for your own review, but let me know if there is anything you need me to implement, otherwise I'll wait for a code review.

If based on your comments though this is already being tackled in another area, it would be good if some info could be placed somewhere as it's hard to track the status of the project and where to direct efforts to assist.

I wonder if some kanban board or something is needed to know the status of the next release and what is still required to be done and how anyone can help.

jzabroski:
These are just notes for me to move into documentation/roadmap pages. 

> If based on your comments though this is already being tackled in another area, it would be good if some info could be placed somewhere as it's hard to track the status of the project and where to direct efforts to assist.

No, this PR is needed.  The notes are just the "economic benefits" and what this would enable long-term, plus a few minor dependency issues I am aware of that are tangentially related.

jzabroski:
Spoke to Simon Cropp.  He told me a lot of popular nuget packages, like Newtonsoft JSON and Serilog, lock the assembly version at the major version number, even though the package might use servicing releases in the SemVer. This will address https://github.com/fluentmigrator/fluentmigrator/issues/1211 mentioned above in this comment: https://github.com/fluentmigrator/fluentmigrator/pull/1600#issuecomment-1118603094

@fubar-coder  Thoughts on this particular solution to the "reflection can't find MigrationAttribute" issue when dynamically loading assemblies via out-of-process runners?

fubar-coder:
@jzabroski This problem usually happens when you're referencing (and using) two different versions of the assembly containing the attribute, which means that the type information instance is different even though it's the same type/code (but from 2 different assemblies).

There are two possible workarounds:

- Avoid loading both assembly assemblies by overriding the assembly resolver
- Compare by type name (and namespace)

jzabroski:
@mattbrailsford So, here are my takeaways from the PR (for release notes, and possibly questions/remarks):

1. Announcers are moving from FluentMigrator.Runner to FluentMigrator.Runner.Core

jzabroski:
Should wrap up reviewing this week.  Got buried with life and work.

jzabroski:
@mattbrailsford OK, I think my review is pretty much done.  Please take a look - it's pretty minor stuff I think, but just "fit and finish" polish before we release.

mattbrailsford:
@jzabroski just a couple of things

1) Most of your review is really code that existed prior to this PR (I just moved it) so by changing that stuff it kinda expands the scope of this PR. I understand those things could do with resolving, but isn't that more of an overall project/v4 concern than a blocker for this PR?

2) If we do address these issues, how far down that rabbit hole do you want to go? as removing `MigrationProcessorFactoryProvider` effectively means removing the whole legacy API no? If that is the case, again, that feels out of scope for this PR, and personally I'm not sure I know the project well enough to be confortable removing all that code as I'm not that sure where / how it is all used.

jzabroski:
@mattbrailsford I see your point of view, and acknowledge your POV makes sense.

danielkraut:
> @mattbrailsford I'll start drafting the release notes and will target 4.0 release this weekend. I don't want to drag it out seeking perfection. Plus, decoupling the runner is a big accomplishment

Hi, what is the current state and do you have a plan when we can expect the 4.0 release? I know you are probably busy but the latest comment about the release is from June 7th.

jzabroski:
I need to release it.  I just spent last weekend fixing fluentmigrator/documentation in prep for 4.0 release.

orientalpers:
I am waiting too

robertcoltheart:
I got tired of waiting and made this: https://www.nuget.org/packages/FluentMigratorBridge which lets you use just a single database runner instead of referencing them all.

dariogriffo:
> I got tired of waiting and made this: https://www.nuget.org/packages/FluentMigratorBridge which lets you use just a single database runner instead of referencing them all.

Is this a fork of 3.3.2?

robertcoltheart:
No, but it is compatible with FluentMigrator 3.3.2. It's a library that allows you to create migrations and a runner with just a single database, instead of pulling in all the libraries for all databases. Instructions are on the github repo.

dariogriffo:
> No, but it is compatible with FluentMigrator 3.3.2. It's a library that allows you to create migrations and a runner with just a single database, instead of pulling in all the libraries for all databases. Instructions are on the github repo.

Starred the repo, amazing job @robertcoltheart thank you so much. Worked like a charmed.
Below my git diff, didn't change a single line of code
image

jzabroski:
I should be pushing 4.0.0 shortly.

PR #1656, opened 2023-04-04
feat: add dotnet 7.0 support for cli tool

PauloDaniel1993:
Add support for dotnet 7.0 in FluentMigrator.DotNet.Cli by adding net7.0 as target framework.


vmandic:
+1 would love to have this, so as a new release supporting this.

Issue #1657, opened 2023-05-05
SQLite: Support binary GUIDs


lrobin136:
We're using the FluentMigrator 3.3.2. System.data.sqlite.core 1.0.113.7 allows us to store guids as binary data with the connection string option "BinaryGuid=True". 

If we use in our migrations: 

``` 
Insert.IntoTable("Table").Row(new { GUID_= Guid.NewGuid() } )
```

a guid is inserted as text string because of the [GenericQuoter](https://github.com/fluentmigrator/fluentmigrator/blob/65f79eb7353d2db8fab0aed48311d4bc87ac04d8/src/FluentMigrator.Runner.Core/Generators/Generic/GenericQuoter.cs#L132):

```
        public virtual string FormatGuid(Guid value)
        {
            return ValueQuote + value.ToString() + ValueQuote;
        }
```

We would like to have a SQLiteQuoter that optionally formats guids directly as hexadecimal string values:

```
        public override string FormatGuid(Guid value) 
        {
            return "x'" + value.ToString("N", CultureInfo.InvariantCulture) + "'"; 
        }
```
Would it be possible to implement this missing feature? 

Right now, we have a workaround that inserts our own SQLiteQuoter implementation into the [SQLiteProcessor](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Runner.SQLite/Processors/SQLite/SQLiteProcessor.cs), but the quoter is not passed on to the [SQLiteColumn](https://github.com/fluentmigrator/fluentmigrator/blob/master/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs) and the solution is not entirely satisfactory.

lrobin136:
Would you be interested if we submitted a PR to support binary guids with SQLite?

jzabroski:
Hi @lrobin136 . Thanks for your patience. Yes, absolutely. 4.0.0 branch is basically in the bag, we could ship it as 4.1.0 feature.

jzabroski:
@lrobin136 As this depends on the caller to know the convention set by the connection string, I will implement this as:

```c#
        public bool BinaryGuid { get; set; } = false;

        /// <inheritdoc />
        public override string FormatGuid(Guid value)
        {
            if (BinaryGuid)
                return "x'" + value.ToString("N", CultureInfo.InvariantCulture) + "'";
            return base.FormatGuid(value);
        }
```

[...]

PR #1712, opened 2024-01-09
Use configured quoter in SQLiteColumn

lrobin136:
PR for the issue:  #1657

-  Use configured quoter in SQLiteColumn to support injecting non-default quoter

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs
@@ -9,8 +9,8 @@ namespace FluentMigrator.Runner.Generators.SQLite
     // ReSharper disable once InconsistentNaming
     internal class SQLiteColumn : ColumnBase
     {
-        public SQLiteColumn()
-            : base(new SQLiteTypeMap(), new SQLiteQuoter())
+        public SQLiteColumn(IQuoter quoter)
+            : base(new SQLiteTypeMap(), quoter)
         {
             // Add UNIQUE before IDENTITY and after PRIMARY KEY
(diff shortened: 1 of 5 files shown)


jzabroski:
1. Create an ISQLiteQuoter, and have the existing concrete SQLiteQuoter and whatever one you wish to inject yourself added there. SQLiteGenerator would then take an ISQLiteQuoter, and the AddSQLite() DI helper would inject the concrete SQLiteQuoter by default
     1. This would serve two purposes - allowing others to create their own custom quoter, and also allowing FluentMigrator.Tests assembly to have a clear contract on the scope of regression testing by defining the set of SQLite-applicable `IQuoter`s.
3. This PR is missing the BinaryGuid formatting logic and associated regression tests, which I thought was the point of the original PR idea.
    1. As discussed previously, add a UseBinaryGuid flag to the existing quoter, and the default constructor for SQLiteQuoter would not break existing clients.  Then you would just need a non-ABI breaking change to AddSQLite DI helper, which would allow you to configure UseBinaryGuid = true.  The regression tests could then handle quoting both ways.  You can see how this is done for Snowflake Quoter in the Snowflake tests, as it regression tests both scenarios.

Overall, this would be the ideal API and maximize community benefit.  The changes as presented seem to only immediately benefit your organization, and anyone who then wants to re-implement the same logic.

lrobin136:
Thanks for your feedback.

The new implementation contains the BinaryGiud formatting logic and  a regression test that ensures that an inserted GUID can be read again. The guid formatting logic is different from the initial approach because of the convention how guids are stored with SQLite and .NET. 

The formatting logic can be configured in AddSQLite() DI using an optional flag.

jzabroski:
Thanks, I need to fix a breaking change with 5.0.0 related to SQLite STRICT table behavior, and then I will ship 5.1.0 hotfix.

Issue #1737, opened 2024-02-20
Sqlite AutoIncrement with other primary keys broken in 5.x


volkanceylan:
**Describe the bug**
A migration for a table that includes an Identity column and one or more other primary key columns are broken for Sqlite in 5.0.0 and 5.1.0 versions.

**To Reproduce**
The following program works fine with FluentMigrator.Runner 3.3.2 and Microsoft.Data.Sqlite 8.0.2:

```cs
using FluentMigrator;
using FluentMigrator.Runner;
using Microsoft.Extensions.DependencyInjection;

namespace SqliteAutoIncrement;

class Program
{
    const string dbName = @".\mydb.db";
    const string connectionString = $"Data Source={dbName}";

    static void Main()
    {
        File.Delete(dbName);

        var serviceProvider = new ServiceCollection()
            .AddFluentMigratorCore()
            .ConfigureRunner(rb =>
            {
                rb.AddSQLite()
                  .WithGlobalConnectionString(connectionString)
                  .ScanIn(typeof(Program).Assembly).For.Migrations();
            })
            .BuildServiceProvider(false);

        var runner = serviceProvider.GetRequiredService<IMigrationRunner>();
        runner.MigrateUp();
    }
}

[Migration(1)]
public class MyMigration: AutoReversingMigration
{
    public override void Up()
    {
        Create.Table("MyTable")
            .WithColumn("Id").AsInt32().Identity().NotNullable()
            .WithColumn("Key1").AsString(100).NotNullable().PrimaryKey()
            .WithColumn("Key2").AsString(100).NotNullable().PrimaryKey();
    }
}
```

But if FluentMigrator 5.0.0 or 5.1.0 is used results in the following exception:
```
System.Exception: 'SQLite Error 1: 'near "AUTOINCREMENT": syntax error'.
While Processing:
"
CREATE TABLE "MyTable" (
   "Id" INTEGER NOT NULL AUTOINCREMENT, 
   "Key1" TEXT NOT NULL, 
   "Key2" TEXT NOT NULL, 
   CONSTRAINT "PK_MyTable" PRIMARY KEY ("Key1", "Key2")
)
"'
```

**Expected behavior**
[...]

PR #1744, opened 2024-03-15
#1737: Don't apply AUTOINCREMENT on non-identity PK

schambers:
Fixes regression introduced where non-identity, primary key columns always attempt to create with "AUTOINCREMENT" which is only valid for integer identity columns. Fixes issue #1737


schambers:
If there's no objections here @jzabroski, I'll merge this, this week and close it out

jzabroski (commented):
Thank you!

Issue #1724, opened 2024-01-30
FluentMigrator 5.0.0 and SQLite - Column Type Defining as TEXT Instead of DATETIME 


Rowandish:
## Problem Description
Since version 5.0.0 of FluentMigrator, when defining a column using `AsDateTime()` in migrations, the column is created as type TEXT instead of DATETIME in SQLite. The same code works correctly with FluentMigrator version 3.2.2.

## Steps to Reproduce
Define a migration that defines a column using `AsDateTime()`. When this migration is run with FluentMigrator 5.0.0, the column is created as TEXT instead of DateTime in SQlite.

## Expected Behavior
The column should be created as type DATETIME when using `AsDateTime()`.

## Actual Behavior
The column is created as type TEXT instead of DATETIME with FluentMigrator 5.0.0.

## Environment
* FluentMigrator Version: 5.0.0
* System.Data.SQLite Version: 1.0.118
* Operating System: Windows 11
* Database: SQLite
* .NET Version: .NET8

## Console Application
I have attached a self-contained single-file console application that reproduces the issue for clarity.

```csharp
using System.Data.SQLite;
using FluentMigrator;
using FluentMigrator.Runner;
using Microsoft.Extensions.DependencyInjection;

class Program
{
    static void Main()
    {
        const string dbName = @".\mydatabase.db";
        const string connectionString = $"Data Source={dbName};Version=3;";
        File.Delete(dbName);

        var serviceProvider = new ServiceCollection()
            .AddFluentMigratorCore()
            .ConfigureRunner(rb =>
            {
                rb.AddSQLite()
                  .WithGlobalConnectionString(connectionString)
                  .ScanIn(typeof(Program).Assembly).For.Migrations();
            })
            .BuildServiceProvider(false);
        var runner = serviceProvider.GetRequiredService<IMigrationRunner>();
        runner.MigrateUp();

        var realColType = VerifyColumnType(connectionString);
        if (realColType != "DATETIME")
            throw new Exception();
    }

    private static string VerifyColumnType(string connectionString)
    {
        using var connection = new SQLiteConnection(connectionString);
        connection.Open();

        using var command = new SQLiteCommand("SELECT type FROM PRAGMA_TABLE_INFO(\"MyTable\") WHERE name = \"MyColumn\"", connection);
        return (string) command.ExecuteScalar();
    }
}

[Migration(42)]
public class DateTimeMigration : Migration
{
    public override void Up()
    {
[...]

Issue #1618, opened 2022-07-19
Add STRICT Tables for SQLite


dittodhole:
With SQLite 3.37.0 the [type affinity](https://www.sqlite.org/datatype3.html) of SQLite got enhanced by [STRICT Tables](https://www.sqlite.org/stricttables.html).

Any chance to opt-in?

jzabroski:
1. This should use FluentMigrator.Infrastructure.ISupportAdditionalFeatures.
2. There is no way currently to specify table-level options. There needs to be a new `ICreateTableWithOptionOrWithColumnOrSchemaSyntax`. This new interface needs to implement ICreateTableWithColumnOrSchemaSyntax.  This should probably look like:
  ```c#
  public interface ICreateTableOptionsSyntax : IFluentSyntax
  {
    /// <summary>
    /// Defines the table as having a tablespace (partitions).
    /// </summary>
    /// <returns>Defines the column or schema for the table</returns>
    ICreateTableWithColumnOrSchemaSyntax Tablespace(); // <-- this doesn't exist yet, but it could...
  }
  ```
  and `CreateTableExpressionBuilder` should implement `ICreateTableOptionsSyntax`
  ```c#
        /// <inheritdoc />
        ICreateTableOptionsSyntax WithOptions()
        {
            return this;
        }
  ```
3. Need to update SQLiteGenerator `public override string Generate(CreateTableExpression expression)` to support appending this if it is set & update SQLiteGeneratorTests to test both scenarios.

The SQLiteGenerator should then be customized accordingly to use the value.:
```c#
private static readonly HashSet<string> _supportedAdditionalFeatures = new HashSet<string>
{
    SQLiteExtensions.StrictTable,
};
```
Recommend features are organized via a Extensions file:

```c#
public static partial class SQLiteExtensions
{
   public const string StrictTable = "SQLiteStrictTable";

  private static TNext SetStrict<TNext, TNextFk>(ITableOptionSyntax<TNext, TNextFk> expression, PostgresGenerationType? generation, PostgresIdentityModificationType modificationType, ISupportAdditionalFeatures castColumn)
    where TNext : IFluentSyntax
    where TNextFk : IFluentSyntax
{
    if (generation.HasValue)
        castColumn.AdditionalFeatures[IdentityGeneration] = generation;

    castColumn.AdditionalFeatures[IdentityModificationType] = modificationType;

    return expression.Identity();
}
}
```

jzabroski:
[...]

PR #1754, opened 2024-03-24
Make TypeMap Injectable

### LINKED ISSUE #1063: Make TypeMap injectable, with fluent TypeMap interface

jzabroski:
Fixes #1063, #1724 and #1618.

Possible breaking changes include:
- `ColumnBase` now raises an obsolete warning: "Use ColumnBase<TTypeMap> instead. Going forward, column models must have a db-specific type map model."
- Various Generators have new constructor signatures added or replaced to support database-specific type maps.


jzabroski:
Closing this as for some reason the pipeline didnt trigger.

Issue #1800, opened 2024-05-12
Creating PrimaryKeys constraint get different constraint names


Sherif-Ahmed:
**Describe the bug**
Creating PrimaryKeys constraint get different constraint names

**To Reproduce**
I have an initial migration for table creation
```
public class InitialMigration : ForwardOnlyMigration
{
    public override void Up()
    {
        Create.Table("MyTable")
            .WithColumn("Id").AsInt32().NotNullable().PrimaryKey()
            .WithColumn("Name").AsString(400).NotNullable();
    }
}
```

**This will generate a constraint name "PK_MyTable"**, and in a later migration I have to re-create the primary key because of a table rename
```
public class InitialMigration : ForwardOnlyMigration
{
    public override void Up()
    {
        Delete.PrimaryKey("PK_MyTable").FromTable("MyTable");
        Rename.Table("MyTable").To("NewTableName");
        Create.PrimaryKey().OnTable("NewTableName").Column("Id");
    }
}
```

**This will generate a constraint name "PK_NewTableName_Id"**
Is this intended or I had an issue.

**Expected behavior**
Same constraint name convention for example "PK_NewTableName" based on  `DefaultPrimaryKeyNameConvention`

**Information (please complete the following information):**
 - OS: Windows 11
 - Platform: .NET 8.0.4
 - FluentMigrator: 5.2.0 "in-process runner"
 - Database Management System: SQLServer

TheSquidCombatant:
Obviously, the observed difference in the name of the primary key occurs due to the fact that, when creating a table, the `DefaultConstraintNameConvention` class is used to generate the name, and when creating a constraint, the `DefaultPrimaryKeyNameConvention` class is used to generate the name.

I agree that it would be correct for the name for the primary key to be generated the same in all cases. I have created a pull request with the appropriate fix, please check https://github.com/fluentmigrator/fluentmigrator/pull/1803.

I won’t reassure you and I’ll say right away that I’m not a maintainer. For now, I suggest keeping in mind that the default primary key name is not guaranteed. And each time explicitly set the name of the primary key.

jzabroski:
This would be a breaking change, but seems reasonable. For now you can implement a different naming convention as a workaround.

jzabroski:
The root cause here is that we're using CreateConstraintExpression.

[...]

PR #1803, opened 2024-05-14
Names for the primary key would be generated the same in all cases

TheSquidCombatant:
Eliminated the difference between classes `DefaultConstraintNameConvention` and `DefaultPrimaryKeyNameConvention` in the default generated primary key name. See issue https://github.com/fluentmigrator/fluentmigrator/issues/1800.

### DIFF
--- a/src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
+++ b/src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
@@ -14,6 +14,9 @@
 // limitations under the License.
 #endregion
 
+using System;
+using System.Linq;
+
 using FluentMigrator.Expressions;
 
 namespace FluentMigrator.Runner.Conventions
@@ -29,23 +32,24 @@ namespace FluentMigrator.Runner.Conventions
         /// <inheritdoc />
         public IColumnsExpression Apply(IColumnsExpression expression)
         {
+            if (expression.Columns.Count(x => x.IsPrimaryKey) > 1)
+                throw new InvalidOperationException("Error creating table with multiple primary keys.");
+
             foreach (var columnDefinition in expression.Columns)
             {
+                if (string.IsNullOrEmpty(columnDefinition.Name))
+                    throw new ArgumentException("An object or column name is missing or empty.");
+
                 if (columnDefinition.IsPrimaryKey && string.IsNullOrEmpty(columnDefinition.PrimaryKeyName))
                 {
                     var tableName = string.IsNullOrEmpty(columnDefinition.TableName)
                         ? expression.TableName
                         : columnDefinition.TableName;
-                    columnDefinition.PrimaryKeyName = GetPrimaryKeyName(tableName);
+                    columnDefinition.PrimaryKeyName = $"PK_{tableName}_{columnDefinition.Name}";
                 }
             }
 
             return expression;
-        }
-
-        private static string GetPrimaryKeyName(string tableName)
-        {
-            return "PK_" + tableName;
         }
     }
 }
--- a/test/FluentMigrator.Tests/Unit/DefaultMigrationConventionsTests.cs
+++ b/test/FluentMigrator.Tests/Unit/DefaultMigrationConventionsTests.cs
@@ -46,7 +46,7 @@ namespace FluentMigrator.Tests.Unit
         private static readonly IMigrationRunnerConventions _default = DefaultMigrationRunnerConventions.Instance;
 
         [Test]
-        public void GetPrimaryKeyNamePrefixeForColumnWithName()
+        public void GetPrimaryKeyNamePrefixForColumnWithName()
         {
             var expr = new CreateColumnExpression()
             {
@@ -64,7 +64,7 @@ namespace FluentMigrator.Tests.Unit
         }
 
         [Test]
-        public void GetPrimaryKeyNamePrefixeForColumnWithoutName()
+        public void GetPrimaryKeyNamePrefixForColumnWithoutName()
         {
             var expr = new CreateColumnExpression()
             {
@@ -82,7 +82,7 @@ namespace FluentMigrator.Tests.Unit
         }
 
         [Test]
-        public void GetPrimaryKeyNamePrefixeForMultipleColumns()
+        public void GetPrimaryKeyNamePrefixForMultipleColumns()
         {
             var expr = new CreateTableExpression()
             {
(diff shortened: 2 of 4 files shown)


src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
```diff
@@ -29,23 +32,24 @@ public class DefaultPrimaryKeyNameConvention : IColumnsConvention
         /// <inheritdoc />
         public IColumnsExpression Apply(IColumnsExpression expression)
         {
+            if (expression.Columns.Count(x => x.IsPrimaryKey) > 1)
```
jzabroski, line 35: I don't understand why this is something that should be discouraged. It is perfectly valid for a table to have a primary key with multiple columns.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
```diff
@@ -29,23 +32,24 @@ public class DefaultPrimaryKeyNameConvention : IColumnsConvention
...
+                throw new InvalidOperationException("Error creating table with multiple primary keys.");
+
             foreach (var columnDefinition in expression.Columns)
             {
+                if (string.IsNullOrEmpty(columnDefinition.Name))
```
jzabroski, line 40: This should be handled through validation logic. There is a separate layer in the framework for validation of expressions.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
```diff
@@ -29,23 +32,24 @@ public class DefaultPrimaryKeyNameConvention : IColumnsConvention
...
+                    throw new ArgumentException("An object or column name is missing or empty.");
+
                 if (columnDefinition.IsPrimaryKey && string.IsNullOrEmpty(columnDefinition.PrimaryKeyName))
                 {
                     var tableName = string.IsNullOrEmpty(columnDefinition.TableName)
                         ? expression.TableName
                         : columnDefinition.TableName;
-                    columnDefinition.PrimaryKeyName = GetPrimaryKeyName(tableName);
+                    columnDefinition.PrimaryKeyName = $"PK_{tableName}_{columnDefinition.Name}";
```
jzabroski, line 48: I think the current behavior is what most users prefer. The actual column name(s) don't really matter to most developers except when needing to change the data type, as that requires dropping and recreating the column name.

Additionally, most DB engines like SQL Server have limit of, say, 255 characters for system object names.

Awhile ago I had the idea of an Opinions assembly that has a different set of default conventions and even some syntax shortcuts. This could live in something like that.

jzabroski (commented):
In general, I don't think I will accept this PR but I left what I hope are helpful comments as to why.

I will look at the original issue this PR was conceived from and see if there is a simpler approach.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
TheSquidCombatant: One primary key on many columns is good. But here we are talking specifically about multiple primary keys, each with its own name, am I wrong?

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
TheSquidCombatant: This part of the code is about use column names. Isn't it good to recognize the error as early as possible? To which specific class do you propose to move this check?

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
TheSquidCombatant: @jzabroski 

If the column name doesn't matter then why shouldn't it be deterministic?

If the name is too long the user will just get an exception, as in the case of `DefaultConstraintNameConvention`.

I don't think this is a good enough reason to complicate the project's ecosystem.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
jzabroski: I am pretty sure you are wrong. But your questioning my statement honestly gives me pause. But I'm 99% sure if I integration test this on a compound key, it will fail.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
jzabroski: Why separate the two? Same reason you would separate input validation from a Controller Action method in ASP.NET MVC - it keeps the core business logic focused on the actual problem domain and allows for more precise errors / warnings if done as a ValidationResult rather than throwing an exception. If you throw, you lose Metadata that could be used to build even more nicer error messages, too - similar to what Patrik Svensson's  SpectreConsole does. In general, throwing exceptions for input validation also prevents "railway oriented programming" where the client gets back a object representing success or failure.

That being said, our Validation layer isn't the best. I realize that's frustrating feedback considering I just told you to use it, but it was pretty good 7 years ago - but third party libraries like FluentValidation have made insane leaps forward since then. We currently pattern the existing one of off         System.ComponentModel.DataAnnotations.Validator. I need to delete the areas of the code that still reference CollectErrors, as they have been marked obsolete for awhile. 

But:

1.The interface for adding validation is here but doesn't let you explicitly handle each expression root type - you would have to do type tests and dispatch off a Type-switch (probably to something like FluentValidation): https://github.com/fluentmigrator/fluentmigrator/blob/main/src%2FFluentMigrator.Abstractions%2FValidation%2FIMigrationExpressionValidator.cs
2. The default implementation should be replaced here: https://github.com/fluentmigrator/fluentmigrator/blob/main/src%2FFluentMigrator.Abstractions%2FValidation%2FDefaultMigrationExpressionValidator.cs

3. The CLI tool supports a Validate command. https://github.com/fluentmigrator/fluentmigrator/blob/main/src%2FFluentMigrator.DotNet.Cli%2FCommands%2FValidate.cs
4. Each Model XxxDefinition class and XxxExpression class has a Validate member.

src/FluentMigrator.Runner.Core/Conventions/DefaultPrimaryKeyNameConvention.cs
jzabroski: From reading the original issue, it sounds like Create.PrimaryKey() uses DefaultConstraintNameConvention - this is a bug. It should use DefaultPrimaryKeyNameConvention.

Default constraints should just be for default values specified by the ANSI SQL DDL `DEFAULT` keyword.

PR #1814, opened 2024-05-30
"NULLS NOT DISTINCT" for indexes in PostgreSQL 15+

Domanser:
This pull request adds support of "NULLS NOT DISTINCT" option for PostgreSQL.

By default, unique indexes in PostgreSQL allow multiple null values (so, null values are considered not equal to each other). In PostgreSQL 15, new option was added: "NULLS NOT DISTINCT", which allows only single null value.
https://www.postgresql.org/docs/current/sql-createindex.html

Before PostgreSQL 15, there was a workaround with two indexes: https://stackoverflow.com/questions/8289100/create-unique-constraint-with-null-columns

To enable "NULLS NOT DISTINCT" for index, UniqueNullsNotDistinct method should be called:
```
Create.Index("MyIndex").OnTable("MyTable1").InSchema("public")
   .OnColumn("Column1").Ascending()
   .OnColumn("Column2").Ascending()
   .WithOptions().UniqueNullsNotDistinct();
```
Runner should be configured with AddPostgres15_0:
```
services
   .AddFluentMigratorCore()
   .ConfigureRunner(builder => builder.AddPostgres15_0());
```

### DIFF
--- a/src/FluentMigrator.DotNet.Cli/Setup.cs
+++ b/src/FluentMigrator.DotNet.Cli/Setup.cs
@@ -92,6 +92,7 @@ namespace FluentMigrator.DotNet.Cli
                         .AddPostgres92()
                         .AddPostgres10_0()
                         .AddPostgres11_0()
+                        .AddPostgres15_0()
                         .AddRedshift()
                         .AddSnowflake()
                         .AddSQLite()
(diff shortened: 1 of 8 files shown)


jzabroski:
@lillo42 would you like to review this feature?

lillo42 (approved):

lillo42:
@jzabroski LGTM

PR #1829, opened 2024-06-19
add custom version info table for tool

2m0nd:
We are using custom table info metadata for our projects. Dotnet tool for fluent migrator don't allow configure version info metadata, i added this feature in my merge request. To avoid adding many parameters i added one json paramter, where we can rewrite all parameters. `-v '{"ColumnName":"version","SchemaName":"public","TableName":"version_info","AppliedOnColumnName":"applied_on","DescriptionColumnName":"description"}'`


jzabroski:
@2m0nd Thanks for submitting this PR. The way it should work is that, through DI, it should first try to resolve the default version info metadata provided by the dotnet-fm tool (or any other FM endpoint like MSBuild), and then try to replace the default with a custom version found in the provided assembly. In this way, there would be no need for any of this code, although updating https://github.com/fluentmigrator/documentation to make this approach clear would be ideal.

PR #1833, opened 2024-07-15
Implement sequence existence check through the Schema

dev-anton-ko:
Implement sequence existence check through the Schema


src/FluentMigrator/Builders/Schema/Sequence/SchemaSyntaxQuery.cs
```diff
@@ -0,0 +1,49 @@
+﻿#region License
```
jzabroski, line 1: Isnt this file supposed to be named SchemaSequenceQuery?

jzabroski (commented):
@dev-anton-ko Can you explain what database you tested this change against? What behavior would you expect for MySQL, where the schema is the current database name? Why do you want this feature? What problem does it solve? Can you show me a sample migration where you would use this?

src/FluentMigrator/Builders/Schema/Sequence/SchemaSyntaxQuery.cs
dev-anton-ko: Yes. I am renaming it.

dev-anton-ko:
> @dev-anton-ko Can you explain what database you tested this change against? What behavior would you expect for MySQL, where the schema is the current database name? Why do you want this feature? What problem does it solve? Can you show me a sample migration where you would use this?

I have tested the changes against MSSQL.
Now I can write following when I am working with a table:

```
if (Schema.Schema("test_schema").Table("test_table").Exists())
{
    Create.Table("test_table").InSchema("test_schema")
    .WithColumn("Id").AsInt32().PrimaryKey().Identity()
    .WithColumn("Name").AsString(255).NotNullable()
    .WithColumn("Description").AsString(255).Nullable();
 }
```
But I cannot write the similar condition for the sequence:

```
if (Schema.Schema("test_schema")._Sequence_("test_sequence")._Exists_())
{
    Create.Sequence("test_table").InSchema("test_schema")
	    .StartWith(1)
	    .IncrementBy(1)
	    .MaxValue(int.MaxValue)
	    .MinValue(int.MinValue)
	    .Cache(10);
}	    
```
Generally, I needed to verify the existence of a sequence.
As far as I know, it is not actual for MySql - it has no sequences.

jzabroski:
> As far as I know, it is not actual for MySql - it has no sequences.

@dev-anton-ko Yikes, good point, you're right. Eek at this:

https://github.com/fluentmigrator/fluentmigrator/blob/f3fc4879b09fac687171057896160ba109659bec/src/FluentMigrator.Runner.MySql/Processors/MySql/MySqlProcessor.cs#L61

It always returns true.  It doesn't even handle CompatibilityMode.  In fact, since it lives on the IMigrationProcessor, it does not even know about CompatibilityMode, which only exists as a member of GeneratorOptions 

It would be great if there was a way to handle compatibility mode, but that is not a fair way to judge this PR.

PR #1842, opened 2024-07-20
#1836: Support custom IVersionTableMetaData DI registration before AddFluentMigratorCore call

### LINKED ISSUE #1836: Registering custom IVersionTableMetaData must be done after AddFluentMigratorCore
**Describe the bug**
I had to use a custom version info table, so I registered it like this:
```csharp
services.AddScoped<IVersionTableMetaData, CustomVersionTableMetaData>();
```
Previously it worked, but after I refactored my code (just moving registrations around) it broke.

The reason:
Previously, I _first_ called `AddFluentMigratorCore(...)` and _after_ that `services.AddScoped<IVersionTableMetaData, CustomVersionTableMetaData>()`. While refactoring the order changed.

**To Reproduce**
This version works (pseudo code):
```csharp
services.AddScoped<IVersionTableMetaData, CustomVersionTableMetaData>();
services.AddFluentMigratorCore(..);
```
this doesn't:
```csharp
services.AddFluentMigratorCore(..);
services.AddScoped<IVersionTableMetaData, CustomVersionTableMetaData>();
```

**Expected behavior**
The order of the registration should not cause the custom migration to be ignored/missed.
I consider this a very suptle but hard to spot and therefore serious bug, despite the workarounds being pretty straight forward.

**Investigation**
Usually, in a library this is caused by using `services.AddScoped<...>()` instead of `services.TryAddScoped<...>()` which overrides user registrations, but I couldn't find something like that.
As I did not register a `AssemblyVersionTableMetaDataSourceItem` with `ScanIn(...).For.VersionTableMetaData()`, this returns `null` for `VersionTableMetaData` property (debugged this).

**Workarounds**
- Call `AddScoped<IVersionTableMetaData, CustomVersionTableMetaData>();` after `AddFluentMigratorCore(...)`
- Use `ScanIn(...).For.VersionTableMetaData()`
- Use `WithVersionTable(new CustomVersionTableMetaData())`

 - OS: Windows 10
 - Platform .NET 8
 - FluentMigrator version 5.2.0
 - FluentMigrator runner "in-process runner"
 - Database Management System Oracle
 - Database Management System Version "Oracle 19c

warappa:
With this PR, FluentMigrator only registers a default `IVersionTableMetaData` implementation if no user-provided implementation was registered prior.

This still does support registering a custom `IVersionTableMetaData` implementation after the `AddFluentMigratorCore()` call, as later registered services are prioritized by MS dependency container.

This PR contains relevant tests.

Fixes #1836


warappa:
@jzabroski 
I added the TryAdd registrations as diskussed.

A few things about the changed registrations
- I turned _all_ library service registrations into `TryAdd` variants
  - Even `TryAddSingleton`
  - All except those which are resolved as `IEnumerable<T>` as this implys that having a list of services is supported by FluentMigrator (eg. `IAssemblySource`)
  - Not only interface types but also concrete types (eg. `AssemblyLoaderFactory`)
 - `CreateServices` is internal, obsolete and as far as I see only used in test code paths, so I left it as-is
- `ConfigureRunner` was untouched as it intentionally registers the service
- I don't think other registrations need the `TryAdd` change, as they are "on-purpose" registrations and not "general" registrations as in `AddFluentMigratorCore`.

Some questions:
- Should non-interface types be registered with `TryAdd` (as I already did)?
- I kept the new-line formatting but on some lines it's kinda weird to have `services.` on a separate line (and sometimes a comment in-between) - should the `services.` be placed on the same line as the next `TryAdd`?

5 tests are failing, but these are already failing before my changes so this should be fine.

jzabroski:
which 5 tests are failing? none failing locally for me or azure dev ops pipeline? can you open an issue with the failing stack trace/nunit error

warappa:
Just a quick update
Those are culture dependent errors. "," instead of "." for numbers, error messages in my locale (`de`) instead of the english ones.

I think the FluentMigrator tests should set the culture to en-US before executing any tests to be stable for all contributors.

jzabroski:
Thanks.  I am about to head on paternity leave so I have been heads down writing documentation on various responsibilities I have at work.  I imagine I will merge this after looking through it more carefully.

jzabroski:
I've finally fixed our ci/cd sfailing. (turns out, i had to add new pipelines and delete the old ones after renaming the master branch to main). This was slowing me down with evaluating PRs as i couldn't even immediately see build pass/fail.

warappa:
No stress. This is an important code path, so take your time.

On another note:
I believe build pipelines were designed to frustrate developers 😩

jzabroski:
@warappa Sorry, I completely lost track of this PR and my eyes got top-heavy reading only the more recent PRs.

warappa:
No problem :) Just ping me if I can help.

jzabroski:
It's been mered. I think yo mentioned wanting to do a PR for ValidateOnBuild?

warappa:
@jzabroski Yes, I started with it but it's not straight forward. But I will try to make it into a PR.

warappa:
I'll take a look.

jzabroski:
The typo is here: https://github.com/fluentmigrator/fluentmigrator/blob/05382383cecfd859a05df16fbf0909452d68225b/src/FluentMigrator.Runner.Core/Generators/SelectingGeneratorAccessor.cs#L55


I will fix it

warappa:
I fixed the tests by configuring a runner. I took SQLite, but that's just a random choice, because it's only for registering any migration generator.

I realized that I just test it with 1 service (`IVersionTableMetaData`) - I hope this is enough for you.

jzabroski:
yes, progress is incemental.

PR #1847, opened 2024-07-30
Use database function to set the value of `AppliedOn`

### LINKED ISSUE #1846: Do not hardcode date when updating `VersionInfo` table
**Is your feature request related to a problem? Please describe.**
The `VersionInfo` table has the column `AppliedOn` to note when a migration was _applied_ on the database.

The `AppliedOn` column value is updated with the expected date when updating a database through a direct connection but it uses a hardcoded date when generating SQL scripts.

**Describe the solution you'd like**
SQL scripts use a date calculated when the script is executed rather than the date when the script was generated.

**Additional context**
The date is hardcoded in the implementations of `IVersionLoader` (note the value that is used for the `AppliedOn` column):

https://github.com/fluentmigrator/fluentmigrator/blob/214345132f8b599ce72fdd21bf966baf894aae2d/src/FluentMigrator.Runner.Core/VersionLoader.cs#L140-L148

https://github.com/fluentmigrator/fluentmigrator/blob/214345132f8b599ce72fdd21bf966baf894aae2d/src/FluentMigrator.Runner.Core/ConnectionlessVersionLoader.cs#L197-L205

**Is this a feature you'd like to submit a PR for?**
Yes.

nathanpovo:
Use a database function to set the value of the `AppliedOn` column of the `VersionInfo` table whenever the version info is updated.

Previously, the value of `AppliedOn` would be decided when the migrations are generated and then hardcoded into the SQL script.

This is not an issue when the database is updated directly but it is a problem when updating the database through SQL scripts.

---

### Before

```sql
INSERT INTO [dbo].[VersionInfo] ([Version], [AppliedOn], [Description]) VALUES (20240101010101, '2024-07-30T09:27:42', N'MyMigrationName');
```

### After

```sql
INSERT INTO [dbo].[VersionInfo] ([Version], [AppliedOn], [Description]) VALUES (20240101010101, GETUTCDATE(), N'MyMigrationName');
```

---

Fixes #1846

### DIFF
--- a/src/FluentMigrator.Runner.Core/ConnectionlessVersionLoader.cs
+++ b/src/FluentMigrator.Runner.Core/ConnectionlessVersionLoader.cs
@@ -200,10 +200,26 @@ namespace FluentMigrator.Runner
 
         protected virtual InsertionDataDefinition CreateVersionInfoInsertionData(long version, string description)
         {
+            object appliedOnValue;
+
+            if (_quoter is null)
+            {
+                appliedOnValue = DateTime.UtcNow;
+            }
+            else
+            {
+                var quotedCurrentDate = _quoter.QuoteValue(SystemMethods.CurrentUTCDateTime);
+
+                // Default to using DateTime if no system method could be obtained
+                appliedOnValue = string.IsNullOrWhiteSpace(quotedCurrentDate)
+                    ? (object) DateTime.UtcNow
+                    : RawSql.Insert(quotedCurrentDate);
+            }
+
             return new InsertionDataDefinition
             {
                 new KeyValuePair<string, object>(VersionTableMetaData.ColumnName, version),
-                new KeyValuePair<string, object>(VersionTableMetaData.AppliedOnColumnName, DateTime.UtcNow),
+                new KeyValuePair<string, object>(VersionTableMetaData.AppliedOnColumnName, appliedOnValue),
                 new KeyValuePair<string, object>(VersionTableMetaData.DescriptionColumnName, description)
             };
         }
--- a/src/FluentMigrator.Runner.Core/ConnectionlessVersionLoader.cs
+++ b/src/FluentMigrator.Runner.Core/ConnectionlessVersionLoader.cs
@@ -42,6 +42,8 @@ namespace FluentMigrator.Runner
         [NotNull]
         private readonly IMigrationInformationLoader _migrationInformationLoader;
 
+        private readonly IQuoter _quoter;
+
         private bool _versionsLoaded;
 
         [Obsolete]
@@ -55,6 +57,7 @@ namespace FluentMigrator.Runner
         {
             _migrationInformationLoader = runner.MigrationLoader;
             _processor = runner.Processor;
+            _quoter = _processor.GetQuoter();
 
             Runner = runner;
             Assemblies = assemblies;
@@ -88,6 +91,7 @@ namespace FluentMigrator.Runner
         {
             _processor = processorAccessor.Processor;
             _migrationInformationLoader = migrationInformationLoader;
+            _quoter = _processor.GetQuoter();
             Conventions = conventions;
             StartVersion = runnerOptions.Value.StartVersion;
             TargetVersion = runnerOptions.Value.Version;
--- a/src/FluentMigrator.Runner.Core/Processors/ProcessorExtensions.cs
+++ b/src/FluentMigrator.Runner.Core/Processors/ProcessorExtensions.cs
@@ -1,0 +1,55 @@
+#region License
+// Copyright (c) 2024, Fluent Migrator Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using FluentMigrator.Runner.Generators;
+using FluentMigrator.Runner.Generators.Generic;
+
+namespace FluentMigrator.Runner.Processors
+{
+    public static class ProcessorExtensions
+    {
+        /// <summary>
+        /// A helper method to get the <see cref="IQuoter"/> for the given migration processor since the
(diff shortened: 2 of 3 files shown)


jzabroski:
The extension method to GetQuoter would not work for all generators. I don't want to support that. Why do you need it on the first place?

jzabroski:
In general, I agree we need to support this but wanted to suggest some changes. 

1. This is a breaking change. Should this be hidden behind a feature flag? As it stands, the quoter will never be null since it's injected now. So, using the presence of a quoter as a feature flag is not a robust approach.

2. Remove the extension method GetQuoter. You don't need it. Use dependency injection and injection the Quoter in the constructor. (Getting the quoter through the generator is a violation of the Law of Demeter, and a sign the code will be quite brittle to extension.)

3. Not every provider supports utc time. It's not clear to me that this will work across every single database provider. But as it's connectionless, it should be easy to at least run and observe what it generates for each ProviderId.

nathanpovo:
Thanks for taking the time to go through this.

Yesterday I wrote a lengthy comment responding to your feedback and it seems to have disappeared.

---

> The extension method to GetQuoter would not work for all generators. I don't want to support that. Why do you need it on the first place?

> Remove the extension method GetQuoter. You don't need it. Use dependency injection and injection the Quoter in the constructor.

The extension method is necessary because of 2 reasons:

1. Each database provider implements its own implementation of `IQuoter` which means that there might be multiple `IQuoter`s in the dependency injection container.
2. Implementations of `IQuoter` are registered as their concrete type in the dependency injection container.

You can see that this is the case by looking at the code of `AddFluentMigratorCore`

https://github.com/fluentmigrator/fluentmigrator/blob/4026fc13ec4a1bfb99cc58bd8b39d8f11c310a5d/src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs#L52-L180

and the builders of the database providers:

https://github.com/fluentmigrator/fluentmigrator/blob/4026fc13ec4a1bfb99cc58bd8b39d8f11c310a5d/src/FluentMigrator.Runner.SqlServer/SqlServerRunnerBuilderExtensions.cs#L40-L52

https://github.com/fluentmigrator/fluentmigrator/blob/4026fc13ec4a1bfb99cc58bd8b39d8f11c310a5d/src/FluentMigrator.Runner.Postgres/PostgresRunnerBuilderExtensions.cs#L118-L130

https://github.com/fluentmigrator/fluentmigrator/blob/4026fc13ec4a1bfb99cc58bd8b39d8f11c310a5d/src/FluentMigrator.Runner.Oracle/OracleRunnerBuilderExtensions.cs#L37-L47

This means that there are no registrations of `IQuoter` in the dependency injection container. And, even if there were, there would be multiple registrations of `IQuoter` in the dependency injection container.

This can be tested out quickly by running the example console application:

image

image

What I originally wanted to do, was add a property to `IMigrationProcessor` that would return the `IQuoter` implementation that the `IMigrationProcessor` uses. But I ran into issues here because the project that contains `IMigrationProcessor` has no reference to the project that contains `IQuoter`.

I did not wish to add any new references between projects, so I settled on the extension method.

---

> This is a breaking change. Should this be hidden behind a feature flag?

I see this as a bug fix since the scripts were not being generated correctly.

However, I understand that users might have relied on the previous behaviour.

Could this new behaviour be based on a configuration option in `RunnerOptions`? It could default to `false` to maintain the previous behaviour. In a future release, the default could be changed to `true`.

---

> Not every provider supports utc time. It's not clear to me that this will work across every single database provider.

Based on the database providers in the code, all `IQuoter` implementations that override `FormatSystemMethods` support `SystemMethods.CurrentUTCDateTime`

https://github.com/fluentmigrator/fluentmigrator/blob/4026fc13ec4a1bfb99cc58bd8b39d8f11c310a5d/src/FluentMigrator.Runner.Core/Generators/Generic/GenericQuoter.cs#L114-L117

jzabroski:
That's a really good reason why you did it this way. I am trying to think of a way we can just inject it. I made each generator take an explicit dependency on purpose for this and type maps.

Injecting it likely has other unforeseen  benefits.

What we really (probably) want is a tiny service locator that can get all generator dependencies as a GeneratorContext. (I'm almost done with a tiny PR to add an ISeparator interface to formalize the concepts of statements and batches, which will further help offline scripting)

I need to think about this more. If I can't arrive at a better solution, I guess we will use your approach.

jzabroski:
I'm going to try to merge this tonight. Just released 6.0, will put this in 6.1. Thanks for your patience. Just had a newborn baby boy in August and just starting to get good nights sleep for the first time in about two months.

nathanpovo:
Congrats on the baby!

Absolutely not an issue. Take all the time that you need.

I worked around the issue in my own project so there is no rush on my side. The change request was a means for me to contribute something to the library that others might be able to make use of.

jzabroski:
@nathanpovo As an update and late change request, I believe there is already infrastructure to get the generator easily - see IGeneratorAccessor. Can you use that?

nathanpovo:
> I believe there is already infrastructure to get the generator easily - see IGeneratorAccessor. Can you use that?

I think this can be used.

I checked what code would need to be modified for this to be done, and the constructors of the `ConnectionlessVersionLoader` and `VersionLoader` classes will need to be changed with an additional parameter for `IGeneratorAccessor` so that it is injected.

Is it fine to modify the constructors? I have done the modification, we can reset it back if the constructors should not be modified.

jzabroski:
I think this change makes sense, even if by some opinion its technical debt. We can always obsolete those constructors later when a better design comes along.

jzabroski:
@nathanpovo It looks like some tests failed - can you take a look?

nathanpovo:
Sorry, didn't notice the failing tests. They have been fixed.

jzabroski:
Looks great. I will do one last read through the code this weekend and then merge it for 6.3.0 assuming I dont catch anything, which I doubt I will.

jzabroski:
@nathanpovo One last thought - should IMigrationGenerator expose the IQuoter directly?  I am trying to think of a reason not to and can't think of one.  I'd prefer that to the cast to GeneratorBase, as someone _could_ in theory write their own Generator without that.

nathanpovo:
That's kind of what I had wanted to do originally (but have `IMigrationProcessor` expose `IQuoter`).

Unfortunately, it's not possible because `IMigrationGenerator` currently does not have access to `IQuoter`. We'd need to add a reference to the `FluentMigrator.Runner.Core` project in the `FluentMigrator.Abstractions` project.

image

Or move `IQuoter` to the `FluentMigrator.Abstractions` project.

Both of which are breaking changes.

jzabroski:
Right, sorry, new parent brain, forgot we discussed this.  I'll merge this and add a ticket to clean this up.

PR #1849, opened 2024-07-30
Bugfixes/1848 issues with non english culture

### LINKED ISSUE #1848: Some tests fail if culture is different from en-US
**Describe the bug**
Running the tests in german culture yields 5 errors. These are all releated to differences in culture ("," instead of ".", localized error messages instead of the english ones).

**To Reproduce**
Run the tests with a German culture
- A) German WIndows
- B) Setting `System.Globalization.CultureInfo.CurrentCulture`/`System.Globalization.CultureInfo.CurrentUICulture` to German culture before any tests executes (untested)

**Expected behavior**
Tests should not vary because of culture of operating system.

 - OS: Windows 11
 - Platform .NET 8
 - FluentMigrator version 5.2.0/source

warappa:
- Ensure tests checking against validation messages have invariant (english) culture
- Fix bug in `PostgresGenerator` if `IndexVacuumCleanupIndexScaleFactor` is used with a floating point number (should use invariant culture).

Fixes #1848 

> [!NOTE]
> Maybe other `ToString()` calls need to be investigated to ensure they do not cause the same issues as in `PostgresGenerator`.

### DIFF
--- a/test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
@@ -109,7 +109,7 @@ namespace FluentMigrator.Tests.Unit.Builders.Delete
         }
 
         [Test]
-        [SetUICulture("en-US")] // Ensure validation messages are in English
+        [SetUICulture("")] // Ensure validation messages are in English
         public void DefaultMigrationExpressionValidatorShouldReturnErrorWhenTableNameIsNotSpecified()
         {
             var mockServiceProvider = new Mock<IServiceProvider>();
--- a/test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
@@ -109,11 +109,9 @@ namespace FluentMigrator.Tests.Unit.Builders.Delete
         }
 
         [Test]
+        [SetUICulture("en-US")] // Ensure validation messages are in English
         public void DefaultMigrationExpressionValidatorShouldReturnErrorWhenTableNameIsNotSpecified()
         {
-            // Ensure error messages are in English
-            CultureInfo.CurrentUICulture = CultureInfo.InvariantCulture;
-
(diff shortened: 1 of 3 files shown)


test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
```diff
@@ -110,10 +111,13 @@ public void CallingIsNullAddsANullColumn()
         [Test]
         public void DefaultMigrationExpressionValidatorShouldReturnErrorWhenTableNameIsNotSpecified()
         {
+            // Ensure error messages are in English
+            CultureInfo.CurrentUICulture = CultureInfo.InvariantCulture;
```
jzabroski, line 115: Why not just use [Test, Culture("en-US")]

jzabroski (changes requested):
Are you familiar with https://docs.nunit.org/articles/nunit/writing-tests/attributes/setculture.html ?  I have not used it but have used the xunit equivalent UseCulture.  In reading the docs, it seems extremely similar.  I am not sure if it will work side-by-side with Setup, but otherwise it should work great?  Note, nunit confusingly has a CultureAttribute, which is a _test filter_.  We want SetCultureAttribute, not CultureAttribute.  SetCultureAttribute sets the culture.  CultureAttribute filters it.

test/FluentMigrator.Tests/Unit/Builders/Delete/DeleteDataExpressionTests.cs
warappa: I didn't know about this but this seems very handy! Thanks for pointing that out!👍

warappa: I use `SetUICultureAttribute`, as the validation messages rely on `CurrentUICulture` and not on `CurrentCulture` (tested).

jzabroski: thanks. will kick off ci service in a bit

warappa: Updated to empty string

PR #1896, opened 2024-10-03
Computed column support

### LINKED ISSUE #1152: Problem with create table with custom type column

Shane32:
Closes #1152 

Adds computed column support

Todo:
- [ ] Verify and/or test syntax for each provider
- [ ] Determine how postgres alters generated/virtual columns
- [x] Verify infrastructure changes are okay
- [x] Verify various database syntaxes are okay
- [x] Add tests

Notes:
- Throws exceptions as appropriate similar to other provider-specific features
- Adds parentheses around the provided expression

### DIFF
--- a/src/FluentMigrator.Abstractions/Builders/IColumnOptionSyntax.cs
+++ b/src/FluentMigrator.Abstractions/Builders/IColumnOptionSyntax.cs
@@ -112,6 +112,14 @@ namespace FluentMigrator.Builders
         TNext NotNullable();
 
         /// <summary>
+        /// Defines the column as a computed column
+        /// </summary>
+        /// <param name="expression">The expression to calculate</param>
+        /// <param name="stored">Whether the computed column is virtual or stored</param>
+        /// <returns>The next step</returns>
+        TNext Computed(string expression, bool stored = false);
+
+        /// <summary>
         /// Specify a unique index for the column
         /// </summary>
         /// <returns>The next step</returns>
--- a/src/FluentMigrator.Abstractions/Model/ColumnDataType.cs
+++ b/src/FluentMigrator.Abstractions/Model/ColumnDataType.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 // Copyright (c) 2023, Fluent Migrator Project
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -34,5 +34,11 @@ namespace FluentMigrator.Model
 
         /// <inheritdoc />
         public string CustomType { get; set; }
+
+        /// <inheritdoc />
+        public string Expression { get; set; }
+
+        /// <inheritdoc />
+        public bool ExpressionStored { get; set; }
     }
 }
--- a/src/FluentMigrator.Runner.Postgres/Generators/Postgres/PostgresColumn.cs
+++ b/src/FluentMigrator.Runner.Postgres/Generators/Postgres/PostgresColumn.cs
@@ -154,5 +154,11 @@ namespace FluentMigrator.Runner.Generators.Postgres
         {
             return FormatType(column);
         }
+
+        /// <inheritdoc />
+        protected override string FormatExpression(ColumnDefinition column)
+        {
+            return column.Expression == null ? null : $"GENERATED ALWAYS AS ({column.Expression}) STORED";
+        }
     }
 }
--- a/src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2005Column.cs
+++ b/src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2005Column.cs
@@ -30,7 +30,7 @@ namespace FluentMigrator.Runner.Generators.SqlServer
         /// <inheritdoc />
         protected override string FormatNullable(ColumnDefinition column)
         {
-            if (column.IsNullable == true && column.Type == null && !string.IsNullOrEmpty(column.CustomType))
+            if (column.IsNullable == true && column.Type == null && !string.IsNullOrEmpty(column.CustomType) && column.Expression == null)
             {
                 return "NULL";
             }
--- a/test/FluentMigrator.Tests/Integration/Migrations/ComputedColumnMigration.cs
+++ b/test/FluentMigrator.Tests/Integration/Migrations/ComputedColumnMigration.cs
@@ -1,0 +1,41 @@
+#region License
+//
+// Copyright (c) 2007-2024, Fluent Migrator Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#endregion
+
+namespace FluentMigrator.Tests.Integration.Migrations
+{
+    [Migration(8)]
+    public class ComputedColumnMigration : Migration
+    {
+        public override void Up()
+        {
+            Create.Table("Products")
+                .WithColumn("Id").AsInt32().PrimaryKey().Identity()
+                .WithColumn("Price").AsDecimal(10, 2).NotNullable()
+                .WithColumn("Quantity").AsInt32().NotNullable()
+                .WithColumn("Total").AsDecimal(10, 2).Computed("Price * Quantity");
+                
+            Alter.Table("Products")
+                .AddColumn("TotalWithTax").AsDecimal(10, 2).Computed("Total * 1.1", true);
+        }
+        
+        public override void Down()
+        {
+            Delete.Table("Products");
+        }
+    }
+}
--- a/test/FluentMigrator.Tests/Unit/Generators/Postgres/PostgresColumnTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Generators/Postgres/PostgresColumnTests.cs
@@ -266,5 +266,41 @@ namespace FluentMigrator.Tests.Unit.Generators.Postgres
             var result = Generator.Generate(expression);
             result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" RENAME COLUMN \"TestColumn1\" TO \"TestColumn2\";");
         }
+
+        [Test]
+        public override void CanCreateColumnWithComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanCreateColumnWithStoredComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithStoredComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddStoredComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithStoredComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
+        }
     }
 }
--- a/test/FluentMigrator.Tests/Unit/Generators/Postgres10_0/Postgres10_0ColumnTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Generators/Postgres10_0/Postgres10_0ColumnTests.cs
@@ -93,5 +93,41 @@ namespace FluentMigrator.Tests.Unit.Generators.Postgres10_0
             var result = Generator.Generate(expression);
             result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" integer NOT NULL GENERATED ALWAYS AS IDENTITY;");
         }
+
+        [Test]
+        public override void CanCreateColumnWithComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanCreateColumnWithStoredComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithStoredComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddStoredComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithStoredComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
+        }
     }
 }
--- a/test/FluentMigrator.Tests/Unit/Generators/Redshift/RedshiftColumnTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Generators/Redshift/RedshiftColumnTests.cs
@@ -216,5 +216,41 @@ namespace FluentMigrator.Tests.Unit.Generators.Redshift
             var result = _generator.Generate(expression);
             result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" RENAME COLUMN \"TestColumn1\" TO \"TestColumn2\";");
         }
+
+        [Test]
+        public override void CanCreateColumnWithComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithComputed();
+            
+            var result = _generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) NOT NULL;");
+        }
+
+        [Test]
+        public override void CanCreateColumnWithStoredComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetCreateColumnExpressionWithStoredComputed();
+            
+            var result = _generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = _generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
+        }
+
+        [Test]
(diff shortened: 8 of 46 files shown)


src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2005Column.cs
```diff
@@ -30,7 +30,7 @@ public SqlServer2005Column(ISqlServerTypeMap typeMap, IQuoter quoter)
         /// <inheritdoc />
         protected override string FormatNullable(ColumnDefinition column)
         {
-            if (column.IsNullable == true && column.Type == null && !string.IsNullOrEmpty(column.CustomType))
+            if (column.IsNullable == true && column.Type == null && !string.IsNullOrEmpty(column.CustomType) && column.Expression == null)
```
Shane32, line 33: I really don't understand the purpose of this override to begin with.  Only in the case of using `AsCustom` does it provide `NULL` for a nullable field.  Why for `AsCustom` but not for any other data types?  I can see that perhaps it was desirable to specify `NULL` or `NOT NULL` to override a non-default database option of `ANSI_NULL_DFLT_OFF`, but then why would it be only for `AsCustom` and no other data types?  If this condition were removed, the entire overridden method could be removed.

jzabroski:
> We could throw an exception or just assume the user intended to create a stored computed column.

Use CompatibilityMode.HandleCompatibility

https://github.com/fluentmigrator/fluentmigrator/blob/38530ea35e11c83e2fffdaae3c911eff9334e3c2/src/FluentMigrator.Runner.Core/Generators/CompatibilityModeExtension.cs#L26

src/FluentMigrator.Runner.SqlServer/Generators/SqlServer/SqlServer2005Column.cs
jzabroski: Most likely it is because for all the other types, there are C# structs that map to them and the default is to be non-null.  You have to wrap a struct as Nullable<T> where T is a value type.

I agree, it is weird but would be a breaking change.  If we break it, it would ideally be done in a separate PR and we rebase this PR - this way the PR auto-generates in the release notes as a breaking change and is a lot clearer to end-users.

Is SqlServer2008Column or later re-overriding this behavior though?

Shane32: > Is SqlServer2008Column or later re-overriding this behavior though?

No it isn't.

src/FluentMigrator.Abstractions/Builders/IColumnOptionSyntax.cs
```diff
@@ -111,6 +111,14 @@ public interface IColumnOptionSyntax<TNext,TNextFk> : IFluentSyntax
         /// <returns>The next step</returns>
         TNext NotNullable();
 
+        /// <summary>
+        /// Defines the column as a computed column
+        /// </summary>
+        /// <param name="expression">The expression to calculate</param>
+        /// <param name="stored">Whether the computed column is virtual or stored</param>
+        /// <returns>The next step</returns>
+        TNext AsExpression(string expression, bool stored = false);
```
PhenX, line 120: Hello, just my two cents, but I think this naming may confuse people, because of C# Expression trees for example, while it is dedicated to computed columns. Why not name it simply `AsComputed(string expression, bool stored = false)` ?

Shane32: Makes sense. Also it’s now treated as an option, so like Identity or Indexed, maybe it should be called Computed?

jzabroski: @PhenX Good question and your feedback serves as a "Keeper of the Flame."

I originally suggested AsExpression because the type is inferred by almost all database systems based on an expression's final type. In addition, I was thinking at some point specifying expressions from a POCO (similar to how third party package EasyMigrator works), so you *could* one day pass in an expression tree.

Incorporating expressions at some point in the future would also be a nice way to specify partition formulas. For a long time I didn't see a ton of value in the EasyMigrator approach but with features I want to do in the future like truly idempotent, offline Migrations and ability to test Migrations, having a current representation of what the schema should be described as entities is quite nice.

How confusing do you think it is? I guess the weird part is nobody in the sql world would call it an expression column.

Shane32: > the weird part is nobody in the sql world would call it an expression column

I think that’s important. I expect one of the goals is to make the library easy to use. If you were a user and were attempting to make a computed / calculated / virtual / generated column, what name would you look for?

Expression certainly still makes sense for the argument name even if it’s not the name of the method. 

> I originally suggested AsExpression because the type is inferred by almost all database systems based on an expression's final type

My research indicated that only SQL Server was able to infer the data type; all other databases require an explicit data type. Hence moving the method to an option.

jzabroski: Ok. I'm out voted 2 to 1. AsComputed

Shane32: lol I’m okay with whatever. Just my opinion. 

Do you think it should be AsComputed or just Computed?  We don’t have AsIdentity which is also a generated column…

PhenX: I find `Computed` may be a better choice especially if we may still need to specify the column type with the "as" methods

jzabroski: Good point. Computed() reads better.

We could also add an `IColumnTypeSyntax.AsComputed<TDbType>(string expression, bool stored = false)` that wraps the long-form expression, e.g.

```c#
.AsComputed<int>("USER_ID()", true)
```

would be short-form for

```c#
.AsInt32().Computed("USER_ID(), true)
```

The relative advantage of the short-form is API discoverability.

`ExpressionBuilderWithColumnTypesBase `would then be something like:

```c#
        /// <inheritdoc />
        public TNext AsComputed<TDbType>(string expression, bool stored = false)
        {
            Column.Type = GetDbType<TDbType>();
            Column.ExpressionValue = expression;
            Column.ExpressionStored = stored;
            return (TNext)(object)this;
        }

        private DbType GetDbType<TDbType>()
        {
             var typeCode = Type.GetTypeCode(typeof(TDbType);
             switch (typeCode)
             {
                  case TypeCode.String: return DbType.String;
                  case TypeCode.Int32: return DbType.Int32;
                  case TypeCode.Int64: return DbType.Int64;
                  case TypeCode.Decimal: return DbType.Decimal;
                  // etc
                  default:
                       throw new InvalidOperationException("TDbType type code {typeCode} does not map to a FluentMigrator DbType.");
             }
        }
```

jzabroski: Thinking about my prototype above a bit more after lunch break - for Decimal and String, this interface is non-ideal since we wouldnt know the length of the string or whether it's an ANSI or UTF string, fixed or variable, or in the case of decimal, the scale and precision.

Shane32: Okay, I'll work on the 'long-form' first (when I have time) and then we can always add the short-form as subsequent PR if desired.

PhenX:
Hey @Shane32, any news about this pull request ? :)

jzabroski:
@PhenX I'll take the blame for the momentum we lost on this PR.  October was a busy time for me with the new born and I was probably too fussy with my feedback here.

I actually think we're not far from :shipit: 

So, relatively good news, @Shane32  changes only had one failing test so far:

FluentMigrator.Tests.Unit.Generators.Postgres.PostgresTableTests.CanCreateTableWithMultiColumnPrimaryKeyWithDefaultSchema(STRICT)

FluentMigrator.Exceptions.DatabaseOperationNotSupportedException : Virtual computed columns are not supported

src/FluentMigrator.Abstractions/Model/ColumnDataType.cs
```diff
@@ -34,5 +34,11 @@ public class ColumnDataType : IColumnDataType
 
         /// <inheritdoc />
         public string CustomType { get; set; }
+
+        /// <inheritdoc />
+        public string Expression { get; set; }
+
+        /// <inheritdoc />
+        public bool ExpressionStored { get; set; }
```
jzabroski, line 42: @PhenX @Shane32 This is why the unit test is failing.  Should it default to true?

PhenX:
@jzabroski I think my proposal to rename the method should be applied before merging though. It will be much more clear imo

jzabroski:
Agree. AsExpression should be renamed to AsComputed

src/FluentMigrator.Runner.Postgres/Generators/Postgres/PostgresColumn.cs
```diff
@@ -36,6 +36,7 @@ public PostgresColumn([NotNull] PostgresQuoter quoter)
         {
             // Note: While Postgres 10.0 introduced the ability to use ICU collations rather than depending on host OS implementations,
             // the syntax for collation requires specifying a type.  Therefore, FormatAlterType handles collation as well.
+            // todo: handle virtual columns
```
Shane32, line 39: todo here

test/FluentMigrator.Tests/Unit/Generators/Postgres/PostgresColumnTests.cs
```diff
@@ -264,5 +264,41 @@ public override void CanRenameColumnWithDefaultSchema()
...
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
```
Shane32, line 292: todo

test/FluentMigrator.Tests/Unit/Generators/Postgres10_0/Postgres10_0ColumnTests.cs
```diff
@@ -93,5 +93,41 @@ public override void CanCreateColumnWithAutoIncrementAndDefaultSchema()
...
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = Generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
```
Shane32, line 121: todo

test/FluentMigrator.Tests/Unit/Generators/Redshift/RedshiftColumnTests.cs
```diff
@@ -214,5 +214,41 @@ public override void CanRenameColumnWithDefaultSchema()
...
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ADD \"TestColumn1\" GENERATED ALWAYS AS (Price * Quantity) STORED NOT NULL;");
+        }
+
+        [Test]
+        public override void CanAlterColumnToAddComputedExpression()
+        {
+            var expression = GeneratorTestHelper.GetAlterColumnExpressionWithComputed();
+            
+            var result = _generator.Generate(expression);
+            result.ShouldBe("ALTER TABLE \"public\".\"TestTable1\" ALTER \"TestColumn1\" TYPE ;");
```
Shane32, line 242: todo

src/FluentMigrator.Abstractions/Model/ColumnDataType.cs
Shane32: I designed as follows: `Expression` is a string containing the computed column SQL expression if it is a computed column; otherwise `null`.  `ExpressionStored` is always `false` unless (a) it is a computed column, and (b) the column is persisted/stored (vs a virtual column).

Shane32:
Anyone know where to find postgres SQL documentation for altering generated columns?  This [link here](https://www.postgresql.org/docs/current/ddl-generated-columns.html) only mentions adding generated columns.

Shane32:
Also, do CI checks automatically run here?  Do they test all providers?  I'm guessing that locally only SQLite runs by default.  I could probably configure MS SQL localdb but I'm not really planning to test any other providers locally.

Shane32:
> So, relatively good news, @Shane32 changes only had one failing test so far:
> 
> FluentMigrator.Tests.Unit.Generators.Postgres.PostgresTableTests.CanCreateTableWithMultiColumnPrimaryKeyWithDefaultSchema(STRICT)

I fixed the bug and checked all the similar conditions.  Also wrapped all the expressions with parentheses, which is required by SQLite at minimum, and probably other providers too.

test/FluentMigrator.Tests/Integration/Migrations/ComputedColumnMigration.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override void Up()
+        {
+            Create.Table("Products")
+                .WithColumn("Id").AsInt32().PrimaryKey().Identity()
+                .WithColumn("Price").AsDecimal(10, 2).NotNullable()
+                .WithColumn("Quantity").AsInt32().NotNullable()
+                .WithColumn("Total").AsDecimal(10, 2).AsExpression("Price * Quantity");
+                
+            Alter.Table("Products")
+                .AddColumn("TotalWithTax").AsDecimal(10, 2).AsExpression("Total * 1.1", true);
```
Shane32, line 33: This code will probably need a conditional to only run for providers that support virtual/generated columns.  Since it only tests SQLite currently, the test passes.  Also, one provider only supports stored/persisted columns, so might need to pass in `, true` for that provider.  Could expand this migration to test altering the column as well, with additional conditionals for providers that support altering virtual/generated columns.

Shane32:
@jzabroski Had any time to look at this PR?

jzabroski:
> Also, do CI checks automatically run here?

Unfortunately it's broken. I think its because I enabled branch protection. I really dislike how GitHub security works.

PhenX:
@jzabroski I hope you get better now, what do we do about the last remaining issues ? Do we merge this PR as is and I make a pull request right after with the resolved TODOs ?

jzabroski:
These two TODO?

- [ ] Verify and/or test syntax for each provider
- [ ] Determine how postgres alters generated/virtual columns

PhenX:
Sorry, I meant the renaming of "AsExpression" to "Computed", but the todos are of course needed

PhenX:
I prepared the changes here, along with test fixes https://github.com/PhenX/fluentmigrator/tree/computed_columns

jzabroski:
Thanks. 

1. Should it be `.Computed()` or `.AsComputed()`?
2. Regarding how Postgres handles generated columns, Grok says the following scenarios - **and** we can compress this response into another prompt that would guide Copilot with Claude Sonnet 4 to generate tests for Postgres iff we explain the unit testing architecture for testing generators:
   > You cannot directly modify the expression of an existing generated column in PostgreSQL (as of PostgreSQL 17). To change the expression, you must:
   > 
   > #### Drop the column and recreate it with the new expression.
   > Steps:
   > 1. Use `ALTER TABLE table_name DROP COLUMN column_name;` to remove the existing generated column.
   > 2. Use `ALTER TABLE table_name ADD COLUMN column_name data_type GENERATED ALWAYS AS (new_expression) STORED;` (or `VIRTUAL` in PostgreSQL 17) to recreate it with the new expression.
   >
   > Example:
   > ```sql
   > -- Original table with a generated column
   > CREATE TABLE products (
   >    id SERIAL PRIMARY KEY,
   >    price NUMERIC,
   >    tax NUMERIC GENERATED ALWAYS AS (price * 0.1) STORED
   > );
   > 
   > -- Drop the generated column
   > ALTER TABLE products DROP COLUMN tax;
   > 
   > -- Recreate with a new expression
   > ALTER TABLE products ADD COLUMN tax NUMERIC GENERATED ALWAYS AS (price * 0.15) STORED;
   > ```
   > 
   > #### Changing the Data Type of a Generated Column
   > You can alter the data type of a generated column using ALTER TABLE ... ALTER COLUMN ... TYPE, but the new data type must be compatible with the expression’s output.
   > 
   > Syntax:
   > `ALTER TABLE table_name ALTER COLUMN column_name TYPE new_data_type;`
   > 
   > Example:
   > ```sql
   > -- Original table
   > CREATE TABLE products (
   >     id SERIAL PRIMARY KEY,
   >     price NUMERIC,
   >     tax INTEGER GENERATED ALWAYS AS (price * 0.1) STORED
   > );
   > 
   > -- Change data type to NUMERIC
   > ALTER TABLE products ALTER COLUMN tax TYPE NUMERIC;
   > ```
   >
   > #### Converting Between Stored and Virtual (PostgreSQL 17+)
   > In PostgreSQL 17, with the introduction of virtual generated columns, you cannot directly convert a STORED generated column to a VIRTUAL one (or vice versa) without dropping and recreating the column.
   > 
   > Steps:
   > 
   > 1. Drop the existing generated column.
   > 2. Recreate it with the desired storage type (STORED or VIRTUAL).
   >
   > Example:
   > ```sql
   > -- Original table with a STORED generated column
   > CREATE TABLE products (
   >   id SERIAL PRIMARY KEY,
   >   price NUMERIC,
   >   tax NUMERIC GENERATED ALWAYS AS (price * 0.1) STORED
   > );
   > 
   > -- Drop the column
   > ALTER TABLE products DROP COLUMN tax;
   > 
   > -- Recreate as VIRTUAL
   > ALTER TABLE products ADD COLUMN tax NUMERIC GENERATED ALWAYS AS (price * 0.1) VIRTUAL;
   > ```

PhenX:
1. Computed, like Indexed, etc
2. OK, this feature will clearly need integration tests before beings released, but I think we will need another pull request.

Shane32:
I just saw some activity here, so I merged `main` in, fixed merge errors, then renamed `AsExpression` to `Computed`.  All tests, such as they are, passed.

PhenX (approved):

test/FluentMigrator.Tests/Unit/Generators/Postgres/PostgresColumnTests.cs
PhenX: @Shane32 @jzabroski While writing integration tests, I'm telling myself I'm not sure we need to support this, or only the possibility to change the expression of an already computed column. What do you think ?

jzabroski: I think line 292 is wrong. 
1. All computed columns in Postgres are STORED.
2. You cannot change the type of the computed column.
3. You cannot replace a non-computed column with a computed column expression.

We could offer CompatibilityMode.Strict as a way to throw an error if someone tries to do this. If in non-Strict mode, drop the existing column and re-create it.

PhenX: Yeah, that's why I asked this, the SQL is clearly wrong, hence @Shane32 's todo
I think we should simply remove theses tests, the SQL error a user might have if he tries that might be explicit enough not for us to throw a custom exception, no ?

jzabroski: Ideally every Generator follow the same Test naming conventions, so I am not in favor of removing tests. In the past this has caused us to miss regressions when making changes.

PhenX:
@jzabroski let's merge this and after I'll make a PR with this what do you think?
 https://github.com/PhenX/fluentmigrator/tree/computed_columns_integration_tests

jzabroski:
its not letting me merge it is the issue. I even tried in codespaces and git cli using the manual merge -no-ff

PhenX:
Hum, dumb question, but : did you approve it ?

jzabroski (approved):

jzabroski:
image


It is claiming it is too complex for the web editor. Same issue with the DbFactory change. Won't let me do --no-ff either, and that is a local operation that GitHub has no control over.

Shane32:
I can rebase it manually, and then it will let you 'rebase and merge'.  I normally use Squash and Merge myself, so I don't run into this issue.

Shane32:
Looking at the merge history, it looks like 'squash and merge' was used in the past, so not sure if it was an intentional change to use 'rebase and merge' now.

Shane32:
@PhenX most likely your branch is now going to have merge conflicts since I rebased this PR...  (ugh)  it would have been the same with squash and merge though probably ...

PhenX:
@Shane32 no problem :)

jzabroski:
@Shane32 Solved via #2131 - I still dont know what went wrong but I think it was a glitch in the GitHub UI. Today, the UI is different, but I am closing this PR. I merged @PhenX #2131 but preserved the commit history so that you get proper credit.

I sometimes squash - especially if it helps hide how dumb/rushed I can be :) Sometimes I only have 30 minutes to code on this project and I just commit stuff without running tests, etc. Not how I would code at my day job but somewhat low risk. Most of my deep thought on this project is reviewing PRs and issues vs. getting time to code. Hoping that will change going forward.

jzabroski:
This is what the Conflict looks like today - way more descriptive than before. 
image

Shane32:
Ok no problem. Interesting it didn't look that way on my pc. It said no conflicts... In any case glad it's merged in!

Shane32:
Note: This feature was merged in via:
- #2131

Issue #1827, opened 2024-06-18
RawSql.Insert cannot be used as entire Where clause or Set clause in UpdateDataExpression/DeleteDataExpression


jzabroski:
**Describe the bug**
RawSql.Insert cannot be used as entire Where clause or Set clause in UpdateDataExpression/DeleteDataExpression

**To Reproduce**
```c#
    [Migration(202406181031, "Testing RawSql.Insert behavior")]
    public class TestMigration : ForwardOnlyMigration
    {
        public override void Up()
        {
            Update.Table("Foo").InSchema("dbo")
                .Set(RawSql.Insert("Baz = CASE WHEN Bar = 1 THEN 2 ELSE 0 END"))
                .Where(RawSql.Insert("Baz IS NULL"));
        }
    }
```

generates the following invalid SQL expression:
```sql
UPDATE [dbo].[Foo] SET [Value] = N'Baz = CASE WHEN Bar = 1 THEN 2 ELSE 0 END' WHERE [Value] = N'Baz IS NULL'
```

**Expected behavior**
RawSql should be... raw sql :)

**Information (please complete the following information):**
 - FluentMigrator version 5.2.0

**Additional context**
Workaround is to use Execute.Sql syntax instead:

```c#
    [Migration(202406181031, "Testing RawSql.Insert behavior")]
    public class TestMigration : ForwardOnlyMigration
    {
        public override void Up()
        {
            Execute.Sql(@"UPDATE dbo.Foo
SET Baz = CASE WHEN Bar = 1 THEN 2 ELSE 0 END
WHERE Baz IS NULL");
        }
    }
```

PhenX:
Is it acceptable with such a syntax ?

```
Update.Table("Foo").InSchema("dbo")
    .Set(new
    {
        Baz = RawSql.Insert("CASE WHEN Bar = 1 THEN 2 ELSE 0 END")
    })
    .Where(new
    {
        Baz = RawSql.Insert("IS NULL")
    });
```

jzabroski:
I did not consider that. What does it generate?

PhenX:
Well for now it's a little bit better but not what we want because of the "=" on the WHERE part

```sql
UPDATE "dbo"."Foo" SET "Baz" = CASE WHEN Bar = 1 THEN 2 ELSE 0 END WHERE "Baz" = IS NULL
```

the fix is very simple though.

[...]

PR #1904, opened 2024-10-04
Bugfix/1827 raw sql

PhenX:
A proposal to fix the #1827 issue


jzabroski:
@PhenX The Postgres tests are failing due to the new code missing a trailing semi-colon at the end of the statement.  Any chance you can fix it?  I think we should include the semi-colon, to be clear, which I believe was the previous behavior.  One reason is preview-only mode scripts need a way to disambiguate grammar when used with ambiguous grammar keywords that start an Execute.Sql block.

PhenX:
> @PhenX The Postgres tests are failing due to the new code missing a trailing semi-colon at the end of the statement.  Any chance you can fix it?  I think we should include the semi-colon, to be clear, which I believe was the previous behavior.  One reason is preview-only mode scripts need a way to disambiguate grammar when used with ambiguous grammar keywords that start an Execute.Sql block.

Yes I feared this detail would fail tests. I'll change the impacted expression. Did I do the right thing removing the overloads that did the same as the base class?

jzabroski:
> Did I do the right thing removing the overloads that did the same as the base class?

I guess it depends if that is why it is breaking.  I am working on a PR to add an `ISeparator` interface to FluentMigrator to separate statements and statement batches. If we had that, we may not need customization per generator?

PhenX:
@jzabroski I pushed the changes, but had to change a few tests for spacing and the condition "1 = 1" present for all other dbms, is that a problem?
Also, when does the pipeline run the checks?

jzabroski:
There's some problem with the pipeline that appears to have started when I renamed the master branch to main. It seems like AzureDevOps still defaults to assuming master in many places.

i have been manually running builds against each pr head commit. Would love to figure this out.

jzabroski:
My take on the Where 1 = 1 is it's an idiom used to silence static analysis tools like RedGate SQL Prompt that warn when a DML statement is missing a where clause. In this sense it makes sense to generate it for .AllRows()

jzabroski:
I kicked off a CI build for this last commit.

PhenX:
> I kicked off a CI build for this last commit.

Oh 😅 looks like I'm not done yet

PR #1931, opened 2024-10-27
Use test containers to facilitate running various integration tests

PhenX:
I fixed a few tests too. I couldn't make Firebird not Oracle Native tests to work, as they require installation of a specific client, which I didn't do.

When running all the tests, when activating Postgresql, Mysql, SQL Server and Oracle containers, take 70 seconds to run when the containers are not already running (but images already downloaded), and nearly 30s when the container is already running.

What do you think ?


PhenX:
@jzabroski In fact, I realized after that the Firebird client seems not to require anything to be installed, and I made a syntax mistake in its connection string (that I just fixed).
I could not make it to work though, whatever user, password or database name I try in either the connection string or the container builder: "unavailable database"

image

I never workd with Firebird, I may be missing something.

About the Oracle native client, I didn't try, but the managed client works well.

test/FluentMigrator.Tests/Containers/FirebirdContainer.cs
```diff
@@ -0,0 +1,33 @@
...
+public class FirebirdContainer : ContainerBase
+{
+    /// <inheritdoc />
+    protected override IntegrationTestOptions.DatabaseServerOptions ServerOptions => IntegrationTestOptions.Firebird;
+
+    /// <inheritdoc />
+    protected override int Port => 3050;
+
+    /// <inheritdoc />
+    protected override DockerContainer Build() => new FirebirdSqlBuilder().WithReuse(true).Build();
```
jzabroski, line 32: I think we should make it easy to do logging.

https://github.com/testcontainers/testcontainers-dotnet/blob/develop/docs%2Fcustom_configuration%2Findex.md 

Nunit supports TestContext.Parameters passed in via --testparam. We can also allow reading a Boolean environment variable for users who always want to turn on container logging: FLUENTMIGRATOR_TESTCONTAINERS_LOGGING_ENABLED

test/FluentMigrator.Tests/Integration/IntegrationTestsSetup.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace FluentMigrator.Tests.Integration
+{
+    [SetUpFixture]
+    public class IntegrationTestsSetup
+    {
+        [OneTimeSetUp]
+        public async Task Setup()
+        {
+            await Task.WhenAll([
```
jzabroski, line 33: Can you measure how much memory is consumed by starting all these?
```
long beforeMemory = GC.GetTotalMemory(true); // Forces a GC collection // Code that allocates memory long afterMemory = GC.GetTotalMemory(true); long usedMemory = afterMemory - beforeMemory; Console.WriteLine($"Managed Memory Used: {usedMemory} bytes");
```
Or does testcontainers have a way to monitor the memory of the container?

test/FluentMigrator.Tests/Integration/Processors/Oracle/OracleProcessorTestsBase.cs
```diff
@@ -14,23 +14,30 @@
...
 #endregion
 
+using System;
+
 using FluentMigrator.Runner.Generators.Oracle;
 using FluentMigrator.Runner.Initialization;
 using FluentMigrator.Runner.Processors.Oracle;
 using FluentMigrator.Tests.Helpers;
 
+using JetBrains.Annotations;
```
jzabroski, line 24: I don't see where you're using this?

test/FluentMigrator.Tests/IntegrationTestOptions.cs
```diff
@@ -39,6 +39,10 @@ static IntegrationTestOptions()
                 .SetBasePath(asmPath)
                 .AddJsonFile("appsettings.json")
                 .AddUserSecrets("FluentMigrator.Tests")
+                // You can specify env vars to override appsettings values, to enable local containers for example:
+                //  * FM_TestConnectionStrings__SqlServer2016__IsEnabled=true
+                //  * FM_TestConnectionStrings__SqlServer2016__ContainerEnabled=true
```
jzabroski, line 44: See comment about logging.

test/FluentMigrator.Tests/appsettings.json
```diff
@@ -17,8 +17,9 @@
             "IsEnabled": false
         },
         "SqlServer2016": {
-            "ConnectionString": "server=.\\MSSQLSERVER2016;Trusted_Connection=yes;database=FluentMigrator",
-            "IsEnabled": false
+            "ConnectionString": "server=.,1433;User Id=sa;Password=yourStrong(!)Password;database=master;Encrypt=false",
```
jzabroski, line 20: is the sql server container using LocalDB or Developer Edition?

test/FluentMigrator.Tests/FluentMigrator.Tests.csproj
```diff
@@ -34,6 +35,12 @@
     <PackageReference Include="Shouldly" Version="4.2.1" />
     <PackageReference Include="Snowflake.Data" Version="4.2.0" />
     <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="2.1.10" />
+    <PackageReference Include="Testcontainers" Version="4.0.0" />
```
jzabroski, line 38: Please add dependabot update group for this.

test/FluentMigrator.Tests/Integration/IntegrationTestsSetup.cs
```diff
@@ -0,0 +1,42 @@
...
+                new MySqlContainer().Start(),
```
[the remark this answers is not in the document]
PhenX: On the Docker side, here are the resource stats for two consecutive runs (first to start the containers, the second to reuse them), Oracle taking nearly 2/3 of the total memory usage :

image

test/FluentMigrator.Tests/Integration/IntegrationTestsSetup.cs
PhenX: I got negative results, weird
image

Concerning WithCreateParameterModifier, I found no documentation about it, but seeing the memory usage below, I wouldn't be concerned abgout it for now.

test/FluentMigrator.Tests/appsettings.json
PhenX: It is 2019 Developer Edition from this image : 

mcr.microsoft.com/mssql/server:2019-CU18-ubuntu-20.04 (https://hub.docker.com/r/microsoft/mssql-server)

image

test/FluentMigrator.Tests/Integration/Migrations/TestMigration.cs
```diff
@@ -30,16 +30,16 @@ public override void Up()
             // successfully tested. At time of implementing, the FK constraint
             // wasn't explicitly used by any tests and so should affect anything.
 
+            Create.Table("Groups")
+                .WithColumn("GroupId").AsInt32().Identity().PrimaryKey()
+                .WithColumn("Name").AsString(32).NotNullable();
+
```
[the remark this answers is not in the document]
PhenX: Yes it failed before this change : 
image

And other tests after the first error fail too, because the Users table already exists, I guess because something went wrong during the cleanup.

test/FluentMigrator.Tests/Containers/FirebirdContainer.cs
PhenX: I can't find a way to inject anything in the IntegrationTestsSetup class to pass a logger to the containers

test/FluentMigrator.Tests/IntegrationTestOptions.cs
PhenX: What do you mean ?

test/FluentMigrator.Tests/Containers/FirebirdContainer.cs
jzabroski: protected override bool TestContainerLogging() => {
    
bool loggingEnabled = false;

if (bool.TryParse(Environment.GetEnvironmentVariable("FM_TESTCONTAINERS_LOGGING_ENABLED"), out var environmentLevelLoggingEnabled) && environmentLevelLoggingEnabled)
{
  loggingEnabled = true;
}

if (TestContext.Exists("FM_TESTCONTAINERS_LOGGING_ENABLED") && bool.TryParse(TestContext.Parameters["FM_TESTCONTAINERS_LOGGING_ENABLED"], out var nunitLevelLoggingEnabled) && nunitLevelLoggingEnabled))
{
    loggingEnabled = true
}


and then in your Builds

var builder =
new PostgreSqlBuilder().WithReuse(true);

if (LoggingEnabled())
{
  // see https://stackoverflow.com/a/78448295
 builder. WithLogging()
}
return builder.Build();

https://docs.nunit.org/articles/nunit/writing-tests/TestContext.html#testparameters

The reason I care about this is debugging some failures may require this level of logging especially on a build server.

PhenX: @jzabroski sure, I totally understand the need for logging, but WithLogging needs a ILogger which I can't get because I don't find a way to inject it or a logger factory into IntegrationTestsSetup. Or I need to use another way to start the containers, but I will need to make sure they are started only once...

jzabroski: Got it. I will look into it today. I primarily use xunit day to day so when these things come up I need to think vs just do from muscle memory.

test/FluentMigrator.Tests/Integration/IntegrationTestsSetup.cs
jzabroski: Yes, TestContainers documentation is not great. This is why I prefer Deffiss.TestEnvironment to TestContainers :-) But, you wrote the PR, not me, so you 100% get more votes on the approach than I do. Thanks for looking into this! Super excited.

test/FluentMigrator.Tests/Containers/FirebirdContainer.cs
jzabroski: https://github.com/fluentmigrator/fluentmigrator/issues/1973

test/FluentMigrator.Tests/IntegrationTestOptions.cs
jzabroski: We could have logging environment variables like `FM_TestContainers_LoggingEnabled`=`true`. No code change necessary. Sorry the comment was extremely brief. I was more excited about what this would allow.

jzabroski:
> @jzabroski In fact, I realized after that the Firebird client seems not to require anything to be installed, and I made a syntax mistake in its connection string (that I just fixed). I could not make it to work though, whatever user, password or database name I try in either the connection string or the container builder: "unavailable database"
> 
> image
> 
> I never workd with Firebird, I may be missing something.
> 
> About the Oracle native client, I didn't try, but the managed client works well.

The maintainer of Firebird is super polite. We may just be able to ask him what we're doing wrong. I believe they have a mailing list to ask questions on. It's been a while since I asked them for something.

PhenX:
After reading test containers releases I saw DB2 ! So I added it, and while adding it I realized I'm an idiot about the Firebird "issue" I talked about : current tests work on a "temporary table" file, that was my issue, so I added a check to use a test containers DB when available.

image

I also fixed tests from a previous PR I did (CanUseRawSqlInUpdateAndDelete).

jzabroski:
Thanks. Going to try to merge this tonight.

Issue #1911, opened 2024-10-07
Incorrect syntax for renaming a column with a default method in MySQL


isaaclyman:
**Describe the bug**
When attempting to rename a column in MySQL 8, if that column has an expression default value (like `CURRENT_TIMESTAMP`), Fluent Migrator incorrectly includes the cosmetic information `DEFAULT_GENERATED` in the query, which causes a MySql exception.

**To Reproduce**

Migration:

```csharp
Create.Table("my_table").WithColumn("my_column").AsDateTime().NotNullable()
  .WithDefault(SystemMethods.CurrentDateTime);

Rename.Column("my_column").OnTable("my_table").To("anything");
```

Generated SQL:

```sql
ALTER TABLE `my_table` CHANGE `my_column` `anything` datetime NOT NULL DEFAULT `CURRENT_TIMESTAMP` DEFAULT_GENERATED
```

Error:

> You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DEFAULT_GENERATED' at line 1

**Expected behavior**
I would expect the rename to ignore the `DEFAULT_GENERATED` and generate a correct query.

**Information (please complete the following information):**
 - OS: MacOS Sonoma 14.6.1
 - Platform: .NET Core 8.0
 - FluentMigrator version: 5.2.0
 - FluentMigrator runner: in-process runner
 - Database Management System: MySQL
 - Database Management System Version: 8.0

**Additional context**
The Phinx migration library for PHP faced a similar issue: https://github.com/cakephp/phinx/pull/2253

jzabroski:
https://github.com/fluentmigrator/fluentmigrator/blob/6fe6395fd3fd9e933e5f49f648aa9bef82ddc55f/src/FluentMigrator.Runner.MySql/Generators/MySql/MySql4Generator.cs#L101-L104

You would need to copy-paste this and add it to the Mysql8Generator, and update the test for Mysql8Generator to not include this.

The interface is here:
https://github.com/fluentmigrator/fluentmigrator/blob/main/test/FluentMigrator.Tests/Unit/Generators/BaseColumnTests.cs#L41-L42

You need to add a new stub tests for 👍 
```c#
        public abstract void CanRenameColumnWithCustomSchemaAndDefaultValue();
        public abstract void CanRenameColumnWithDefaultSchemaAndDefaultValue();
```

[...]

PR #1945, opened 2024-12-01
Incorrect syntax for renaming a column with a default method in MySQL #1911 (Fix Column Renaming Syntax for MySQL 8)

Cretesh-Anuj:
This pull request addresses the issue of incorrect SQL syntax when renaming columns in MySQL, particularly when the column has an expression default value (e.g., CURRENT_TIMESTAMP). The current implementation erroneously includes DEFAULT_GENERATED in the generated SQL, resulting in syntax errors during migration.

Changes Made
Modified the MySql8Generator to ensure that the SQL generated for renaming columns does not include the DEFAULT_GENERATED clause.
Implemented new unit tests to validate the renaming functionality for columns with default values in MySQL 8.


jzabroski:
Thanks. Changes look good - need to run it through the build server but will otherwise merge once tests pass. I'm having some difficulty with the "fluentmigrator CI for PRs" pipeline triggering.

jzabroski (approved):

PR #1963, opened 2025-01-23
Fix auto reversing with insert data

AdrianoRNascimento:
When a table has an FK to itself, the current approach of auto-reverting the inserts fails because it tries to delete rows in the same order they were inserted.

As an example, SQL Server generates this error:
```
The DELETE statement conflicted with the SAME TABLE REFERENCE constraint "<FK_Name>".
```

To solve this error, deleting the rows in the inverted order they were inserted is appropriate, exactly as it happens with the expressions.


jzabroski:
This looks great. Assuming the build passes I will merge it. Thanks for your patience! Had a few important unexpected family events recently.

jzabroski (approved):

jzabroski:
Apologies while I was sorting out some build pipeline stuff.

It looks like your code has an issue 

https://dev.azure.com/fluentmigrator/fluentmigrator/_build/results?buildId=1324&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=86dd5fa6-f51e-57a1-6dd4-575fd75fc7ef

AdrianoRNascimento:
@jzabroski 
Thank you.
The build was complaining about the `Assert.IsAssignableFrom` method.
Just pushed the fix.

jzabroski:
@AdrianoRNascimento Your test stub class for auto-reversing broke an unrelated test that was checking to see if we applied all migrations in the assembly - I fixed it and added it to your branch. Re-running the build now and all should pass, then we will merge it. Thanks for your wonderful contribution.

jzabroski (approved):
Address test failure: Foreign keys are not supported in SQLite

AdrianoRNascimento:
@jzabroski 
No problem. I'll try to contribute more often as I use fluentmigrator intensively.
Thank you for the amazing tool.

PR #1984, opened 2025-03-05
Feature/nunit testcasesource

PhenX:


test/FluentMigrator.Tests/Integration/TestCases/ProcessorTestCaseSourceExcept.cs
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace FluentMigrator.Tests.Integration.TestCases;
+
+public class ProcessorTestCaseSourceExcept<TIgnore1>() : ProcessorTestCaseSource(true, typeof(TIgnore1));
```
PhenX, line 19: I just realized that I forgot the type constraints here, I think this should be enough to ensure generic arguments derive from ProcessorBase what to you think ? 
The ProcessorTestCaseSource constructor is protected by the way.

jzabroski: Why not change the TIgnore1 stuff to be TMigrationProccesor1 where TMigrationProcessor : ProcessorBase

jzabroski:
The build passed

test/FluentMigrator.Tests/Integration/TestCases/ProcessorTestCaseSourceExcept.cs
PhenX: Yep that's what I did, just not with the same name

jzabroski:
@PhenX I would like to merge this if its OK with you?  What is it missing?  I know you like using 'Draft' PR status, so wasn't sure.

PhenX:
@jzabroski I think it's complete! I'm removing the draft status

PR #2025, opened 2025-03-30
Add StatementFormat to handle more consistently semicolons

PhenX:
In this PR I made a change to handle more consistently statement separators (semicolon here), I may add a ISeparator like you suggested, only for statements for now maybe.

I also removed code in overriden methods that acted like parent classes.

For now all the tests don't pass but I'm working on it.

Tell me what you think and if that's going in a good direction


src/FluentMigrator.Runner.Db2/Generators/Db2/Db2Generator.cs
```diff
@@ -50,14 +50,9 @@ public Db2Generator(
         {
         }
 
-        protected override StringBuilder AppendSqlStatementEndToken(StringBuilder stringBuilder)
-        {
-            return stringBuilder.Append(" ");
```
jzabroski, line 55: This actually seems incorrect; initial thought, should default to semi-colon.

Per https://stackoverflow.com/a/77266596/1040437

> When scripting in Db2, it is necessary to tell the tool (that submits the SQL) how to distinguish between the end of a statement (default ; ), or the end of a block. A block contains multple statements.
> 
> In your script you have an anonymous block, between BEGIN and END. So you need to delimit the end of the block with a different character than ; . Many people use the @ symbol although other characters are also allowed.
> 
> [...]
> 
> For example, if you submit your script via the db2 command line processor (at the operating system command line shell), or if you can configure you tool to shell out to the operating-system to run scripts, then you can include --#SET TERMINATOR @ inside your script, or use the -td@ command-line flag. This --#SET... can appear multiple times inside a script if required, using different delimiters if neeed. You then use BEGIN...END@ for the block, and any subsequent statements would be terminated with the @ (instead of @) until and unless you change the terminator again inside the script.

jzabroski (commented):
Need to see what unit test CI says but only have 2 main comments.

PhenX:
OK, there are actually a few tests that still don't pass on my machine, but I'll fix them.
Is the change about the semicolons that are added to all the statement ok for you? Because I was not sure that it would not be considered a massive breaking change.

jzabroski:
My thought is we should put together an ADR (Architecture Decision Record) for an `ISeparator` interface.  As I mentioned, I had changes for this in my local dev environment, just hadn't pushed it up yet.  I think, taking a step back, I should have started with an ADR to explain / motivate the change.  

I honestly don't know if the semicolons added to all the statements is OK or not, but my definition of OK would be if it doesn't break any consumers and is functionally the same, then it is "OK".  Does that make sense?

In terms of what the ADR would talk about:

1. MySQL default is ; but supports `DELIMITER <char sequence>` statement to override
2. Postgres supports "[dollar quoting](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)" e.g.
   ```pgsql
   $function$
   BEGIN
    RETURN ($1 ~ $q$[\t\r\n\v\\]$q$);
   END;
   $function$ 
3. Firebird default is ; but supports`SET TERM` keyword phrase
4. DB2 default is ; but ` --#SET TERMINATOR` phrase
5. Oracle SQLPlus `set cmdsep` phrase
6. Hana: In SAP HANA, the default statement delimiter for executing multiple SQL statements is a semicolon (;), but you can specify a different delimiter using the -c <separator> command-line option in HDBSQL. 
7. T-SQL and SyBase: ; and GO for batches.
8. Redshift uses ;
9. and so on

PhenX:
That makes absolutely sense. I'm not sure about the dollar quoting though. Isn't it about IQuoter?
I think my changes may not go against what you started to work on, it that is about the ISeparator.
I was curious about what EF Core did with migration scaffolding, and saw that they use a single dependency for quotes and separators, what do you think about it?

src/FluentMigrator.Runner.Db2/Generators/Db2/Db2Generator.cs
PhenX: I'm finally not sure I'm following you, it now defaults to semicolon (in GeneratorBase) because I removed this override, am I missing something ?

jzabroski: I mean the existing behavior seems so wrong that i wonder if anyone is using it

PhenX: Oh yeah ! that's weird

PhenX:
@jzabroski All tests pass, what do you think about implementing the ISeparator in this pull request ?

jzabroski:
> @jzabroski All tests pass, what do you think about implementing the ISeparator in this pull request ?

I meant to reply last week but was pretty down with allergies. I will try tonight to push up to my jzabroski/FluentMigrator repo what I have so you can look at it.

jzabroski:
I'm getting back to this today. Was pretty low on energy due to seasonal allergies the last two weeks, my apologies.

PhenX:
No worries ! Take care

jzabroski:
@PhenX Plan to merge this today

PR #2126, opened 2025-08-25
Make Oracle integration tests run and fix Oracle single statement execution code

PhenX:


src/FluentMigrator.Runner.Oracle/Processors/Oracle/OracleProcessorBase.cs
```diff
@@ -265,6 +290,92 @@ public override void Process(PerformDBOperationExpression expression)
             expression.Operation?.Invoke(Connection, Transaction);
         }
 
+        /// <summary>
+        /// Splits a SQL script into individual statements, taking into account Oracle-specific syntax rules.
+        /// </summary>
+        private static List<string> SplitOracleSqlStatements(string sqlScript)
```
PhenX, line 298: I just noticed (again) the existence of SemicolonSearcher, SqlBatchParser etc, I should use it, what do you think ?

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoter.cs
```diff
@@ -16,33 +16,85 @@
...
+        [ContractAnnotation("name:null => false")]
+        protected override bool ShouldQuote(string name)
         {
-            return base.QuoteConstraintName(UnQuote(constraintName), UnQuote(schemaName));
+            if (string.IsNullOrEmpty(name))
+            {
+                return false;
+            }
+
+            if (_keywords.Contains(name))
```
jzabroski, line 60: Why not just always quote identifiers?

src/FluentMigrator.Runner.Oracle/Processors/Oracle/Oracle12CManagedProcessor.cs
```diff
@@ -58,6 +58,6 @@ public Oracle12CManagedProcessor(
 
         /// <inheritdoc />
         public override IList<string> DatabaseTypeAliases { get; } = new List<string>()
-            { "Oracle-12c-Manager", "Oracle 12c Managed", "Oracle Managed", "Oracle" };
+            { "Oracle-12c-Managed", "Oracle 12c Managed", "Oracle Managed", "Oracle" };
```
jzabroski, line 61: 😰

src/FluentMigrator.Runner.Oracle/Processors/Oracle/OracleProcessorBase.cs
PhenX: I tried to use it (with the same implementation as SnowflakeBatchParser) but it seems that it matches semicolons inside "ranges", in my case : strings 🫠
I'm too afraid to change it 😨

jzabroski: I am not a big fan of the batch parser. sql is a very complicated dialect

PhenX: Oh I think I understood ... My string spans multiple lines (I'm in the long string tests for Oracle)

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoter.cs
PhenX: Because it would be too breaking. Quoted identifiers in Oracle become case sensitive 🙃
Needless to say it breaks everything

src/FluentMigrator.Runner.Oracle/Processors/Oracle/OracleProcessorBase.cs
PhenX: > I am not a big fan of the batch parser. sql is a very complicated dialect

Yeah, neither am I, but it still a lot better than the Regex previously used

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoter.cs
jzabroski: Got it. I would put that in a separate PR though with a separate issue, right? Because currently we're already in case sensitive behavior, which I can see as non-desirable. We would need to list that in the release notes as a breaking change to move to case sensitive by default. We may also want a way to easily allow forcing case sensitive identifiers to maintain backward compatibility, right?

src/FluentMigrator.Runner.Oracle/Processors/Oracle/OracleProcessorBase.cs
jzabroski: The simplest regex would be if the line ends in a `;\s*$`

PhenX: how to handle strings spaning multiple lines, containing a semicolon ? I'm sure the batch parser could be solid, but it needs to handle "ranges" spanning multiple lines, which it does not ATM.

src/FluentMigrator.Runner.Oracle/Generators/Oracle/OracleQuoter.cs
PhenX: Why not, but without it the integration tests on Oracle won't work, or I ignore them for now ?

jzabroski: Ok, can you create a separate Issue and I will put it in the 7.2 milestone. I think GitHub will just generate release notes correctly.

PhenX: #2127

src/FluentMigrator.Runner.Oracle/Processors/Oracle/OracleProcessorBase.cs
PhenX: I finally gave up on using the SqlBatchParser, its handling of ranges line-by-line seems broken

jzabroski: my regex hack makes it so long as the last non whitespace character on a line is a `;` , it will treat it as a separator.

PR #2171, opened 2025-10-26
Added namespace filtering to maintenance migrations

spopp23:
Hi,

I noticed that the the namespace filtering of migration scripts is not applied to maintenance migrations and provided the necessary change for it.
I hope this is something you guys find useful as well.

Please let me know if you have any feedback or suggestions for improving my PR.

Best regards,
Stefan

PR #2173, opened 2025-11-10
Limit CI test targets by OS: run net8.0 on Linux and net48 on Windows only

spopp23:
The Windows job in CI was failing due to running a net48 test target on Linux (see PR #2171).
PR #2162 appears to have introduced an extra test run (net8.0 on Windows) although discussion suggested net48 should be Windows-only.

## What I changed

Linux runners now execute tests only for net8.0.

Windows runners now execute tests only for net48.

## Why

According to the comments on PR #2162  .NET 8.0 is supposed to run on Linux only; running it on Windows is unnecessary and fails.

Keeps net8.0 coverage on Linux while preserving net48 coverage on Windows.

## Impact

Green CI on Windows again.

No loss of coverage—just correct platform targeting.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -26,8 +26,6 @@ jobs:
             dotnet-tfm: net48
           - os: ubuntu-latest
             dotnet-tfm: net8.0
-#          - os: ubuntu-latest
-#            dotnet-tfm: net9.0
 
     steps:
     - name: Checkout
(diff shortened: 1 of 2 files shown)


jzabroski:
I changed it slightly; honestly, the way this is done is not great, I just haven't had the bandwidth to clean it up. Going to make some progress this month in prep for Microsoft releasing .NET 10 tomorrow.

spopp23:
Yeah, I just noticed. Thank you! 

I will close this PR then.

jzabroski:
We can still go with your approach as a base.

jzabroski:
I was struggling to understand why this was still failing, and then I realized that its due to how I set-up permissions to prevent people from submitting PRs that change this file to execute Confused Deputy attacks and do things like potentially steal variables, viz a viz 'pull request targets' security feature of GitHub Actions.

Will merge; think we are good here.

PR #2193, opened 2025-11-22
Refactor: Update ColumnBase to use generic type and fix deprecation w…

PhenX:
…arnings


jzabroski (approved):
LGTM. But my thought is we publish 7.2 without net10.0 support and publish 8.0 with net10.0 and this. Thoughts?

PhenX:
@jzabroski 7.2 has so many changes that it could be 8.0 😅 
But yeah, I'm ok with this

jzabroski:
they're not really breaking changes, though. the ones that are, have been marked obsolete for almost a decade lol 😆

PR #2212, opened 2025-12-26
Fix tag matching for RequireAll behaviour to use Any instead of All

amitkumar-sh:
Previously, when matching migrations with RequireAll tag behavior, the code required ALL requested tags to match ALL migration tags. This was too strict and prevented migrations with Tags from being matched.

The fix changes the logic to check if ANY requested tag matches ANY migration tag, which aligns with the expected behavior: when you request migrations with tags Schema and Data, you want migrations that have either Schema OR Data tags (or both), not migrations that must have both tags.

### DIFF
--- a/src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
+++ b/src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
@@ -128,9 +128,10 @@ namespace FluentMigrator.Runner.Infrastructure
                 return false;
 
             var tagNamesForAllBehavior = tags.Where(t => t.Behavior == TagBehavior.RequireAll).SelectMany(t => t.TagNames).ToArray();
-            // For RequireAll behavior: Check if ANY requested tag matches ANY migration tag
-            // This allows migrations with [Tags("Schema")] to match when requested tags are ["Schema", "Data"]
-            if (tagNamesForAllBehavior.Any() && matchTagsList.Any(t => tagNamesForAllBehavior.Any(tag => tag.Equals(t))))
+            // For RequireAll behavior: ALL of the migration's tags must be present in the requested tags
+            // Migration: [Tags("Schema")], Request: ["Schema", "Data"] → TRUE (Schema is in the request)
+            // Migration: [Tags("Schema", "Data")], Request: ["Schema"] → FALSE (Data is missing)
+            if (tagNamesForAllBehavior.Any() && tagNamesForAllBehavior.All(tag => matchTagsList.Any(t => tag.Equals(t))))
             {
                 return true;
             }


PhenX:
I'm sorry, but I don't see any difference with RequireAny, am I missing something ?

amitkumar-sh:
@PhenX updated the code.

The original code checked if all requested tags are in migration tags (backwards). This fix correctly checks if all migration tags are in requested tags.

Example: Migration [Tags(Schema)] with request [Schema, Data] should match TRUE because Schema is in the request.

jzabroski:
> Example: Migration [Tags(Schema)] with request [Schema, Data] should match TRUE because Schema is in the request.

Hi, happy new year!

1. There are no unit tests, so I am not going to accept this. If you think there is a bug, I would recommend opening a BUG issue type.
2. Please see the existing tests here - https://github.com/fluentmigrator/fluentmigrator/blob/d73e81e49849cf013e1963e1a9be0204f6fb2c6c/test/FluentMigrator.Tests/Unit/DefaultMigrationConventionsTests.cs#L272-L531
3. That said, I am surprised the checks passed for this PR despite the tests linked above. Something does seem off if you managed to make a change that failed none of the tests, so at minimum we are missing some test coverage... or these tests are not getting run on GitHub PRs.

The default behavior is to match All incoming tags, which is documented https://github.com/fluentmigrator/fluentmigrator/blob/d73e81e49849cf013e1963e1a9be0204f6fb2c6c/docs-website/migration-types/tags.md?plain=1#L67 https://fluentmigrator.github.io/migration-types/tags.html#requireall-default

I wonder if you are thinking about this backwards, and expecting all the tags to exist on the migration. It does not work that way, consider the following example:

```c#
[Tags("Development", "NonProduction", "Production")]
[Migration(0, "Migrate up")]
public class M0_MigrateUp : ForwardOnlyMigration
{
  public void Up() {
   // Do stuff
  }
}
```

If you pass in Development from a Continuous Deployment server intending to deploy to the Development environment, it should match this migration and execute it. Similarly (and especially) for maintenance migrations.

Separately, I'm not sure why you are directly tagging migrations as Data vs. Schema. While I do separate out the two, I don't see the utility in tagging them and filtering migrations by those concepts. Normally, the only reason I put these in separate migrations is readability as well as decreasing transaction sizes so that the transaction log has a chance to commit work and create a checkpoint for log and differential backups. - In the test cases, the tags "CA", "UK", etc denote different regions or "tenants", and the Data migraiton component varies by tenant, which does make sense. - If your reason to tag Data migrations is based on the environment, I recommend creating Environment named tags.

amitkumar-sh:
@jzabroski happy new year! 

Thanks for the reply. 

I managed to get this working:
 [Tags(TagBehavior.RequireAny, "Schema")]
and
[Tags(TagBehavior.RequireAny, "Data")]

We need Schema vs Data tagging because our reporting server needs only Schema changes. Data is synced through different route

jzabroski:
Yes, that is the correct way to do it as Schema and Data are technically orthogonal groups of concerns. Happy you found a solution.

Issue #2053, opened 2025-06-04
Simplify Migrate() call to be as simple as EntityFrameworkCore


jzabroski:
**Is your feature request related to a problem? Please describe.**
EntityFrameworkCore's solution in a single page application has gotten very simple, and with the new .NET 10 #:package syntax, we really need to think about user experience to make FluentMigrator adoption more likely in the future, or we will fall behind ergonomically due to having a slightly steeper adoption curve. Simple wins.

**Describe the solution you'd like**
Here is an example taken from the CryptomonServer github open source project for how they configure EntityFrameworkCore code-first migrations:

```c#
using (var scope = app.Services.CreateScope())
{
    var db = scope.ServiceProvider.GetRequiredService<CryptomonDbContext>();
    db.Database.Migrate();
    db.Database.OpenConnection();
    // Npqsql provider caches type information, so need to reload it after migrations complete.
    ((NpgsqlConnection)db.Database.GetDbConnection()).ReloadTypes();
}
```

**Additional context**
https://github.com/MinaNavigator/CryptomonServer/blob/dec1b472bf1465d2038d2dc822f80c0a355d2cff/CryptomonServer/Program.cs#L68C1-L74C2

**Is this a feature you'd like to submit a PR for?**
Yes

DanielTuran:
EF Core also requires you to register migrations somewhere, so this EF core sample is not quite complete.

I was thinking:

```
var services = new ServiceCollection()
   .AddFluentMigrator<Database1Migrator>()
   .Build();

var database1Migrator = services.GetRequiredService<Database1Migrator>();
database1Migrator.MigrateUp();

class Database1Migrator : FluentMigratorBase
{
     protected override void Configure(FluentMigratorOptions options)
     {
          options.ConnectionString = ....
          options.AddMigrationsFromAssembly(....)
          options.VersionInfo.SchemaName = ....
     }
}
```

It would be then easy to  setup multiple migrators.

PR #2239, opened 2026-01-25
[Experiment] EF core - code first migration generator

PhenX:
In this experiment I tried to see how far we could go to work together with EF Core _Code first_.

It take advantage of the EF Core model snapshot feature to create Fluent Migrator migtrations, from the EF Core model.

Typical scenario in code first : 

 * Developer writes models and adapt his DbContext
 * He runs `dotnet ef migrations add <MigrationName>` to generate the missing migrations, that he can adapt as he wants, or add migrations himself.
 * He runs `dotnet fm up` (didn't try this part yet)

There are details in the [example project readme](https://github.com/fluentmigrator/fluentmigrator/blob/experiment-code-first-ef-core/samples/FluentMigrator.Example.EFCore/README.md).


PhenX:
@jzabroski I tried something in this pull request. I personally don't use code first with EF core yet, but we may change it at work, we are currently doing something in between 😅 
While we love EF core, we also love FM, so this may help us take the leap, and it may help others adopt Fluent Migrator too.

jzabroski:
Smart idea. Related is the issue i created to simplify instantiating a runner.

PhenX:
Thank you! Do you think it can make it as an official package, with refinement of course.
I'll make an ADR if you want. 

What issue are you talking about?

jzabroski:
This is the other issue I was thinking of being related to this: https://github.com/fluentmigrator/fluentmigrator/issues/2053

Where should the intelligence lie for this feature? In the Roslyn code generator, or inside FluentMigrator?

If it lives in FluentMigrator, why not create a new top-level syntax called simply Define.  Rather than force users to know the prior state of migrations, they could simply write Define operations with or without code gen support.  We've indirectly seen proof-of-concept of such ideas by Quentin Starin's EasyMigrator plug-in that sits on top of FluentMigrator, but he did not create a top-level syntax.

The idea behind Define is that it simply describes the outcome, rather than the procedural steps required.  Then internally the Define syntax generator is responsible for building the migration SQL.

Just some food for thought. For V1, it's probably fine to have the logic live in this external code generator, but probably as a "public preview" rather than RTM type release.

What would your current team feel most comfortable debugging? I have not used EF code-first in about 5 years so I am not deeply familiar with how it currently works, but back in the day the only support they had was that lousy ToDebugString() method on the modelBuilder, which was still useful but not really easy to debug and understand what SQL it generated, etc. The way I told Andrew Vickers it should work is as a full outer join of the POCO and SQL table mapping logic. By creating a cross-product of all possibilities, you can then detect ambiguous mappings as any side that has duplicate mapping rows.

PhenX:
You mean recreate what "ef migrations add" does? At this time it compares the hole model to a snapshot file (under source control) to generate the diff as migrations that the developer can customize when needed. Having these files to the hand of the developer is necessary in my opinion : for him to keep control of the changes, and to customize it. Making it a source generator does not seem a viable idea to me. 
If I'm not mistaken, EasyMigrator does not compare current schema to the C# model (What you would call Define), the gap to make a diff seems huge to me, and EF does this already. 
Of course we can discuss it deeper, I may not have understood well what you meant 😉

jzabroski:
Define would just load a yaml file, something like dbt/ lightdash schema.yml files 

Dbt uses schema.yml  - we could reuse and interoperability with it to allow smooth transition from python to c#. Think many shops have python and c# as a combo, so it makes sense as they're very similar languages these days.

jzabroski:
@copilot+claude-fable-5 Review this PR - Feel free to also make improvements.

PR #2252, opened 2026-02-20
NativeAOT support

### LINKED ISSUE #2231: Does FluentMigrator support AOT?
I want to use  FluentMigrator  in  my desktop app with SQLite, and i want o know whether  FluentMigrator supports AOT. 
Any response would be appreciated.

GerardSmit:
Fixes #2231

**AI Disclaimer**
The first commit (`937b960`) and this text was handwritten.
The merge commit and last commit were made together with Copilot; to revert most of the changes that I made in the first commit 😅

The rest of the description is also put together with Copilot so it's easier to understand (English is my second language) 😉
All decisions were made and validated by me, so if anything needs to be changed feel free to ask.

---

## Why AOT Didn't Work Before

FluentMigrator previously relied entirely on runtime reflection to find migration types:

1. **Assembly scanning** — `AddFluentMigratorCore()` registered `AssemblySource` and `AssemblyTypeSource`, which use `AppDomain.GetAssemblies()` and `Assembly.GetExportedTypes()` to discover all migration types at runtime. The .NET trimmer cannot statically determine which types will be loaded through these code paths, so it may strip required types from the output.

2. **Missing `[RequiresUnreferencedCode]` annotations** — None of the reflection-heavy APIs carried trimming annotations. Trimmed builds would silently produce incorrect binaries with no compiler warnings.

3. **Anonymous-type overloads** — `Row(object anonymousType)`, `Set(object)`, etc. use `GetProperties()` at runtime to read the object''s properties. The trimmer removes property metadata unless explicitly preserved.

4. **Assembly loading from disk and GAC** — `ReflectionBasedDbFactory` falls back to loading provider assemblies from the file system and the GAC. Both mechanisms are completely unavailable in NativeAOT.

5. **`ValidationContext` in the expression validator** — `DefaultMigrationExpressionValidator` relies on `System.ComponentModel.DataAnnotations`, which is not trimming-safe.

---

### New Abstractions

| Type | Description |
|---|---|
| `ITypeSource` | New interface with a single `GetTypes()` method. Decouples migration type discovery from assembly scanning. |
| `AssemblyTypeSource` | Implements `ITypeSource` by scanning `IAssemblySource`. Annotated `[RequiresUnreferencedCode]` — preserves existing runtime behaviour but not AOT-safe. |
| `ArrayTypeSource` | Implements `ITypeSource` from an explicit `Type[]`. No reflection — AOT-safe. |
| `EmptyAssemblySource` | An `IAssemblySource` that returns no assemblies. Used when assembly loading is intentionally excluded. |
| `TypeSourceConventionSetAccessor` | An `IConventionSetAccessor` that resolves convention sets from an `ITypeSource` instead of assembly scanning. |

### New APIs

| Member | Description |
|---|---|
| `IServiceCollection.AddFluentMigratorSlim()` | Sets up all runner services **without** the assembly-loader infrastructure. AOT-safe. Named after ASP.NET Core's `WebApplication.CreateSlimBuilder()`, which follows the same convention of providing a minimal, AOT-friendly setup. |
| `IMigrationRunnerBuilder.WithTypes(params Type[])` | Register migration types explicitly. Use together with `AddFluentMigratorSlim()` as the AOT entry point. |

#### Usage examples

**Default (reflection-based, not AOT-safe) — unchanged:**
```csharp
// Discovers migrations from assemblies at runtime.
// Produces an IL2026 warning in trimmed builds.
services.AddFluentMigratorCore()
    .ConfigureRunner(rb => rb
        .AddSQLite()
        .WithGlobalConnectionString("Data Source=:memory:")
        .ScanIn(typeof(MyMigration).Assembly).For.Migrations());
```

**AOT-safe (explicit type registration):**
```csharp
// No IL2026 warning. Types are known at compile time.
services.AddFluentMigratorSlim()
    .ConfigureRunner(rb => rb
        .AddSQLite()
        .WithGlobalConnectionString("Data Source=:memory:")
        .WithTypes(typeof(Migration_001_CreateUsers), typeof(Migration_002_AddEmail)));
```

### Trimming Annotations Added

The following members now carry `[RequiresUnreferencedCode]` on the `NET` build target, correctly flagging reflection usage to the trimmer:

- `IInsertDataSyntax.Row(object)` / `Rows(params object[])` — anonymous type reflection
- `InsertDataExpressionBuilder.Row(object)` / `Rows(params object[])`
- `UpdateDataExpressionBuilder.Set(object)` / `Where(object)`
- `MigrationSource` constructor (when no explicit `ITypeSource` is provided)
- `MaintenanceLoader` constructor (when no explicit `ITypeSource` is provided)
- `AssemblySourceConventionSetAccessor`
- `AssemblyLoaderFromFile`
- `RuntimeHost.FindAssemblies()` / `FindAssemblies(string)`
- `ReflectionBasedDbFactory.TryCreateFactory()` / `FindAssembliesInGac()`

### Projects Targeting `net8.0`

All runner packages now target `net8.0` with `<IsTrimmable>true</IsTrimmable>`, except `FluentMigrator.Runner.Jet`. The Jet provider uses `System.Data.OleDb`, which itself is not trimmable or NativeAOT-compatible, making a `net8.0` target impractical.

---

## Breaking Changes

1. **`[RequiresUnreferencedCode]` on anonymous-type API methods**
   Callers using `Row(object anonymousType)`, `Rows(params object[])`, `Set(object)`, `Where(object)` in a trimmed build will now receive an analyzer warning (`IL2026`). Switch to the `IDictionary<string, object>` overloads to avoid this.

2. **`MaintenanceLoader` constructor signature**
   An optional `ITypeSource typeSource = null` parameter was added. Existing DI-registered usage is unaffected; only direct instantiation (`new MaintenanceLoader(...)`) may need updating.

3. **`MigrationSource` now implements `IMigrationSource`**
   Previously, `MigrationSource` only implemented `IFilteringMigrationSource`. It now also implements `IMigrationSource` to support DI registrations that request the non-filtering interface.

---

## Database Provider Support

Every provider factory uses `Type.GetType("FullTypeName, AssemblyName")` with a string literal. The trimmer recognises this pattern and preserves the referenced type. The `TryGetInstance` helper carries the necessary `[DynamicallyAccessedMembers]` annotation. When the provider assembly is statically referenced, **this path is fully trim-safe**.

The fallback paths (AppDomain scanning, file loading, GAC) are **not** NativeAOT-compatible, but they are already guarded by `RuntimeFeature.IsDynamicCodeSupported` and are skipped at runtime in an AOT context.

| Provider | Factory path | Fully trimmed |
|---|---|---|
| All providers | ✅ `Type.GetType(literal)` is trim-safe when the driver assembly is a static reference | ❌ Processor/generator code not yet audited |
| All providers | ⚠️ Fallback paths (AppDomain/file/GAC) are runtime-guarded — not called in AOT | |

---

### 1. Source generator
Consumers must manually call `WithTypes(typeof(Migration1), ...)`. A Roslyn Incremental Generator that scans for `[Migration]`, `[Profile]`, and `[Maintenance]` attributes at compile time and emits the `WithTypes` call would make AOT adoption seamless.

### 2. Trimming-safe expression validator
`DefaultMigrationExpressionValidator` uses `System.ComponentModel.DataAnnotations.ValidationContext`, which is not trimming-safe. In a trimmed build, validation is currently skipped via a `RuntimeFeature.IsDynamicCodeSupported` guard. A proper attribute-based validator would restore validation in AOT builds.

### 3. Remaining reflection in the runner
The following areas have not yet been audited for AOT:
- `MigrationRunner` (activating migration instances via reflection)
- `ProfileLoader`
- `VersionTableMetaDataAccessor`
- `TypeFinder` and related convention helpers that use `Type.GetCustomAttributes`

### 4. Provider-specific code
Each provider's processor and generator code may contain additional reflection that has not yet been annotated. The `net8.0` target and `<IsTrimmable>true</IsTrimmable>` are now in place; a full trimming audit per provider is still needed. The factory pattern is already correct — the remaining work is in the per-provider implementation classes.

### 5. AOT integration test
An end-to-end test that publishes a minimal AOT binary and runs a migration (e.g. against SQLite) is needed to validate the complete stack. NUnit does not support NativeAOT, so a test runner switch (e.g. to TUnit) is required before this can be implemented.


src/FluentMigrator.Abstractions/Expressions/ExecuteEmbeddedSqlScriptExpressionBase.cs
```diff
@@ -77,10 +77,9 @@ protected static (string name, Assembly assembly) GetQualifiedResourcePath(
             [NotNull] string sqlScriptName)
         {
             //resource full name is in format `namespace.resourceName`
-            var sqlScriptParts = sqlScriptName.Split('.').Reverse().ToArray();
+            var sqlScriptParts = Enumerable.Reverse(sqlScriptName.Split('.')).ToArray();
```
GerardSmit, line 80: This is needed because of fist-class spans, see: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/first-class-span-types#calling-reverse-on-an-array

src/FluentMigrator.Abstractions/Validation/DefaultMigrationExpressionValidator.cs
```diff
@@ -45,6 +45,15 @@ public DefaultMigrationExpressionValidator([CanBeNull] IServiceProvider serviceP
         /// <inheritdoc />
         public virtual IEnumerable<ValidationResult> Validate(IMigrationExpression expression)
         {
+#if NET
+            if (!System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported)
+            {
+                return Enumerable.Empty<ValidationResult>();
```
GerardSmit: This probably needs to be changed. Maybe an exception, or we should implement this before merging?

src/FluentMigrator.Runner.Core/Infrastructure/RuntimeHost.cs
```diff
@@ -104,24 +107,6 @@ public static IEnumerable<AssemblyName> FindAssemblies(string name)
...
-        /// <returns>
-        /// An <see cref="IEnumerable{T}"/> of <see cref="AssemblyName"/> objects representing the assemblies found
-        /// in the specified directory.
-        /// </returns>
-        /// <remarks>
-        /// This method iterates through subdirectories of the specified GAC directory, parsing metadata such as 
-        /// version, culture, and public key token to construct <see cref="AssemblyName"/> instances for each assembly.
-        /// Invalid or incomplete metadata is ignored.
-        /// </remarks>
         private static IEnumerable<AssemblyName> GetAssemblyNames(string fullGacDirectory, string assemblyName)
```
GerardSmit, line 128: Ah.. I missed this, I will revert this after the previous comments has been answered and maybe need additional work.

src/FluentMigrator.Runner.Core/Initialization/AssemblyLoader/AssemblyLoaderFromFile.cs
```diff
@@ -60,13 +62,13 @@ public AssemblyLoaderFromFile(string name)
...
-        /// <exception cref="FileNotFoundException">
+        /// <exception cref="System.IO.FileNotFoundException">
         /// Thrown if the specified assembly file does not exist.
         /// </exception>
-        /// <exception cref="BadImageFormatException">
+        /// <exception cref="System.BadImageFormatException">
         /// Thrown if the file is not a valid assembly or was compiled with a later version of the runtime than is currently loaded.
         /// </exception>
-        /// <exception cref="FileLoadException">
+        /// <exception cref="System.IO.FileLoadException">
```
GerardSmit: This is a Rider-specific change. Rider couldn't find the type back then. Not sure if this is fixed. Should I revert this?

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -28,33 +28,30 @@ namespace FluentMigrator.Runner.Processors
...
     using System.Data.Common;
 
-    /// <summary>
-    /// Provides a database factory implementation that uses reflection to dynamically load database provider factories.
-    /// </summary>
-    /// <remarks>
-    /// This class serves as a base for creating database provider factories by attempting to load assemblies and types
-    /// dynamically at runtime. It supports scenarios where database drivers are not directly referenced in the project
-    /// but need to be loaded from external assemblies.
-    /// </remarks>
```
GerardSmit, line 38: Ditto. Will revert comments in this file as well.

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -598,40 +460,14 @@ private static bool TryGetInstance(
...
-        /// When this method returns, contains the <see cref="DbProviderFactory"/> instance if the cast was successful;
-        /// otherwise, <c>null</c>.
-        /// </param>
-        /// <returns>
-        /// <c>true</c> if the object was successfully cast to a <see cref="DbProviderFactory"/>; otherwise, <c>false</c>.
-        /// </returns>
-        /// <exception cref="ArgumentNullException">Thrown if the <paramref name="value"/> is <c>null</c>.</exception>
         private static bool TryCastInstance(
-            [NotNull] object value,
+            [CanBeNull] object value,
```
GerardSmit: Not sure why this was changed. Will change back.

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
```diff
@@ -204,6 +267,17 @@ public static IServiceCollection AddFluentMigratorCore(
...
+        /// Adds migration runner (except the DB processor specific) services to the specified <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <returns>The updated service collection</returns>
+        [NotNull]
+        [Obsolete("Use AddFluentMigratorSlim() instead for AOT-compatible setups. This method will be removed in a future version.")]
+        public static IServiceCollection AddFluentMigratorCoreWithoutAssemblyLoader(
+            [NotNull] this IServiceCollection services)
+            => AddFluentMigratorSlim(services);
+
```
GerardSmit:
```suggestion
```
This was my old method before I made it `AddFluentMigratorCoreSlim()` (sounds better 😉).
Need to remove this.

GerardSmit (commented):
Self-review; I still need to make some changes. But the beginning is here.

jzabroski:
Superficially, this makes sense, but it's a lot to review, so it might take me a couple weeks before I merge it. @PhenX if you have time to look it over, I would appreciate it.

src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj
```diff
@@ -7,13 +7,16 @@
     <SuppressPseudoWarning Condition="'$(Configuration)' == 'Debug'">true</SuppressPseudoWarning>
   </PropertyGroup>
   <PropertyGroup>
-    <TargetFrameworks>net48;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net48;netstandard2.0;net8.0</TargetFrameworks>
```
PhenX: you should target net 8.0, 9.0 and 10.0 too, everywhere you added it

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -655,60 +782,31 @@ private static IEnumerable<string> GetPathsFromAppDomain()
                 yield return assemblyDirectory;
             }
         }
+#if NET
+        [return: System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicParameterlessConstructor | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicFields | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicProperties)]
+#endif
+        protected delegate Type GetTypeDelegate();
 
-        /// <summary>
```
PhenX, line 659: can you restore the comments as before ?

src/FluentMigrator.Runner.Firebird/FluentMigrator.Runner.Firebird.csproj
```diff
@@ -1,12 +1,15 @@
...
   <PropertyGroup>
-    <TargetFrameworks>net48;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net48;netstandard2.0;net8.0</TargetFrameworks>
     <RootNamespace>FluentMigrator.Runner</RootNamespace>
     <Description>FluentMigrator is a database migration framework for .NET written in C#. The basic idea is that you can create migrations which are simply classes that derive from the Migration base class and have a Migration attribute with a unique version number attached to them. Upon executing FluentMigrator, you tell it which version to migrate to and it will run all necessary migrations in order to bring your database up to that version. In addition to forward migration support, FluentMigrator also supports different ways to execute the migrations along with selective migrations called profiles and executing arbitrary SQL.</Description>
     <PackageTags>fluent;migrator;fluentmigrator;migration;database</PackageTags>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\FluentMigrator.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
PhenX: this change should be put in a single file, imported by all the runners
you could take advantage of this change to also set the TargetFrameworks

PhenX (changes requested):

src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj
jzabroski: Why is it needed?  Does AOT require specific end-target assemblies?

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -73,9 +90,14 @@ protected ReflectionBasedDbFactory(IServiceProvider serviceProvider, params Test
         /// An instance of <see cref="DbProviderFactory"/>.
         /// </returns>
         /// <exception cref="AggregateException">
-        /// Thrown when the factory cannot be created. This exception contains details about all 
+        /// Thrown when the factory cannot be created. This exception contains details about all
         /// the errors encountered during the factory creation process.
         /// </exception>
+#if NET
```
jzabroski, line 99: Basically, if "NET", then assume it is any modern framework we support?

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
jzabroski: Agree, unclear why this was removed. Maybe ChatGPT did it?

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -655,60 +782,31 @@ private static IEnumerable<string> GetPathsFromAppDomain()
...
-        /// Represents an entry used for testing the creation of database provider factories via reflection.
-        /// </summary>
-        /// <remarks>
-        /// This class encapsulates the necessary information, such as the assembly name and the database provider factory type name, 
-        /// to dynamically load and create database provider factories at runtime. It is primarily used internally by 
-        /// <see cref="ReflectionBasedDbFactory"/> and its derived classes.
-        /// </remarks>
         protected class TestEntry
         {
-            /// <summary>
```
jzabroski, line 669: Restore

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -655,60 +782,31 @@ private static IEnumerable<string> GetPathsFromAppDomain()
...
-                [NotNull] string dbProviderFactoryTypeName)
+                [NotNull] string dbProviderFactoryTypeName,
+                [CanBeNull] GetTypeDelegate typeFactory)
             {
                 AssemblyName = assemblyName;
                 DBProviderFactoryTypeName = dbProviderFactoryTypeName;
+                TypeFactory = typeFactory;
             }
 
-            /// <summary>
```
jzabroski, line 690: Restore

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -655,60 +782,31 @@ private static IEnumerable<string> GetPathsFromAppDomain()
...
-            /// </summary>
-            /// <remarks>
-            /// This property provides the assembly name required to dynamically load the database provider factory 
-            /// during runtime. It is used internally by the <see cref="ReflectionBasedDbFactory"/> to locate and 
-            /// instantiate the appropriate factory.
-            /// </remarks>
             [NotNull]
             public string AssemblyName { get; }
 
-            /// <summary>
```
jzabroski, line 701: Restore

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
```diff
@@ -655,60 +782,31 @@ private static IEnumerable<string> GetPathsFromAppDomain()
...
-            /// This property specifies the type name of the database provider factory that is used to create 
-            /// database connections via reflection. It is primarily utilized internally by the 
-            /// <see cref="ReflectionBasedDbFactory"/> class to dynamically load and instantiate the appropriate 
-            /// database provider factory at runtime.
-            /// </remarks>
             [NotNull]
             public string DBProviderFactoryTypeName { get; }
+
+            [CanBeNull]
+            public GetTypeDelegate TypeFactory { get; }
```
jzabroski, line 860: Document

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
```diff
@@ -49,42 +51,98 @@ public static class FluentMigratorServiceCollectionExtensions
...
+#endif
         public static IServiceCollection AddFluentMigratorCore(
             [NotNull] this IServiceCollection services)
         {
             if (services == null)
                 throw new ArgumentNullException(nameof(services));
 
+#pragma warning disable IL2026 // Assembly scanning; callers are warned by [RequiresUnreferencedCode] above
             services
-                // Add support for options
```
jzabroski, line 60: Why did you remove logging and options here? This seems like a major breaking change, and I don't understand it because AddFluentMigratorSlim supports it.

src/FluentMigrator.Runner.Core/MaintenanceLoader.cs
```diff
@@ -39,38 +39,6 @@ public class MaintenanceLoader : IMaintenanceLoader
...
-        /// <param name="conventions">
-        /// The conventions used to identify and process migrations and maintenance stages.
-        /// </param>
-        /// <param name="serviceProvider">
-        /// The service provider used to resolve dependencies for migration and maintenance instances.
-        /// </param>
-        /// <exception cref="ArgumentNullException">
-        /// Thrown if any of the required parameters are <c>null</c>.
-        /// </exception>
-        public MaintenanceLoader(
```
jzabroski, line 61: This is a breaking change - is it possible to keep this and still support AOT? As a best practice, we mark obsolete code as obsolete before breaking it, so people have a chance to upgrade and change.

src/FluentMigrator.Runner.Db2/Processors/Db2/iSeries/Db2ISeriesDbFactory.cs
```diff
@@ -33,7 +33,7 @@ public class Db2ISeriesDbFactory : ReflectionBasedDbFactory
     {
         private static readonly TestEntry[] _testEntries =
         {
-            new TestEntry("Net.IBM.Data.DB2.iSeries", "Net.IBM.Data.DB2.iSeries.iDB2Factory"),
+            new TestEntry("Net.IBM.Data.DB2.iSeries", "Net.IBM.Data.DB2.iSeries.iDB2Factory", () => Type.GetType("Net.IBM.Data.DB2.iSeries.iDB2Factory, Net.IBM.Data.DB2.iSeries")),
```
jzabroski, line 36: Did you test this? What if Type.GetType returns null?

src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj
GerardSmit: No it's not needed. The attributes are used for trimming, which are available in a lower TFM. `RequiresUnreferencedCodeAttribute` is added in .NET 5 so we could target everything after that.

I chose .NET 8.0 because that's the lowest TFM that is currently supported.

src/FluentMigrator.Runner.Core/Processors/ReflectionBasedDbFactory.cs
GerardSmit: Correct, thanks to the .NET naming change 😅

Before .NET 5 it was `#if NET_CORE` and now it's `#if NET`.

src/FluentMigrator.Runner.Db2/FluentMigrator.Runner.Db2.csproj
```diff
@@ -1,12 +1,15 @@
...
-    <TargetFrameworks>net48;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net48;netstandard2.0;net8.0</TargetFrameworks>
     <RootNamespace>FluentMigrator.Runner</RootNamespace>
     <Description>FluentMigrator is a database migration framework for .NET written in C#. The basic idea is that you can create migrations which are simply classes that derive from the Migration base class and have a Migration attribute with a unique version number attached to them. Upon executing FluentMigrator, you tell it which version to migrate to and it will run all necessary migrations in order to bring your database up to that version. In addition to forward migration support, FluentMigrator also supports different ways to execute the migrations along with selective migrations called profiles and executing arbitrary SQL.</Description>
     <PackageTags>fluent;migrator;fluentmigrator;migration;database</PackageTags>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\FluentMigrator.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <IsTrimmable>true</IsTrimmable>
```
jzabroski: @PhenX I'm not sure this is the best way to do this, but not sure what feedback to give @GerardSmit on it - at least not yet.

One thing we may need to decide is whether to verify transitive references via VerifyReferenceTrimCompatibility MSBuild property https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#verify-referenced-assemblies-are-trim-compatible

The other thing is whether unit tests consume trimmed assemblies or not, as that would trivially satisfy Microsoft best practices to "[Creating a trimming test app that uses the library and enabling trimming for the test app. It's not necessary to reference all the APIs in the library."](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#enable-library-trim-warnings:~:text=Creating%20a%20trimming%20test%20app%20that%20uses%20the%20library%20and%20enabling%20trimming%20for%20the%20test%20app.%20It%27s%20not%20necessary%20to%20reference%20all%20the%20APIs%20in%20the%20library.)

src/FluentMigrator.Runner.Firebird/FluentMigrator.Runner.Firebird.csproj
jzabroski: See this comment https://github.com/fluentmigrator/fluentmigrator/pull/2252/changes/BASE..d5d31c1071eb9334bef8a2e42680d4e41f4b5dd6#r2843276922

src/FluentMigrator.Runner.Db2/Processors/Db2/iSeries/Db2ISeriesDbFactory.cs
GerardSmit: There is a null check here: https://github.com/fluentmigrator/fluentmigrator/pull/2252/changes#diff-b00a3e2a8a1a0f447587cdea2ad4d2c8b4cb6ccc039de03f94720b6d10822c9eR230-R236

If it's null, it falls back to the original code.

src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj
jzabroski: no change needed. Just getting familiar with AOT, thanks

jzabroski: well, maybe we should put a note on why we target net8.0 as I would forget myself a year from now

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
GerardSmit: It's moved down to `AddFluentMigratorSlim` (which we call in this method) so there isn't duplicated code.

src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
```diff
@@ -54,10 +54,19 @@ private DefaultMigrationRunnerConventions()
         /// <inheritdoc />
         public Func<IMigration, IMigrationInfo> GetMigrationInfoForMigration => GetMigrationInfoForMigrationImpl;
 
+#if NET
+        // TODO(AOT): We should consider making a delegate with DynamicallyAccessedMembers annotations to avoid the need for these warnings,
+        // however this would be a breaking change to the interface, so for now we will just suppress the warnings on the implementations.
```
GerardSmit: This was a remaining warning from the compiler, which we cannot fix because it would be a breaking change. 

[TypeHasTagsImpl](https://github.com/fluentmigrator/fluentmigrator/pull/2252/changes#diff-28a577aca64a09ee268b923f8e055c92bc85b1a9ff5089e7b96879f7c9e7cde8R117) (the method) has the attribute, but because we make it a Func the compiler cannot see the attribute when TypeHasTags  is called.

TypeHasMatchingTags  has the same issue.

src/FluentMigrator.Abstractions/FluentMigrator.Abstractions.csproj
GerardSmit: Moved to TrimAot.props: https://github.com/fluentmigrator/fluentmigrator/pull/2252/changes#diff-5700fc3d75a3b6f8af2314c505a7af9243805dea9f9a74134953d8aa5101dd4bR3-R7

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
jzabroski: But AddFluentMigratorCore is the existing public API. If the goal is to expose a trimming compatible API, shouldn't we use the attribute on Core to say use Slim (i would prefer CoreTrimmable, but open to others here and view the name as an open handed matter)

GerardSmit: Which attribute?

For the others reading this; I decided on `AddFluentMigratorSlim ` because ASP.NET Core has [WebApplication.CreateSlimBuilder](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.webapplication.createslimbuilder), which was made for trimming in mind.

jzabroski: Thanks, that is helpful context. I asked AI about that naming convention, and it notes CreateSlimBuilder internally calls AddRoutingCore instead of AddRouting, and there was some discussion about calling it AddRoutingSlim. I asked it to find other projects that use Slim, but it said none do.

Grok thinks the "official Microsoft pattern" is to use `Add***Core` for trimmable versions, and `Add***` for non-trimmable versions.

This poses an interesting conundrum given we have allocated the name AddFluentMigratorCore years ago as part of moving to ".NET Core" (now just .NET).

One option is to rename AddFluentMigratorCore, but with AI, most LLMs now "know" that phrase either directly or indirectly through late interaction models like ngram searches of GitHub. AddFluentMigratorSlim is probably the best bet. - Maybe we should just do it now and accept that we will temporarily break LLM understanding of the code base for this one API call.

jzabroski: As best as I can tell, "`Add***Core`" is meant for minimal registration without configuring "additional options." From this lens, perhaps reflection-based resolution is not Core long-term, and we should move that stuff out to optional. @fubar-coder Do you have an opinion here?

GerardSmit: Correct, currently `Add***Core` does assembly scanning, so that's why I made a new method. Otherwise it was a breaking change.

---

Other idea:

What I did in my project is source generate assembly attributes: https://github.com/WebFormsCore/WebFormsCore/blob/0ba0242dc2a8a57b9aa926279a6426c044d5136c/src/WebFormsCore.Parser/Templates/designer.scriban#L169

Which I then fetch by going through the assembly attributes: https://github.com/WebFormsCore/WebFormsCore/blob/0ba0242dc2a8a57b9aa926279a6426c044d5136c/src/WebFormsCore/Internal/AppDomainControlTypeProvider.cs#L68

Because I added `DynamicallyAccessedMembers` to the attribute, and it's in the assembly attributes, the type + constructors don't get trimmed: https://github.com/WebFormsCore/WebFormsCore/blob/0ba0242dc2a8a57b9aa926279a6426c044d5136c/src/WebFormsCore/Attributes/RootNamespaceAttribute.cs#L12-L21

This could be another route; use a source generate (or MSBuild task) to register all migrations in assembly attributes, then we can scan the in `AddFluentMigrator` for the migrations.

**Upside:** 
- No new API.
- Consumer of FluentMigrator doesn't have to do anything. Just compile to AOT.

**Downside:** 
- Complexity; source generators/MSBuild tasks will make it more complex.
- The libraries and app has to have the source generator running (aka they have to recompile with the new version); otherwise the migration won't get applied. We could make an analyzer for this (check if a assembly has migrations but no assemblies, but this is more complexity...) / or scan the libraries as well to make assembly attributes for external migrations as well.

src/FluentMigrator.Runner.Db2/FluentMigrator.Runner.Db2.csproj
jzabroski: @GerardSmit Getting back to review. This should use the MSBuild intrinsic [MSBuild]::IsTargetFrameworkCompatible

To my pleasant surprise, it's actually documented as a best practice by Microsoft, here: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#target-framework-requirements - How serendipitous! :)

src/FluentMigrator.Runner.Core/Infrastructure/DefaultMigrationConventions.cs
jzabroski: The breaking change would be to:
* Obsolete `public Func<Type, bool> TypeHasTags`; replace it with:
   ```c#
   public virtual bool TypeHasTags(
   #if NET
       [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
   #endif
       Type type)
    ```
* Obsolete `public Func<Type, IEnumerable<string>, bool> TypeHasMatchingTags`; replace it with:
   ```c#
   public virtual bool TypeHasMatchingTags(
   #if NET
       [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
   #endif
       Type type,
       IEnumerable<string> tagsToMatch)
   ```

I agree, this is a significant breaking change, as this is probably one of the more frequently customized classes in FluentMigrator. We can publish a version of FM in 8.x.y version that marks these as obsolete and adds the above replacement interfaces, so that people have a transition clue. Then we would merge this PR to create FM 9.0.0 release.

TrimAot.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <TargetFrameworks>$(TargetFrameworks);net8.0</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
jzabroski: Use MSBuild intrinsic function [MSBuild]::IsTargetFrameworkCompatible rather than direct equality comparison

TrimAot.props
```diff
@@ -0,0 +1,8 @@
...
+    <IsTrimmable>true</IsTrimmable>
```
jzabroski, line 10: @GerardSmit I think we should verify references are AOT-compatible as well. What do you think?

See: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#verify-referenced-assemblies-are-aot-compatible

jzabroski:
@GerardSmit See my latest comments. I think this is close to complete. My main ask is that we think about a way to signal ReflectionBasedDbFactory should no longer be used, and add a separate PR to address the dynamically accessed members in DefaultMigrationConvention.cs so that we can ship a version of FluentMigrator that warns about those interfaces going away and being replaced by virtual interfaces. Let me know if I misunderstood the concern about DefaultMigrationConvention, though. I just think if the simple fix is to stage deprecation of features with something AOT compatible, that is preferred vs. building a lot of extra machinery to maintain long-term.

src/FluentMigrator.Abstractions/Validation/DefaultMigrationExpressionValidator.cs
jzabroski: Add it to the PR we should do for 8.x.y before merging.

jzabroski:
Trying to help push blockers here forward. Assigned https://github.com/fluentmigrator/fluentmigrator/issues/2256 to Copilot.

GerardSmit:
Sorry, hectic week 🙏

I did some work on it. I got the validations working together with https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/ILLink-files.md and made some tests with TUnit (so I'm 100% sure validations work).

cristiano-linvix:
Hi guys, I hope you're all doing well!
Thank you for the excellent work!

Is there a plan to release this?

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
```diff
@@ -129,10 +129,13 @@ public static IServiceCollection AddFluentMigratorSlim(
...
                 .TryAddSingleton<IAssemblySource, EmptyAssemblySource>();
 
+            services
+                .TryAddSingleton<ITypeSource>(new ArrayTypeSource(Array.Empty<Type>()));
+
 #pragma warning disable IL2026
             services
                 // Configure the loader for migrations that should be executed during maintenance steps
-                .TryAddSingleton<IMaintenanceLoader, MaintenanceLoader>();
+                .TryAddSingleton<IMaintenanceLoader>(sp => ActivatorUtilities.CreateInstance<MaintenanceLoader>(sp));
```
GerardSmit, line 138: I hate that I had to do this, but because of backwards capability on the public constructor (`[NotNull] IAssemblySource source`) I was getting an ambiguous constructor exception. So there was no other way around.

I could remove the constructor with `[NotNull] ITypeSource source`, then check if there is a `ITypeSource` registered with `IServiceProvider` in the existing constructor, and use that one? Then we don't have to do this "hack".

GerardSmit:
Most of the work is done; it works but there is still one thing I'm reconsidering (look at my last comment).

jzabroski:
My general policy is to try to avoid responding to "updates?" questions, as they just waste time giving status updates versus actually doing work reviewing PRs, testing changes, etc. I appreciate the interest, though, and do intend to incorporate this PR soon.

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
jzabroski: Can you use [ActivatorUtilitiesConstructorAttribute](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilitiesconstructorattribute?view=net-10.0-pp)?

GerardSmit: Yes, that is what the code is doing now. `ActivatorUtilitiesConstructorAttribute` is only for `ActivatorUtilities.CreateInstance`

jzabroski:
@GerardSmit I do want to merge this PR but I am stuck on #2257 and the best path for creating a migration bridge to minimize impact to users. Let me know your thoughts.

jzabroski:
@GerardSmit I read about platform-specific CLI Tool support that was added as of the .NET 10 SDK - https://learn.microsoft.com/en-us/dotnet/core/tools/rid-specific-tools#:~:text=Package%20.,of%20the%20following%20MSBuild%20properties: - Thanks to your changes, we can ship a thinner FluentMigrator.DotNet.CLI tool! Let me know if you're interested in incorporating that. I also asked Copilot about it here, as it's a blocker for adopting Hana in FluentMigrator.DotNet.CLI: https://github.com/fluentmigrator/fluentmigrator/issues/2276

GerardSmit:
I've commented in https://github.com/fluentmigrator/fluentmigrator/pull/2257#pullrequestreview-4215262428

--- 

About the FluentMigrator.DotNet.CLI tool; personally I'm not using it but if I understand it correctly; if this PR is okay, then we can just add RuntimeIdentifiers and set TrimMode and it should be good?

src/FluentMigrator.Runner.Core/FluentMigratorServiceCollectionExtensions.cs
jzabroski: We will use AddFluentMigratorSlim. Sorry for taking so long to review all this. Personal time for hobbies flies when you have two little kids.

jzabroski:
@GerardSmit Merging this today. I know I made you wait because of the backward compatibility concerns. I think this one is just going to have to have a bright clear indication in the release notes. I try really hard not to break users existing code bases. Hope you understand.

GerardSmit:
No worries, completely understandable.

I was wondering, maybe enabling https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/overview can help with this?

jzabroski:
create a issue. im always for best practices. I think once Hana libs are gone from source control, Claire Novotny's idempotent builds approach will be easier too.

PR #2267, opened 2026-04-09
Check for existing Transaction when validating connection in Runner

concept-hf:
## Problem

`ValidateConnection()` in `MigrationRunner.cs` wants to open and immediately close a transaction to validate the connection string.

This will force a preexisting Transaction to be rolled back, because the `BeginTransaction` will not open a new one, but the `RollbackTransaction` will roll back the existing one.

This causes the Processor later to create it's own Transaction per migration disregarding (pre-existing) transaction scopes.

## Fix

Introduce a `HasTransaction` helper method on processor, and check it on `ValidatingConenction()`.

### DIFF
--- a/src/FluentMigrator.Abstractions/IMigrationProcessor.cs
+++ b/src/FluentMigrator.Abstractions/IMigrationProcessor.cs
@@ -85,6 +85,12 @@ namespace FluentMigrator
         void RollbackTransaction();
 
         /// <summary>
+        /// Checks for active transaction
+        /// </summary>
+        /// <returns><c>true</c> when a transaction exists</returns>
+        bool HasTransaction();
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


concept-hf:
Hello,

@jzabroski could you please retrigger the build? It seems to be an intermittent problem of the build system...

Also any chance for review?
Imo this is a pretty easy fix for a nasty bug...

jzabroski:
I am on my phone. Was this caused by a recent change?

concept-hf:
The issue is caused by this commit: seems to be a copilot change
https://github.com/concept-hf/fluentmigrator/commit/e2636448f65f353ee2aef1dbd5fa03073de5fe46

7 month ago. is it recent?
I've only caught the bug recently.

jzabroski (approved):

Issue #1981, opened 2025-02-27
Support for Dynamic Passwords


Merlin-Taylor:
**Is your feature request related to a problem? Please describe.**

> In some cloud scenarios, logging into PostgreSQL is done with an auth token that is rotated every time interval (e.g. one hour).
> — [Npgsql documentation § Auth token rotation and dynamic password](https://www.npgsql.org/doc/security.html?tabs=tabid-1#auth-token-rotation-and-dynamic-password)

The Postgres migration runner [uses `Npgsql.NpgsqlFactory`](https://github.com/fluentmigrator/fluentmigrator/blob/471cdb89f09cf06434108781cc50e85aee60824f/src/FluentMigrator.Runner.Postgres/Processors/Postgres/PostgresDbFactory.cs#L27). [`Npgsql.NpgsqlFactory`](https://github.com/npgsql/npgsql/blob/main/src/Npgsql/NpgsqlFactory.cs/) has methods to create various Npgsql objects but it does not pass the arguments necessary to use dynamic passwords to any of them.

**Describe the solution you'd like**

A new extension method, `UsePasswordProvider` in [MigrationRunnerBuilderExtensions.cs](https://github.com/fluentmigrator/fluentmigrator/blob/main/src/FluentMigrator.Runner.Core/MigrationRunnerBuilderExtensions.cs), similar to [`Npgsql.NpgsqlDataSourceBuilder.UsePasswordProvider`](https://github.com/npgsql/npgsql/blob/3146bcda307d49ecd21d9e73c87a142dd1ee8cba/src/Npgsql/NpgsqlDataSourceBuilder.cs#L359).

**Additional context**

I am working with AWS RDS IAM Database Authentication as documented in [Connecting to your DB instance using IAM authentication](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html).

**Is this a feature you'd like to submit a PR for?**

Yes.

jzabroski:
Hey @Merlin-Taylor - That sounds awesome. We have had similar requests for SQL Server in the past. See: https://github.com/fluentmigrator/fluentmigrator/issues/908

1. SQLite supports SQLCipher which doesn't use a normal connection string to connect to the database. Rather, the user must issue a [PRAGMA KEY = 'passphrase'](https://github.com/sqlcipher/sqlcipher#encrypting-a-database) command to decrypt the database.
[...]

PR #2268, opened 2026-04-21
Custom Connection factory

Sherif-Ahmed:
This PR adds support for creating migration database connections through a configurable connection factory.

The main goal is to allow applications to provide fully configured `IDbConnection` instances instead of relying only on a connection string. This is useful for scenarios where connection setup requires runtime logic or provider-specific configuration, such as Azure SQL access tokens, PostgreSQL dynamic passwords, tenant-specific connection resolution, or external secret stores.

## Changes

- Added `IMigrationConnectionFactory`.
- Added a default connection-string based implementation to preserve existing behavior.
- Added `WithConnectionFactory(...)` to `IMigrationRunnerBuilder`.
- Updated `GenericProcessorBase` to create connections through `IMigrationConnectionFactory`.
- Updated processor selection logic so factory-based connections are not treated as connectionless.
- Updated version-loader selection logic to use connection factory availability instead of checking only the connection string.
- Updated provider processors to use `IMigrationConnectionFactory`.
- Added tests for:
  - Existing `.WithGlobalConnectionString(...)` behavior.
  - Factory-based connection creation when no connection string is configured.
  - Opening factory-created connections.
  - Disposing factory-created connections.
  - Provider execution through `WithConnectionFactory(...)`.

## Example

```csharp
services
    .AddFluentMigratorCore()
    .ConfigureRunner(runner => runner
        .AddSqlServer()
        .WithConnectionFactory(serviceProvider =>
        {
            var configuration = serviceProvider.GetRequiredService<IConfiguration>();

            return new SqlConnection(
                configuration.GetConnectionString("DefaultConnection"));
        })
        .ScanIn(typeof(SomeMigration).Assembly).For.Migrations());
```

Related to #1981 and PR #2002.

```csharp
services.AddSingleton(sp =>
{
    var builder = new NpgsqlDataSourceBuilder(connectionString);

    builder.UsePasswordProvider(
        passwordProvider: settings => GetPassword(),
        passwordProviderAsync: (settings, cancellationToken) => GetPasswordAsync(cancellationToken));

    return builder.Build();
});

services
    .AddFluentMigratorCore()
    .ConfigureRunner(runner => runner
        .AddPostgres()
        .WithConnectionFactory(sp =>
            sp.GetRequiredService<NpgsqlDataSource>().CreateConnection())
        .ScanIn(typeof(SomeMigration).Assembly).For.Migrations());
```

**We can also support externally-owned connections later**


jzabroski (changes requested):
What's the reason for not providing a backward compatible constructor and just using IConnectionStringAccessor with a new MigrationConnectionFactory assigned in the constructor so that people have a clear warning that code is becoming obsolete?

Sherif-Ahmed:
The main reason is to avoid constructor ambiguity in DI, but I agree that keeping the old ctor is better for backward compatibility and gives users a clear obsolete warning instead of a breaking change. 

I can rework this so the existing `IConnectionStringAccessor` constructor is kept and marked obsolete, delegating to the connection-string based factory path. The new `IMigrationConnectionFactory` path will be used by the service registrations / internal construction path, while direct callers of the old constructor will get an obsolete warning.

Just to give some context: I need this functionality quite urgently in my current project, which is why I started the implementation. The ability to provide a fully configured connection through `WithConnectionFactory(...)` would solve an immediate issue for me, but I still want to make sure the design is aligned with FluentMigrator before going further.

I’m happy to update the PR based on your feedback.

jzabroski:
The main thing I never settled is whether we should use the existing DbDataSource.  https://github.com/dotnet/runtime/issues/64812 If you look, someone contributed a related (still open) PR but I never pulled it in because I never spent the brain cycles to compare and contrast the two approaches, and there wasn't a ton of demands. But I absolutely am open to adding this, possibly this week, as I'm on a stay at home vacation right now.

jzabroski:
See #2002

Sherif-Ahmed:
Good point. I had not fully considered DbDataSource.

This PR uses IMigrationConnectionFactory mainly because it works for all current provider targets and allows the runner to receive a ready IDbConnection regardless of whether it came from a connection string, token callback, NpgsqlDataSource, SqlConnection configuration, tenant resolution, etc.

But I agree DbDataSource is probably the more idiomatic ADO.NET direction where it is available. It represents a configured data source/pool and is intended to hand out ready connections, while DbProviderFactory is more of a provider object factory. The runtime proposal explicitly argues for keeping DbProviderFactory and DbDataSource separate because they represent different concepts.

Maybe the best approach is to support both:

- Keep IMigrationConnectionFactory as the internal abstraction used by processors.
- Add a WithConnectionFactory(...) overload for simple/custom connection creation.
- Add a WithDataSource(...) overload when callers already have a DbDataSource.
- The DbDataSource overload can adapt to IMigrationConnectionFactory internally by calling dataSource.CreateConnection().

Something like:
```
public static IMigrationRunnerBuilder WithDataSource(
    this IMigrationRunnerBuilder builder,
    Func<IServiceProvider, DbDataSource> dataSourceFactory)
{
    return builder.WithConnectionFactory((sp, _) =>
    {
        var dataSource = dataSourceFactory(sp);
        return dataSource.CreateConnection();
    });
}
```

Regarding #2002, I checked it. My understanding is that it solves the PostgreSQL dynamic password case by allowing a custom DbProviderFactory to be supplied for Postgres only. It looks intentionally minimal and delegates the dynamic password behavior to Npgsql. I also noticed the note there that DbProviderFactory may be reasonable internally but may not be the best public API surface.

That is why I was thinking the current PR could be a more general version of the same idea:

- #2002: provider-specific, Postgres-only, custom DbProviderFactory
- this PR #2268: provider-neutral, caller-configured connection creation through IMigrationConnectionFactory / WithConnectionFactory(...)

I agree DbDataSource is worth considering, especially for NpgsqlDataSource and dynamic password scenarios. My preference would be to first support caller-configured/factory-created connections, because it solves the immediate need and also keeps the internal design flexible. Then DbDataSource could be added as a follow-up or additional overload that adapts into the same internal connection-factory abstraction.

Something like:
`WithDataSource(Func<IServiceProvider, DbDataSource> dataSourceFactory)`
could internally call:
`dataSource.CreateConnection()`
and reuse the same IMigrationConnectionFactory path.

My thinking is that the immediate need is support for caller-provided / externally-created connections, because that unblocks cases where the connection has to be configured at runtime before FluentMigrator opens it.

I will bring back the IConnectionStringAccessor' ctors first and mark them as `Obsolete` to give users a clear warning instead of a breaking change.

Sherif-Ahmed:
I addressed the backward-compatibility and DI ambiguity feedback and pushed the updates.

Main changes:
- restored the existing IConnectionStringAccessor constructors and marked them obsolete
- kept IMigrationConnectionFactory as the preferred/internal path
- resolved constructor ambiguity in DI
- aligned the tests with the existing integration test infrastructure

If you get a chance, could you please take another look?

fubar-coder:
@jzabroski I like the idea behind this PR, as it would fix the outstanding issue where you need to configure the - for example - NpgsqlConnection in ways that cannot be expressed in the connection string.

jzabroski:
@fubar-coder does that mean you reviewed it and approve

jzabroski:
@Sherif-Ahmed can you add this to the PR

```c#
public static IMigrationRunnerBuilder WithDataSource(
    this IMigrationRunnerBuilder builder,
    Func<IServiceProvider, DbDataSource> dataSourceFactory)
{
    return builder.WithConnectionFactory((sp, _) =>
    {
        var dataSource = dataSourceFactory(sp);
        return dataSource.CreateConnection();
    });1
}
```

fubar-coder:
> @fubar-coder does that mean you reviewed it and approve

Sorry, it was just meant as an opinion and a reason for why this PR is useful.

Sherif-Ahmed:
@jzabroski I updated the PR accordingly.

src/FluentMigrator.Runner.Core/MigrationRunnerBuilderExtensions.cs
```diff
@@ -305,5 +308,68 @@ public IMigrationRunnerBuilder All()
...
+                if (connection == null)
+                {
+                    throw new InvalidOperationException("The configured connection factory returned null.");
+                }
+
+                return connection;
+            }
+        }
+
+#if NET7_0_OR_GREATER
```
jzabroski, line 354: Add comment 
// DbDataSource was introduced in .NET 7 and is not supported on .NET Framework.

src/FluentMigrator.Runner.Db2/Processors/Db2/iSeries/Db2ISeriesProcessor.cs
```diff
@@ -62,6 +65,20 @@ public Db2ISeriesProcessor(
             Quoter = quoter;
         }
 
+        /// <inheritdoc />
+        [ActivatorUtilitiesConstructor]
```
jzabroski, line 69: Thanks for making this NativeAOT/Trimmable friendly.

src/FluentMigrator.Runner.Core/Initialization/IMigrationConnectionFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+#endregion
+
+using System.Data;
+using System.Data.Common;
+
+using JetBrains.Annotations;
+
+namespace FluentMigrator.Runner.Initialization
+{
+    [PublicAPI]
```
jzabroski, line 24: Are there no tests that reference this interface? Otherwise, I would not expect Resharper to complain about this being eligible to make private.

test/FluentMigrator.Tests/Integration/ConnectionFactoryTests.cs
```diff
@@ -0,0 +1,153 @@
...
+using FluentMigrator.Runner.Processors.SqlServer;
+using FluentMigrator.Tests.Integration.TestCases;
+
+using Microsoft.Data.SqlClient;
+using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+using MySql.Data.MySqlClient;
+
```
jzabroski: can you remove some of these excess line breaks?

jzabroski:
OK. I appreciate your effort there.

I added some review comments on the code itself; I am still not done reviewing every line but making progress. I clicked submit on my current comments hoping that I won't have anything else to add.


Some last changes requested:

1. We have an ADR for architecture patterns in FluentMigrator, as it relates to DependencyInjection. Can you also update https://github.com/fluentmigrator/fluentmigrator/blob/main/adr/proposed/DependencyInjection.md to add a 13th row, after connection strings? I think the item label should be Connections, Interface = IDbConnection.
    a. One thing DbDataSource does that SqlConnection does not is handle the concept of a [Batch](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbdatasource.createbatch?view=net-10.0); I appreciated your analysis above, although I do wonder if it was generated by Claude I still found it very helpful. I think the main question we should ask if to what extent we may want to eliminate round trips in the future. This does not need to be solved in this PR.
2. It is not clear to me how this feature should interact with the WithGlobalConnectionString API. How do you think it should work? https://github.com/fluentmigrator/fluentmigrator/blob/039466aac6b0a039b01370519b696332996e42f6/src/FluentMigrator.Runner.Core/MigrationRunnerBuilderExtensions.cs#L59-L86
    a. Ideally, add a unit test to pencil in that behavior to avoid regression risk. I see them as orthogonal settings that can be independently combined and you should also be able to "opt in" to the new factory without it disrupting setting WithGlobalConnectionString - WithGlobalConnectionString, if you squint, was our pre-DbDataSource way of handling half of the problem with SqlConnection lacking something like a DbDataSource abstraction. 
4. VitePress docs
    a. Can you update https://github.com/fluentmigrator/fluentmigrator/blob/main/docs-website/providers/postgresql.md with your DynamicPasswords example code? I would put it under the Configuration section, as another sub-section after "Basic Configuration".
    b. Update https://github.com/fluentmigrator/fluentmigrator/blob/main/docs-website/runners/in-process.md to replace `WithGlobalConnectionString` with something more appropriate using `DbDataSource`, and a `> [!NOTE]` markdown note to communicate the preferred approach for .NET 8 onwards is to configure a `DbDataSource`.
       a. Alternatively, maybe WithGlobalConnectionString should be updated to set-up the default connection factory.

jzabroski (changes requested):

src/FluentMigrator.Runner.Db2/Processors/Db2/iSeries/Db2ISeriesProcessor.cs
Sherif-Ahmed: :thumbsup:

src/FluentMigrator.Runner.Core/MigrationRunnerBuilderExtensions.cs
Sherif-Ahmed: Added

src/FluentMigrator.Runner.Core/Initialization/IMigrationConnectionFactory.cs
Sherif-Ahmed: Fixed, I added a direct DI registration test for a custom `IMigrationConnectionFactory` to make the intended public extension point explicit.

test/FluentMigrator.Tests/Integration/ConnectionFactoryTests.cs
Sherif-Ahmed: Fixed

Sherif-Ahmed:
Thanks, that makes sense.

My understanding is that DbDataSource is a better long-term abstraction than only passing an IDbConnection, because it can represent the configured data source/pool and leaves room for future provider features such as batching or reducing round-trips, but for now keeping the scope focused on configurable connection creation, while keeping the API compatible with adding more DbDataSource-based behavior later if you want to move in that direction.

And yes, I used AI assistance to help compare the approaches and to help me express the analysis clearly, especially because English is not my mother tongue :smile:. That said, I reviewed and adapted the conclusions against the actual FluentMigrator design and the ADO.NET APIs before applying anything.

1. I updated the ADR with the new row after connection strings
I can also add a short note under the existing “Connections/connection strings” section to clarify that this new connection abstraction is separate from connection string resolution:
“Database connections are treated as a dependency-injected concept represented by `IDbConnection`. The runner obtains connections through `IMigrationConnectionFactory`, allowing connection creation to be configured independently from connection string resolution.”
a. One small nuance: `DbConnection` also has `CreateBatch()` on newer .NET targets, but I agree the `DbDataSource` point still stands. `DbDataSource` is the better long-term abstraction because it represents the configured data source/pool, not just one connection instance. That leaves room for future provider features such as batching or reducing round-trips.

2. I added tests for the combined cases to prevent regressions.
My initial approach was to move the actual connection creation path away from depending directly on `IConnectionStringAccessor`, so the runner could create connections through `IMigrationConnectionFactory`.
After the backward-compatibility discussion, I think the better behavior is to keep `WithGlobalConnectionString` fully supported and treat it as orthogonal to the new APIs:

- `WithGlobalConnectionString(...)` continues to configure `ProcessorOptions.ConnectionString` / `IConnectionStringAccessor`.
- The default `IMigrationConnectionFactory` implementation uses that configured connection string, preserving existing behavior.
- `WithConnectionFactory(...)` or `WithDataSource(...)` explicitly replace the connection creation path, but they do not clear or invalidate the global connection string.
- If both are configured, the explicit factory/data source is used to create the connection, and the global connection string remains available for that factory if it wants to use it.
So the new APIs are an opt-in replacement for connection creation, not a replacement/removal of the connection string configuration API.
Longer term, whether `WithGlobalConnectionString` should stay as the primary API, become a convenience API over the default factory, or eventually be de-emphasized in favor of `DbDataSource` is more of a design decision for the project.

3. I updated the VitePress docs in two places:
a. Add a PostgreSQL “Dynamic Passwords” subsection under Configuration, after Basic Configuration, showing NpgsqlDataSourceBuilder + WithDataSource(...).
b. Update the in-process runner examples to use a DbDataSource-based setup, with a note that for .NET 8+ the preferred approach is to configure a DbDataSource and pass it through WithDataSource(...).
c. The current design already does this indirectly:
```
WithGlobalConnectionString(...)
    -> ProcessorOptions.ConnectionString
    -> IConnectionStringAccessor
    -> ConnectionStringMigrationConnectionFactory
    -> IDbConnection
```
So `WithGlobalConnectionString` remains the existing configuration API, and the default `IMigrationConnectionFactory` consumes it. That keeps old behavior stable and avoids making call order confusing.

jzabroski:
> 1. One small nuance: `DbConnection` also has `CreateBatch()` on newer .NET targets, but I agree the `DbDataSource` point still stands. `DbDataSource` is the better long-term abstraction because it represents the configured data source/pool, not just one connection instance. That leaves room for future provider features such as batching or reducing round-trips.

I didn't notice that; thanks for pointing that out!

> After the backward-compatibility discussion, I think the better behavior is to keep WithGlobalConnectionString fully supported and treat it as orthogonal to the new APIs

👍 

Thanks for updating everything. I have to fix the GitHub Actions then I will look into merging a bunch of outstanding PRs. There may be some merge conflicts with Gerard Smit's nativeaot/trimmable PR #2252

jzabroski:
This will be included shortly via Copilot. It used a combination of your ideas and the other PR.

jzabroski:
Completed via #2320

PR #2269, opened 2026-04-24
added separate OracleManagedGenerator and test case

bdariusb:
Issue:
FM CLI commands using **OracleManaged** processor could not find the **OracleManaged** Id
For example, a command
`dotnet fm list migrations -p OracleManaged -a "MyAssembly.dll"`
throws:
`System.InvalidOperationException: A migration generator with the ID OracleManaged couldn't be found. Available generators are: Db2, Db2ISeries, Oracle, Oracle12C, Firebird, Hana, MySql4, MySql5, MySql8, Oracle, Oracle12C, Oracle, Oracle12C, Postgres15_0, Postgres92, Postgres10_0, Postgres11_0, Postgres15_0, Redshift, Snowflake, SQLite, SqlServer2016, SqlServer2000, SqlServer2005, SqlServer2008, SqlServer2012, SqlServer2014, SqlServer2016`
Fix:
register own generator with `GeneratorIdConstants.OracleManaged` key

### DIFF
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/IOracleManagedGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/IOracleManagedGenerator.cs
@@ -1,0 +1,27 @@
+#region License
+// Copyright (c) 2026, Fluent Migrator Project
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


jzabroski:
@bdariusb Was this functionality that broke for you? Im not sure it understood the issue.

jzabroski:
@bdariusb I think this is not 100% the right fix.

* OracleManagedProcessor (which is what actually talks to the database engine driver library, not the generator) needs a DatabaseType override to add this OracleManaged alias.
* You still need your part of the fix, as we likely do not want "OracleManaged" to be tied to the existing Generator, even if there is nothing special about this one and it is just a pass-through - eventually the "Managed" suffix may go away (it's up to Oracle's preference how they bundle drivers, independent of the database engine technology).

bdariusb:
> * OracleManagedProcessor (which is what actually talks to the database engine driver library, not the generator) needs a DatabaseType override to add this OracleManaged alias.

DatabaseType is already overriden in `OracleManagedProcessor` constructor
or do you mean `DatabaseTypeAliases` which is Oracle in the base class (which is actually wrong because Oracle != OracleManaged)?

jzabroski:
No, you're right. I missed that.

jzabroski (approved):

PR #2292, opened 2026-06-08
Update indexes.md

toondedo:
The `OnColumn` has `ICreateIndexColumnOptionsSyntax` as return type instead of `ICreateIndexOnColumnsSyntax`. Index column options must follow after specifying a column before the next column can be added to the index. The documentation does not reflect this.


jzabroski:
You're right; the examples are wrong (outdated?)

PR #2303, opened 2026-07-09
Handle partial assembly type load failures

### LINKED ISSUE #2258: FluentMigrator throws DLL not found exception on .NET 10 projects using the shared AspNetCore framework.
**Describe the bug**
When migrating a solution to .NET 10 from .NET 10, our migrations started to error out with a message similar to the following:
```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly
'Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions, Version=10.0.0.0, Culture=neutral,
PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
   at System.Reflection.RuntimeAssembly.GetExportedTypes()
   at FluentMigrator.Runner.Initialization.AssemblySourceMigrationRunnerConventionsAccessor.<>c.<.ctor>b__1_1(Assembly a)
```

Upon some investigation with Claude's help, this appears to be caused by FluentMigrator loading the assemblies off of the migration projects that were copied to its output directory. Our migration projects contain also some code which uses the framework libraries. Since previously in .NET 8 assemblies such as Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions would be copied to the output directories of the projects that included them, this was not a problem and the application works fine. However, with the advent of .NET 10, this scenario changed, because e.g. those HealthChecks.Abstractions, instead of being provided as standalone nugets, are now provided by the Microsoft.AspNetCore.App shared framework and thus no longer copied to the output path. FluentMigrator cannot find it and, in its assembly loading steps, will throw an error. This has caused us to have to split our Data projects into two, one of them having only the migrations and no framework code, as a workaround which we would very much want to do away with as soon as possible. 

**To Reproduce**
Full repro and more detailed explanation (thanks Claude) are provided here: https://github.com/scarpart/fluentmigrator-repro

**Expected behavior**
No exception should be thrown by simply switching to .NET 10. Projects should work in-place and not require change on the part of FluentMigrator. 

.NET SDK: 8.0 and 10.0
FluentMigrator.DotNet.Cli: 8.0.1
FluentMigrator (library): 8.0.1
OS: macOS / Linux
DOTNET_ROLL_FORWARD=LatestMajor (required for .NET 10 since dotnet-fm targets net8.0)

**Additional context**
This is a real bug caught in a real project from a real company, which means it is likely to affect many others in the near future as people start migrating to .NET 10.

hopkienne:
Fixes #2258

## Summary

This updates assembly type scanning so FluentMigrator can continue using loadable exported types when `Assembly.GetExportedTypes()` throws `ReflectionTypeLoadException`.

This can happen when a migration assembly contains public types that depend on assemblies provided by a shared framework, such as .NET 10 / `Microsoft.AspNetCore.App`, and those dependency DLLs are not copied beside the migration assembly.

## Changes

- Added `GetLoadableExportedTypes()` helper.
- Updated assembly-backed migration/convention/version metadata scanners to use loadable types from `ReflectionTypeLoadException.Types`.
- Added regression tests for:
  - `AssemblyTypeSource`
  - `AssemblyMigrationSourceItem`
  - `AssemblySourceMigrationRunnerConventionsAccessor`

## Verification

- `dotnet test test\FluentMigrator.Tests\FluentMigrator.Tests.csproj --configuration Debug --no-restore --filter "FullyQualifiedName~AssemblyTypeLoadingTests" --verbosity minimal`
  - Passed: 3
- `dotnet build src\FluentMigrator.Runner.Core\FluentMigrator.Runner.Core.csproj --configuration Debug --no-restore --verbosity minimal`
  - Succeeded
- `dotnet test test\FluentMigrator.Tests\FluentMigrator.Tests.csproj --configuration Debug --no-restore --verbosity minimal`
  - Passed: 5157, Skipped: 909

### DIFF
--- a/src/FluentMigrator.Runner.Core/Initialization/AssemblyExtensions.cs
+++ b/src/FluentMigrator.Runner.Core/Initialization/AssemblyExtensions.cs
@@ -1,0 +1,41 @@
+#region License
+// Copyright (c) 2007-2024, Fluent Migrator Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace FluentMigrator.Runner.Initialization
+{
+    internal static class AssemblyExtensions
+    {
+#if NET
+        [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("This method gets the exported types from assemblies, which may not be preserved in trimmed applications.")]
+#endif
+        public static IEnumerable<Type> GetLoadableExportedTypes(this Assembly assembly)
+        {
+            try
+            {
+                return assembly.GetExportedTypes();
+            }
+            catch (ReflectionTypeLoadException ex)
+            {
+                return ex.Types.Where(t => t != null);
+            }
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


jzabroski:
Hi, I responded to the original issue - apologies I did not see this issue sooner. The fix was obvious to me and is actually documented in the sample, but I am so sorry you spent time on this - it trips up a lot of .NET Developers and I personally only know the solution because I previously maintained RazorLight, so I have a ton of experience loading "side car assemblies" and bypassing the deps.json generation process. We still recommend customers use In-Process Runners as MSBuild will generate the deps.json and completely footgun-proof this for you.

PR #2304, opened 2026-07-10
Prevent unsafe conditional auto-reversal

hopkienne:
## Summary

- Reject automatic reversal of conditional table creation and conditional column changes.
- Add unit and SQLite integration regression coverage that preserves pre-existing data and migration state.
- Document conditional DDL as forward-only when ownership cannot be established.

## Root cause

`CreateTableExpression.Reverse()` and `CreateColumnExpression.Reverse()` predated the conditional flags and always generated destructive reverse expressions. When the database skipped `CREATE TABLE IF NOT EXISTS` or `ALTER TABLE IF EXISTS`, rollback could still drop a pre-existing object or fail against a missing table.

Conditional expressions now delegate to the established base `Reverse()` implementation, which throws `NotSupportedException`. This requires an explicit or forward-only migration instead of silently producing an unsafe rollback.

## Verification

- Focused regression tests: 3 passed
- Release build succeeded for `net48`, `netstandard2.0`, and `net8.0`
- Full test suite: 5,157 passed, 909 skipped, 0 failed
- `git diff --check`: clean

### DIFF
--- a/docs-website/migration-types/auto-reversing.md
+++ b/docs-website/migration-types/auto-reversing.md
@@ -353,6 +353,7 @@ public class CustomReversalLogic : AutoReversingMigration
 2. **Don't assume auto-reversal works for custom SQL** - it only applies to fluent API operations
 3. **Don't forget to test rollback scenarios** even with auto-reversible migrations
 4. **Don't mix auto-reversal with partial custom Down() methods** without understanding the interaction
+5. **Don't use conditional DDL such as `IfNotExists()` or `IfExists()`** because FluentMigrator cannot know whether the database applied or skipped the operation
 
 ## Complex Scenarios
 
--- a/docs-website/operations/alter-tables.md
+++ b/docs-website/operations/alter-tables.md
@@ -77,7 +77,7 @@ public class RemoveColumns : Migration
 Use `IfExists()` to skip an `ALTER TABLE` statement when the target table doesn't exist. This avoids the boilerplate of manually checking `Schema.Table(...).Exists()` before altering a table, and is especially useful when a migration touches an optional or database-specific table.
 
 ```csharp
-public class AddColumnIfTableExists : Migration
+public class AddColumnIfTableExists : ForwardOnlyMigration
 {
     public override void Up()
     {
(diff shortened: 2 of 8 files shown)


millonbigs-cpu (approved):

jzabroski:
Hi - It seems like the correct behavior of reversible migrations depends on whether (a) you are using a ForwardOnlyMigration or bi-direction `Migration` (b) ability to memoize the state of the migration and what conditionals flowed through at the time of the migration running.

I acknowledge your fix is likely a clear improvement, but wanted to get your thoughts on the above first.

hopkienne:
@jzabroski I agree that the lifecycle distinction matters. This change is deliberately limited to generated reversal: Reverse() is invoked automatically by AutoReversingMigration, which rebuilds the Up expressions and reverses them. ForwardOnlyMigration already rejects Down collection, and a regular two-way Migration executes its author-defined Down() rather than calling these expression Reverse() methods automatically. The ForwardOnlyMigration documentation is therefore a recommendation for this conditional-DDL case, not a behavior change for ordinary migrations.

Persisting a conditional outcome could make generated reversal possible, but it would need durable, per-expression ownership state. Today VersionInfo records only the migration version, applied time, and description after the whole Up succeeds; it has no record of which conditional DDL was applied or skipped. Inspecting the schema later cannot prove whether a pre-existing table or column was created by this migration.

That looks like a broader metadata and migration-contract feature, especially across processors with different conditional-DDL support. The narrow behavior here is to fail closed: do not synthesize a destructive Down when the runner cannot establish ownership. An explicit Down() remains available when the migration author can provide that domain-specific guarantee.

I am happy to reshape or split this if you would prefer to design state tracking before changing the auto-reversal contract.

jzabroski:
No, I think we go with this approach for now but if you're interested in submitting an ADR to do the more sophisticated approach  im interested in discussing it. I've thought a lot about it.

hopkienne:
@jzabroski Understood. I will keep this PR limited to the fail-safe behavior and the regression coverage already included.

I am interested in the ADR. I would submit it separately rather than widen this bug-fix PR. Its initial scope would cover per-expression execution and ownership records, VersionInfo evolution and migration, transaction and retry semantics, and processor support. That should let us decide whether conditional auto-reversal is feasible without relying on ambiguous schema inspection.

jzabroski:
@hopkienne Can you rebase this to main

hopkienne:
@jzabroski Rebased onto the current main (8553f253) and force-pushed the updated branch. The new head is f565f47d.

Validation after the rebase: FluentMigrator.Tests passed with 5,095 passed, 910 skipped, and 0 failed.

Issue #2090, opened 2025-06-23
Is it possible to create a foreign key to a composite primary key for SQLite?


tom42:
**Describe the question**

I'm trying to create a foreign key to a composite primary key for SQLite as shown below:

```
CREATE TABLE Area(
  ArticleId TEXT NOT NULL,
  AreaGroupIndex INTEGER NOT NULL,
  "Index" INTEGER NOT NULL,
  PRIMARY KEY(ArticleId, AreaGroupIndex, "Index"),
  FOREIGN KEY(ArticleId, AreaGroupIndex) REFERENCES AreaGroup(ArticleId, "Index") ON DELETE CASCADE
)
```

The issue is that SQLite allows foreign key creation only as part of a `CREATE TABLE` statement as shown above, and I cannot figure out how to do this with FluentMigrator's fluent syntax or whether it is possible at all (I searched the issues here and the answer *seems* to be 'no'?)

**Expected benefit**
I can of course create my tables with `Execute.Sql`, but I'd really prefer if it was possible with fluent syntax. We make use of fluent syntax to basically reuse fragments of SQL code, and we lose that ability whenever we write raw SQL.

**Information (please complete the following information):**
 - FluentMigrator 7.1

rhegner:
I'm having the same issue, see https://github.com/fluentmigrator/fluentmigrator/issues/1747

ArthurSilvaAutoLab:
I'm also facing this issue. I'm here just to "upvote" this issue.

This one is the last obstacle so we can use only FM (and get rid of ugly .sql scripts). I'm very very anxious about it 😆

tom42:
What if the `Create.Table` syntax was extended with an API similar to the existing one to create foreign keys, so that one can do:

```
Create.Table("myTable")
  .WithColumn("a").AsString()
  .WithColumn("b").AsString()
  .WithForeignKey() // This gets us access to syntax similar to what we get from Create.ForeignKey().FromTable()
  .ForeignColumns("a", "b").ToTable("someOtherTable").PrimaryColumns("c", "d");
```

For a quick mock up it looks neat to me as a user of the API, but might be a pain to implement.

ArthurSilvaAutoLab:
This issue appeared again to me, and I'm still thrilled to have this feature implemented...

I will use this comment to cheer @jzabroski for the hard-work!

[...]

Issue #1783, opened 2024-04-17
How to get rid of "Description:" in every column description?


Lelishe:
Аfter upgrading to FluentMigrator 5.2.0 in every column description, generated via WithColumnDescription(..) there is a prefix "Description:". We don't use AdditionalColumnDescriptions.
Is it possible to disable this new feature somehow? Or use a feature flag? I'ts hardcoded in GenericDescriptionGenerator.GenerateDescriptionStatements:
`                string initialDescriptionStatement = GenerateColumnDescription(
                    "Description",
                    expression.SchemaName,
                    expression.TableName,
                    column.Name,
                    "Description:" + column.ColumnDescription);`

 - FluentMigrator 5.2.0
 - FluentMigrator runner : in-process
 - Postgresql 14

martinib77:
The same is happening to. me. I think is a bug that was introduced with the commit b8dd18d445bd6e5927f619011dd0b5f2bad7b224.

jzabroski:
@azerios Would you be able to look at this issue for me?

jzabroski:
I'll try to address this this week as part of shipping 6.0. Very sorry I didn't catch this in review.

jzabroski:
I'm almost done fixing this.  The remaining issue is that the new AdditionalDescriptions feature doesn't support four types of expressions that otherwise support expressions: `AlterTableExpression`, `CreateColumnExpression`, and `AlterColumnExpression`.

https://github.com/fluentmigrator/fluentmigrator/blob/214345132f8b599ce72fdd21bf966baf894aae2d/src/FluentMigrator.Runner.Core/Generators/IDescriptionGenerator.cs#L9-L14

jzabroski:
I've worked on this in my local but not yet ready to ship the full fix.

PR #2345, opened 2026-07-29
Explain what to do instead when SQLite rejects a foreign key

### LINKED ISSUE #1784: Can no longer create foreign key on SQLite with FluentMigrator.Runner 5.2.0
**Describe the bug**
A migration for a table that includes an foreign key column and one  other primary key columns are broken for Sqlite in 5.2.0 version.
I get the following error:

> ----> FluentMigrator.Exceptions.DatabaseOperationNotSupportedException : Foreign keys are not supported in SQLite

**To Reproduce**
Create a foreign key like this:
```
Create.ForeignKey("fk_TestTable2_TestTableId_TestTable_Id")
			.FromTable("TestTable2").ForeignColumn("TestTableId")
			.ToTable("TestTable").PrimaryColumn("Id");
```

**Expected behavior**
I expect to be able to create a foreign key in SQLite, like in earlier versions of FluentMigrator.Runner

OS: Windows 11
Platform: .NET 8
FluentMigrator version: 5.2.0
FluentMigrator runner: in process
Database Management System: SQLite
Database Management System Version: Microsoft.Data.Sqlite 8.0.4

### LINKED ISSUE #2117: Foreign keys are not supported in SQLite
**Describe the bug**

Getting such exception when trying to create some tables on the empty SQLite DB, even with Foreign Keys set to True in connection string. Also checking DB file with DB Browser for SQLite, Foreign Keys pragma is enabled.

```
Foreign keys are not supported in SQLite
   at FluentMigrator.Runner.Generators.CompatibilityModeExtension.HandleCompatibility(CompatibilityMode mode, String message)
   at FluentMigrator.Runner.Generators.SQLite.SQLiteGenerator.Generate(CreateForeignKeyExpression expression)
   at FluentMigrator.Runner.Processors.ProcessorBase.Process(CreateForeignKeyExpression expression)
   at FluentMigrator.Expressions.CreateForeignKeyExpression.ExecuteWith(IMigrationProcessor processor)
   at FluentMigrator.Runner.MigrationRunner.<>c__DisplayClass75_0.<ExecuteExpressions>b__1()
   at FluentMigrator.Runner.StopWatch.Time(Action action)
   at FluentMigrator.Runner.MigrationRunner.AnnounceTime(String message, Action action)
   at FluentMigrator.Runner.MigrationRunner.ExecuteExpressions(ICollection`1 expressions)
   at FluentMigrator.Runner.MigrationRunner.ExecuteMigration(IMigration migration, Action`2 getExpressions)
   at FluentMigrator.Runner.MigrationRunner.ApplyMigrationUp(IMigrationInfo migrationInfo, Boolean useTransaction)
   at FluentMigrator.Runner.MigrationRunner.MigrateUp(Int64 targetVersion, Boolean useAutomaticTransactionManagement)
   at FluentMigrator.Runner.MigrationRunner.MigrateUp(Boolean useAutomaticTransactionManagement)
   at FluentMigrator.Runner.MigrationRunner.MigrateUp()
   at Gw2Gizmos.Data.Initialization.DatabaseInitializer.EnsureDatabase(String connectionString) in C:\Users\gosfe\Repositories\gw2-gizmos\Gw2Gizmos.Data.Initialization\DatabaseInitializer.cs:line 41
   at Program.<>c__DisplayClass0_0.<<Main>$>b__1(HostBuilderContext context, IServiceCollection services) in C:\Users\gosfe\Repositories\gw2-gizmos\Gw2Gizmos.Data.Worker\Program.cs:line 32
```

**To Reproduce**
Create migration which tries to add foreign key with `Create.ForeignKey` on a table.

**Expected behavior**
Foreign key gets created.

 - OS: Windows 11
 - Platform: .NET8
 - FluentMigrator version: 7.1.0
 - FluentMigrator runner: FluentMigrator.Runner.SQLite
 - Database Management System: SQLite
 - Database Management System Version [ e.g., "SQL Server 2019 Enterprise Edition", "Oracle 18xe", etc ]

**Additional context**
~~Add any other context about the problem here.~~

jzabroski:

Milestone 9.0.0, first of the SQLite items. **#1784 and #2117 are the same root cause** — I'd close one as a duplicate of the other.

## They are not a missing feature

Both issues report "Foreign keys are not supported in SQLite" and conclude FluentMigrator cannot do foreign keys on SQLite. It can. SQLite accepts them only inside `CREATE TABLE` — there is no `ALTER TABLE ... ADD CONSTRAINT` — and FluentMigrator already generates that inline form, marking the follow-up expression with `$$IGNORE$$_` so it is skipped rather than rejected ([SQLiteColumn.cs:43](src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteColumn.cs#L43), [SQLiteGenerator.cs:150](src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs#L150)).

So the working call was one line away from the failing one, and the message gave no hint of it. #2117's reporter worked this out themselves and said exactly what was needed:

> the exception message could be more clear, specifying that it's not supported because of that method usage, but that it works via `WithColumn(<...>).<...>.ForeignKey`

That is what this PR does.

## The new message

`Create.ForeignKey`, for `FK_Test` on `TestTable1`:

> Foreign key `'FK_Test'` cannot be created with Create.ForeignKey on SQLite. SQLite accepts foreign keys only inside a CREATE TABLE statement; it has no ALTER TABLE ... ADD CONSTRAINT. Declare the key on the column when the table is created — `Create.Table("TestTable1").WithColumn("TestColumn1").AsInt32().ForeignKey("TestTable2", "TestColumn2")` — which FluentMigrator generates inline and SQLite accepts. To add one to a table that already exists, rebuild the table: create a replacement with the key declared, copy the rows across, drop the original and rename. If you would rather these statements be skipped than fail, register SQLite with `AddSQLite(compatibilityMode: CompatibilityMode.LOOSE)`.

The example is built from the key's own tables and columns, so it can be pasted rather than translated.

**Composite keys get a generic example instead.** There is no single-column fluent equivalent to suggest for them — that is #2090 — and printing one that would not compile is worse than being generic. Pinned by a test so a future change to #2090 has to revisit it deliberately.

`Delete.ForeignKey` explains the table rebuild, since dropping a constraint has no `ALTER` form either.

## Why the LOOSE hint earns its place

#1784 is titled "Can no longer create foreign key", and @jzabroski's diagnosis on the thread was:

> I think the issue is CompatibilityMode is now strict by default in 5.0

That is the actual regression: the statement was silently skipped before, and STRICT turned the skip into an exception. `AddSQLite` already takes a `compatibilityMode` parameter — the thread's suggested workaround of copy-pasting the DI extension has since shipped — so the message points at a switch that exists.

## Scope

**No behaviour change.** The same expressions throw under STRICT and are skipped under LOOSE, exactly as before. This is a diagnostics change.

I deliberately did **not** implement `Create.ForeignKey` via the 12-step table rebuild. That would silently recreate a user's table — copying data, dropping and renaming — as a side effect of a call that reads like a constraint addition. If that is wanted it should be an explicit opt-in with its own design discussion, not smuggled in behind an error-message fix.

## Tests

Four tests in `SQLiteConstraintsTests`, pinning the parts of the message a user has to act on rather than the whole string:

- names the offending key
- says *why* (`only inside a CREATE TABLE statement`, `ALTER TABLE`)
- gives the working form built from this key's own tables and columns
- offers the LOOSE escape hatch
- falls back to a generic example for composite keys
- and `LooseCompatibilityModeSkipsForeignKeysInsteadOfFailing` asserts the behaviour the message advertises actually holds

Full suite: `5213 passed, 940 skipped, 0 failed` (net48).

## Remaining milestone items

- **#2090** — genuine feature gap, and per @jzabroski's comment on that thread it is cross-provider, not SQLite-specific. Next up as an ADR rather than an implementation, per the same comment.
- **#1783** — unrelated to SQLite; the `"Description:"` prefix is still hardcoded in `GenericDescriptionGenerator` and `MySqlColumn`. Separate PR.

Fixes #1784.
Fixes #2117.

### DIFF
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs
@@ -211,8 +211,8 @@ namespace FluentMigrator.Runner.Generators.SQLite
         }
 
         /// <summary>
-        /// The trailing sentence pointing at LOOSE compatibility mode, which downgrades these
-        /// failures to a warning.
+        /// The trailing sentence pointing at LOOSE compatibility mode, which skips these statements
+        /// instead of failing.
         /// </summary>
         private const string CompatibilityModeHint =
             "If you would rather these statements be skipped than fail, register SQLite with " +
--- a/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs
+++ b/src/FluentMigrator.Runner.SQLite/Generators/SQLite/SQLiteGenerator.cs
@@ -143,6 +143,11 @@ namespace FluentMigrator.Runner.Generators.SQLite
         }
 
         /// <inheritdoc />
+        /// <remarks>
+        /// SQLite accepts foreign keys only as part of a <c>CREATE TABLE</c> statement - there is no
+        /// <c>ALTER TABLE ... ADD CONSTRAINT</c>. Declaring the key on the column when the table is
+        /// created does work and is generated inline; see <see cref="SQLiteColumn"/>.
+        /// </remarks>
         public override string Generate(CreateForeignKeyExpression expression)
         {
             // If a FK name starts with $$IGNORE$$_ then it means it was handled by the CREATE TABLE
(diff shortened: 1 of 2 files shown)


jzabroski:
PR Review Notes:
1. I like the new message Claude wrote.
2. Need to investigate Copilot's feedback as it is surprising.

PR #2347, opened 2026-07-29
ADR: table-level (composite) foreign keys in Create.Table

jzabroski:

Milestone 9.0.0, third of the open items. **Design only — no production code.**

Per @jzabroski on #2090: *"we can have Copilot first propose an ADR as an intermediate task, and then have it implement that ADR in a separate task."* This is that intermediate task.

## The gap, stated precisely

| Want | Column-level `ForeignKey(...)` | `Create.ForeignKey(...)` |
|---|---|---|
| Single-column FK, inline in `CREATE TABLE` | yes | no — emits `ALTER TABLE` |
| **Composite FK, inline in `CREATE TABLE`** | **no** | **no** |
| Composite FK, as `ALTER TABLE` | no | yes |

Confirming @jzabroski point 2: this is **not SQLite-specific**. No provider can express an inline composite foreign key today. SQLite merely makes it load-bearing, because it has no `ALTER TABLE ... ADD CONSTRAINT` at all — so the `Create.ForeignKey` route is closed and inline is the only one that exists. That is why #2090, #1784 and #2117 all arrive as SQLite reports.

## Where I disagree with the proposed API

The ADR recommends @tom42 table-level clause over the column-level overloads in the same thread, and I want that disagreement visible rather than buried in the document.

The proposed overloads widen the **primary** side to many columns while the **foreign** side stays fixed at the single column the call hangs off:

```csharp
TNextFk ForeignKey(string primaryTableName, string[] primaryColumnNames);
```

The motivating example needs two *foreign* columns — `FOREIGN KEY(ArticleId, AreaGroupIndex)` — and there is no column to hang that from. So these overloads do not close the gap.

Beyond that, `1 foreign column → N primary columns` is not a valid foreign key in any provider; every accepted form has equal degree on both sides. So the only shapes these overloads add beyond today are shapes the database rejects.

I know the column-count concern was already raised and answered with *"let people write it, run it, and if it fails, fix it... possibly with an Analyzer"*. That is reasonable for a mismatch a user could plausibly intend to write correctly. It is weaker for an API where **every** new overload is degenerate by construction. Happy to be overruled — the ADR is a proposal — but that is the argument.

`ReferencedBy(...)` from the same comment is a genuinely separate and useful idea. It is orthogonal to composite support and the ADR leaves it to be decided on its own merits.

## What it recommends instead

```csharp
Create.Table("Area")
  .WithColumn("ArticleId").AsString().NotNullable()
  .WithColumn("AreaGroupIndex").AsInt32().NotNullable()
  .WithColumn("Index").AsInt32().NotNullable()
  .WithForeignKey()
    .ForeignColumns("ArticleId", "AreaGroupIndex")
    .ToTable("AreaGroup").PrimaryColumns("ArticleId", "Index")
    .OnDelete(Rule.Cascade);
```

Mirrors SQL — a table-level constraint is a peer of the column list, not a property of one column — and reuses `Create.ForeignKey` vocabulary and the existing `ForeignKeyDefinition`, so there is one set of names for one concept.

Generation lands in `GenericColumn.Generate(columns, tableName)`, the same place `SQLiteColumn` already appends its per-column clauses, so every provider inherits it and SQLite existing path becomes a special case of the general one rather than a quirk.

## Also in the ADR

- **Provider support table.** Table-level FK in `CREATE TABLE` is SQL-92 and universally accepted; no provider needs a fallback. Snowflake and Redshift parse but do not enforce, which is worth recording but not an exclusion.
- **Consequences**, including the real cost: the `Create.Table` interface chain grows a branch, and those `ICreateTableColumnOption*` interfaces are public.
- **Out of scope**, explicitly: adding a foreign key to an existing table on SQLite needs the 12-step rebuild (create replacement, copy, drop, rename). That carries real data risk and deserves its own ADR rather than being folded in here.
- **Validation plan**, including the point that syntax acceptance does not prove the constraint landed — integration coverage on an enforcing provider should show a violating insert actually rejected.

## Relationship to the other milestone PRs

- #2345 — SQLite FK diagnostics (#1784, #2117). Its composite-key branch deliberately prints a generic example rather than a single-column one that would not compile, precisely because of this gap. If this ADR is implemented, that fallback should be revisited.
- #2346 — `"Description:"` prefix (#1783). Unrelated.

Refs #2090.

### DIFF
--- a/adr/proposed/TableLevelForeignKeys.md
+++ b/adr/proposed/TableLevelForeignKeys.md
@@ -150,8 +150,8 @@ Specifically:
    and a `Named(string)` for an explicit constraint name.
 2. Populate the existing `ForeignKeyDefinition` and attach it to `CreateTableExpression`
    as a new `ForeignKeys` collection, rather than inventing a parallel model.
-3. Generate the clauses in `GenericColumn.Generate(columns, tableName)` — the same place
-   `SQLiteColumn` already appends its per-column clauses — so every provider inherits the
+3. Generate the clauses in `ColumnBase.Generate(IEnumerable<ColumnDefinition> columns, string tableName)` — the same method
+   `SQLiteColumn` overrides to append its FK clauses — so every provider inherits the
    behaviour and SQLite's existing path becomes a special case of the general one.
 4. Keep `$$IGNORE$$_` semantics for the column-level route so existing migrations are
    untouched.
--- a/adr/proposed/TableLevelForeignKeys.md
+++ b/adr/proposed/TableLevelForeignKeys.md
@@ -1,0 +1,208 @@
+# ADR: Table-Level (Composite) Foreign Keys in Create.Table
+
+## Status
+Proposed
+
+Written as the intermediate design step requested in [#2090](https://github.com/fluentmigrator/fluentmigrator/issues/2090). No implementation is proposed here; the intent is to settle the syntax first.
+
+## Context
+
+FluentMigrator can express a foreign key in two ways, and neither can express a
+**composite foreign key declared as part of `CREATE TABLE`**.
+
+The request in #2090 is the SQL this cannot produce:
+
+```sql
+CREATE TABLE Area(
+  ArticleId TEXT NOT NULL,
+  AreaGroupIndex INTEGER NOT NULL,
+  "Index" INTEGER NOT NULL,
+  PRIMARY KEY(ArticleId, AreaGroupIndex, "Index"),
+  FOREIGN KEY(ArticleId, AreaGroupIndex) REFERENCES AreaGroup(ArticleId, "Index") ON DELETE CASCADE
+)
+```
+
+Two foreign columns referencing two primary columns, inside the table definition.
+
+### Current state
+
+**Column-level.** `IColumnOptionSyntax<TNext, TNextFk>` exposes:
+
+```csharp
+TNextFk ForeignKey(string primaryTableName, string primaryColumnName);
+```
+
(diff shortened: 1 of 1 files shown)


jzabroski:
@tom42 Let me know your thoughts - I actually like Claude's approach the best.

tom42:
> @tom42 Let me know your thoughts - I actually like Claude's approach the best.

TBH I am absolutely not sure what's being suggested here.

jzabroski:
@tom42 To be honest, re-reading what Claude proposed and you proposed, I see Claude maybe took credit for your idea. :(

You proposed:
```c#
 Create.Table("myTable")
  .WithColumn("a").AsString()
  .WithColumn("b").AsString()
  .WithForeignKey() // This gets us access to syntax similar to what we get from Create.ForeignKey().FromTable()
  .ForeignColumns("a", "b").ToTable("someOtherTable").PrimaryColumns("c", "d");
```

If you replace Claude's example with your identical variables, it becomes:

```c#
Create.Table("myTable")
  .WithColumn("a").AsString()
  .WithColumn("b").AsString()
  .WithForeignKey()
    .ForeignColumns("a", "b")
    .ToTable("someOtherTable").PrimaryColumns("c", "c")
    .OnDelete(Rule.Cascade);
```

...which is essentially an exact match, it just added `OnDelete(Rule.Cascade)` at the end.

Anyway, I plan to ship it.

tom42:
Well it is not my idea either, what I proposed is IIRC basically the API from elsewhere, maybe alter table.

PR #2350, opened 2026-07-30
Persist entries passed to WithColumnAdditionalDescriptions

### LINKED ISSUE #2348: `WithColumnAdditionalDescriptions` (Dictionary overload) silently discards all descriptions
### Summary

`IColumnOptionSyntax.WithColumnAdditionalDescriptions(Dictionary<string,string>)` is documented as:

> Sets any additional column descriptions using a Dictionary containing multiple pairs of name and description content

However, the implementation **never stores the entries**. It validates the dictionary (null / empty / duplicate-key checks) and then returns without adding anything to the column. The call is effectively a no-op: no exception is thrown and the descriptions are lost silently.

The **singular** overload `WithColumnAdditionalDescription(string, string)` works correctly, so this is easy to miss.

### Steps to reproduce

Create a migration that adds additional column descriptions via the Dictionary overload:

```csharp
using FluentMigrator;
using System.Collections.Generic;

[Migration(1)]
public class AddAdditionalDescriptions : Migration
{
    public override void Up()
    {
        Create.Table("Users")
            .WithColumn("Id").AsInt32().PrimaryKey()
            .WithColumn("Email").AsString()
                .WithColumnAdditionalDescriptions(new Dictionary<string, string>
                {
                    { "PII",         "Contains personal data" },
                    { "Sensitivity", "High" },
                });
    }

    public override void Down() => Delete.Table("Users");
}
```

Run the migration in preview mode against a provider that supports column descriptions, or inspect the generated `CreateColumnExpression`.

### Expected result

The two additional descriptions (`PII`, `Sensitivity`) are stored on the column and emitted in the generated description SQL the same as when calling the singular overload twice:

```csharp
.WithColumnAdditionalDescription("PII", "Contains personal data")
.WithColumnAdditionalDescription("Sensitivity", "High")
```

### Actual result

`Column.AdditionalColumnDescriptions` is **empty** - both descriptions are dropped. No error, no warning. The generated SQL contains no additional-description statements.

### Workaround

Until this is fixed, call the singular overload once per entry instead of passing a dictionary:

```csharp
.WithColumnAdditionalDescription("PII", "Contains personal data")
.WithColumnAdditionalDescription("Sensitivity", "High")
```
#

### Notes
Related to #2340

Bafyn:
Fixes #2348

### Fix

The `Dictionary` overload now persists every pair. Each builder delegates to its singular `WithColumnAdditionalDescription`, which keeps the null/empty/duplicate/reserved-key validation in one place and actually adds the entries to the column.

Reproduction:

```csharp
Create.Table("Users")
    .WithColumn("Email").AsString()
        .WithColumnDescription("Primary email")
        .WithColumnAdditionalDescriptions(new Dictionary<string, string>
        {
            { "PII",         "Contains personal data" },
            { "Sensitivity", "High" },
        });
```

### Tests

- **Builder unit tests** for all four builders.
- **Generator coverage for multiple additional descriptions**
- **SQL Server end-to-end integration test** that runs a migration exercising `Create.Table`, `Create.Column` and `Alter.Table`/`AddColumn`, then reads the extended properties back and asserts the additional descriptions are persisted in the single `MS_Description` property.

### Verification

```bash
dotnet test test/FluentMigrator.Tests --filter "TestCategory=Generator|TestCategory=Builder"
```

### Notes

- The `Dictionary` overload now throws `InvalidOperationException` for a duplicate key (surfaced by the shared singular validation) rather than silently doing nothing.
- The error message previously said the argument exception was for `descriptionName`, but it validates the value (`description` agument).

### DIFF
--- a/src/FluentMigrator/Builders/Alter/Column/AlterColumnExpressionBuilder.cs
+++ b/src/FluentMigrator/Builders/Alter/Column/AlterColumnExpressionBuilder.cs
@@ -85,12 +85,12 @@ namespace FluentMigrator.Builders.Alter.Column
         {
             // we need to do a drop constraint and then add constraint to change the default value
             var dc = new AlterDefaultConstraintExpression
-                         {
-                             TableName = Expression.TableName,
-                             SchemaName = Expression.SchemaName,
-                             ColumnName = Expression.Column.Name,
-                             DefaultValue = value
-                         };
+            {
+                TableName = Expression.TableName,
+                SchemaName = Expression.SchemaName,
+                ColumnName = Expression.Column.Name,
+                DefaultValue = value
+            };
 
             _context.Expressions.Add(dc);
 
@@ -110,15 +110,15 @@ namespace FluentMigrator.Builders.Alter.Column
         public IAlterColumnOptionSyntax WithColumnAdditionalDescription(string descriptionName, string description)
         {
             if (string.IsNullOrWhiteSpace(descriptionName))
-                throw new ArgumentException("Cannot be the empty string.", "descriptionName");
+                throw new ArgumentException("Cannot be the empty string.", nameof(descriptionName));
 
-            if(description.Equals("Description"))
+            if (descriptionName.Equals("Description"))
                 throw new InvalidOperationException("The given descriptionName is already used as a keyword to create a description, please choose another descriptionName.");
 
             if (string.IsNullOrWhiteSpace(description))
-                throw new ArgumentException("Cannot be the empty string.", "description");
+                throw new ArgumentException("Cannot be the empty string.", nameof(description));
 
-            if (Expression.Column.AdditionalColumnDescriptions.Keys.Count(i => i.Equals(descriptionName)) > 0)
+            if (Expression.Column.AdditionalColumnDescriptions.Keys.Any(i => i.Equals(descriptionName)))
                 throw new InvalidOperationException("The given descriptionName is already present in the columnDescription list.");
 
             Expression.Column.AdditionalColumnDescriptions.Add(descriptionName, description);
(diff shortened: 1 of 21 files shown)


src/FluentMigrator/Builders/Alter/Column/AlterColumnExpressionBuilder.cs
```diff
@@ -110,15 +110,15 @@ public IAlterColumnOptionSyntax WithColumnDescription(string description)
         public IAlterColumnOptionSyntax WithColumnAdditionalDescription(string descriptionName, string description)
         {
             if (string.IsNullOrWhiteSpace(descriptionName))
-                throw new ArgumentException("Cannot be the empty string.", "descriptionName");
+                throw new ArgumentException("Cannot be the empty string.", nameof(descriptionName));
 
-            if(description.Equals("Description"))
+            if (descriptionName.Equals("Description"))
```
Bafyn, line 115: This must've been another inconsistency: the error msg says descriptionName, but it validates the value

jzabroski: Thanks for your attention to detail. I will move this comment to your PR description body.

jzabroski:
Since @hopkienne raised #2360, I thought I would mention that was somethin Claude did flag in a private review - but I did not think it was essential enough to stop the PR.

> Things I'd request changes or ask about:
> 
> 1. **Non-atomic plural add.** Delegating in a loop means if entry #2 fails validation (duplicate, reserved "Description" key, whitespace value), entry #1 is already committed to AdditionalColumnDescriptions. Since these are now per-entry checks that didn't exist before, partial mutation is newly reachable. Suggest validating all entries first, then adding — or at least a doc comment.
> 2. **Behavior/exception changes worth release-noting.** Duplicate keys previously threw ArgumentException from the plural path; now it's InvalidOperationException from the singular. Per-entry whitespace and reserved-key validation is also new for the plural overload — previously-"working" (well, silently ignored) migrations could now throw.
> 3. **Null check inconsistency.** Three builders throw ArgumentException("Cannot be null.", nameof(...)) for a null dictionary; AlterTableExpressionBuilder uses ArgumentNullException. Standardize on ArgumentNullException — the new null tests use Should.Throw<ArgumentException> so they'd still pass.
> 4. **Integration test gaps.** The migration exercises all four builders, but the test only asserts Col1–Col3; Col4 (Alter.Column path) is migrated and never verified — if that path still doesn't persist through the SQL Server description generator, this test won't catch it. Also AssertPerKeyAdditionalDescriptions is dead code (remove or use), and the rollback Execute adds AddScoped<TaskExecutor>() while the migrate one doesn't — presumably redundant either way.
> 5. **Diff noise.** Roughly half the src changes are object-initializer re-indentation in untouched methods (the FK builders, ToSchema, etc.), plus similar churn in SqlServer2005GeneratorTests. Fine if you're okay with the blame noise, but it obscures the four-line functional fix in each builder.

PR #2361, opened 2026-07-31
Validate bulk column descriptions before adding

### LINKED ISSUE #2360: WithColumnAdditionalDescriptions leaves partial state when a later entry is invalid
## Summary

`WithColumnAdditionalDescriptions` mutates the column definition before it has validated every entry in the supplied dictionary. If a later entry conflicts with an existing description, the method throws but entries processed earlier remain in the migration expression.

## Reproduction

```csharp
var expression = new CreateColumnExpression
{
    Column = new ColumnDefinition { Name = "TestColumn" },
};
expression.Column.AdditionalColumnDescriptions.Add("Existing", "Value");

var builder = new CreateColumnExpressionBuilder(expression, context);

builder.WithColumnAdditionalDescriptions(new Dictionary<string, string>
{
    ["New"] = "NewValue",
    ["Existing"] = "AnotherValue",
});
```

The call throws `InvalidOperationException` for `Existing`, but `New` has already been added. The expression therefore has two descriptions after a rejected operation.

## Expected behavior

A bulk operation that rejects one input entry should leave the target definition unchanged. The method should validate the whole input collection before adding any descriptions.

## Scope

The same sequential add pattern is present in all four builders changed by #2350:

- `CreateColumnExpressionBuilder`
- `CreateTableExpressionBuilder`
- `AlterColumnExpressionBuilder`
- `AlterTableExpressionBuilder`

## Validation

A focused unit test against `main` at `bc60e1e2` fails: expected the existing single entry to remain, actual count is two. I searched open and closed issues/PRs for this behavior and found no duplicate.

hopkienne:
Fixes #2360.

## Summary
- validate all additional column descriptions before mutating the column definition
- apply the transactional bulk behavior to Create.Column, Create.Table, Alter.Column, and Alter.Table
- add regression coverage for a new entry followed by a duplicate entry in every builder

## Root Cause
The bulk overload delegated to the single-entry method while iterating. A validation failure after an earlier successful entry threw an exception after the migration expression had already been partially mutated.

## Validation
`dotnet test test/FluentMigrator.Tests/FluentMigrator.Tests.csproj --configuration Debug --no-restore --filter "FullyQualifiedName~CreateColumnExpressionBuilderTests|FullyQualifiedName~CreateTableExpressionBuilderTests|FullyQualifiedName~AlterColumnExpressionBuilderTests|FullyQualifiedName~AlterTableExpressionBuilderTests" --verbosity minimal`

Passed: 322 tests.

### DIFF
--- a/src/FluentMigrator/Builders/Alter/Column/AlterColumnExpressionBuilder.cs
+++ b/src/FluentMigrator/Builders/Alter/Column/AlterColumnExpressionBuilder.cs
@@ -109,6 +109,15 @@ namespace FluentMigrator.Builders.Alter.Column
         /// <inheritdoc />
         public IAlterColumnOptionSyntax WithColumnAdditionalDescription(string descriptionName, string description)
         {
+            ValidateColumnAdditionalDescription(descriptionName, description);
+
+            Expression.Column.AdditionalColumnDescriptions.Add(descriptionName, description);
+
+            return this;
+        }
+
+        private void ValidateColumnAdditionalDescription(string descriptionName, string description)
+        {
             if (string.IsNullOrWhiteSpace(descriptionName))
                 throw new ArgumentException("Cannot be the empty string.", nameof(descriptionName));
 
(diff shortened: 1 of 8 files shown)


jzabroski:
Well, that's a first. Was wondering when copilot reviews run out

PR #2363, opened 2026-07-31
Use per-key value for SQL Server Alter.Column additional descriptions.

### LINKED ISSUE #2340: SQL Server: Alter.Column writes the column's main description as the value for every additional description
### Summary

When **altering** a column that has additional descriptions, the SQL Server description generator emits one extended property per description key (`MS_<key>`). The `@value` written for each of those properties is the column's **primary** description text (`ColumnDescription`) instead of that key's own value.

### Steps to reproduce

```csharp
using FluentMigrator;
using System.Collections.Generic;

[Migration(2)]
public class AlterWithAdditionalDescriptions : Migration
{
    public override void Up()
    {
        Alter.Column("Email").OnTable("Users").AsString()
            .WithColumnDescription("Primary email")
            .WithColumnAdditionalDescriptions(new Dictionary<string, string>
            {
                { "PII", "Contains personal data" },
            });
    }

    public override void Down() { /* ... */ }
}
```

Run in preview mode (`migrate --preview`) and inspect the generated SQL for the `MS_PII` extended property (or read `sys.extended_properties` back after applying).

### Expected result

```sql
EXEC sys.sp_addextendedproperty @name = N'MS_PII', @value = N'Contains personal data', ...
```

### Actual result

```sql
EXEC sys.sp_addextendedproperty @name = N'MS_PII', @value = N'Primary email', ...
```

Every additional key is stored with the column's primary description as its value.

### Information (please complete the following information)
 - OS: Windows 11
 - Platform .NET 8
 - FluentMigrator runner: in-process runner
 - Database Management System: SQL Server

**Additional context**
The **create** path is unaffected: it concatenates `key:value` pairs into the single `MS_Description` extended property (via the generic base generator) using the correct values.

The generator is inherited by the SQL Server 2008/2012/2014/2016 generators, so all SQL Server versions are affected.

Bafyn:
Fixes #2340 

`SqlServer2005DescriptionGenerator.GenerateDescriptionStatement(AlterColumnExpression)` emits one extended property per additional description (MS_<key>), but writes the column's MAIN description as the `value` for each of them instead of that key's own value. So on ALTER COLUMN every additional key is stored with the wrong value.

Tests:
- Updated the SQL Server 2005 alter-column additional-description assertions to expect
  the per-key value
- Added a SQL Server end-to-end integration test that alters a column with additional
  descriptions and reads the extended properties back


jzabroski:
I need to think through the implications of fixing this. In other words, could it break behavior across versions. It might require an FMUP0002 analyzer. FMUP prefix is for helping people upgrade across FluentMigrator versions.

jzabroski:
This does make sense to include in 9.0.0. The fundamental issue was nobody ever tested Alter.Column before you wrote this test. It's probably not that common to add extended properties. I will tag this as a breaking change and note in the release notes, but I think the impact will be pretty minor.

I explored adding an analyzer, but I just think its unlikely to be that useful.

Issue #2354, opened 2026-07-31
$(name) renders DateTime as 07/28/2026 13:45:06 - invariant, but not a portable SQL datetime literal


jzabroski:
Claude `claude-opus-5[1m]` filing on behalf of @jzabroski.

Split out of #2336, which deferred this deliberately. The coverage work there is done; this is the one behavioural question it recorded rather than resolved. `test/FluentMigrator.Tests/Unit/TokenSubstitution/BASELINE-FAILURES.md` carries the note: *"Noted in #2336 rather than changed here."*

## Describe the bug

`$(name)` renders a `DateTime` in .NET's invariant general format, which is US-style `MM/dd/yyyy`:

```
$(cutoff)   ->   07/28/2026 13:45:06
$[cutoff]   ->   '2026-07-28T13:45:06'
```

The value is culture-independent — #2324 saw to that — so this is not the German-build-agent bug. It is a *portability* problem: `07/28/2026` is not a datetime literal any supported database will read the same way, and several will not read it at all.

The natural way to use a raw token for a value is to supply the quotes in the script:

```csharp
Execute.Sql("DELETE FROM Audit WHERE CreatedAt < '$(cutoff)'", new Dictionary<string, object>
{
    ["cutoff"] = new DateTime(2026, 7, 28, 13, 45, 6)
});
// emitted: DELETE FROM Audit WHERE CreatedAt < '07/28/2026 13:45:06'
```

What that does depends on the server:

| Provider | `'07/28/2026 13:45:06'` |
|---|---|
| SQL Server | Parses, but interpretation depends on `SET DATEFORMAT` / login language. `07/28` happens to be unambiguous; `07/08` would not be |
| PostgreSQL | Depends on `DateStyle`. Rejects or misreads under `DMY` |
| MySQL | Not a recognised datetime literal; yields `0000-00-00` or an error depending on `sql_mode` |
| Oracle | Depends on `NLS_DATE_FORMAT`; usually an error without an explicit `TO_DATE` |
| SQLite | Stored as the literal text, so date comparisons silently stop working |

The dangerous row is SQL Server, where a day-of-month ≤ 12 will parse under one server's settings and mean a different date under another's.

## Current behaviour, verified

From the matrix snapshots on `main` (`TokenSubstitutionMatrixTests.TokenMatrix_IsStable_family=SqlServer.verified.md`):

[...]

PR #2365, opened 2026-08-01
Require an IQuoter for $[name] token substitution

### LINKED ISSUE #2362: Require an IQuoter for SQL script token substitution (implements ADR from #2355, amended by #2364)
Claude `claude-opus-5[1m]` filing on behalf of @jzabroski.

Implementation task for the ADR at `adr/proposed/RequireQuoterForTokenSubstitution.md`, merged in #2355 and amended in #2364. Filed separately so the change has an issue to reference in release notes; the ADR records the reasoning, this records the work.

> **Rescoped 2026-07-31** to Option I, per the amendment in #2364. The original text described Option G, which would have added a `StandardSqlQuoter` to `FluentMigrator.Abstractions`. That is no longer the plan — see [Why not a fallback quoter](#why-not-a-fallback-quoter) below.

## Summary

`SqlScriptTokenReplacer.ReplaceSqlScriptTokens` takes an optional `IQuoter`. When it is null, `$[name]` falls back to `QuoteSqlLiteral` — a private method in `FluentMigrator.Abstractions` that reimplements the default rendering of `GenericQuoter.QuoteValue` from another assembly. Two implementations of value rendering, held in agreement by a single test.

Delete the duplicate rather than relocate it, and let the one caller that genuinely has no quoter stop needing one.

## Work items

1. **Delete `QuoteSqlLiteral`** and its private helpers. **Add nothing to `FluentMigrator.Abstractions`.**
2. **Make `quoter` required** on the `IDictionary<string, object>` overload by dropping the `= null` default.
3. **Guard inside the `$[name]` branch, not at method entry.** A script containing only `$(name)` never consults the quoter, so an entry guard would reject callers with no need of one — including four of the five mock-based expression test files, which use `$(name)` exclusively. The exception message names the token that forced the failure.
4. **Give the private core an `expandSafeTokens` parameter.** The obsolete `IDictionary<string, string>` overload passes `false`, restoring 8.0.1 behaviour: `$(name)` expands, `$[name]` passes through untouched, no quoter consulted. No public surface grows.
5. **Change the two production call sites** from `processor?.Quoter` to `processor.Quoter`. The `?.` is vestigial — `ExecuteSqlStatementExpression.ExecuteWith` dereferences `processor` unconditionally on the next line, so it never guarded a null processor.
6. **Update the 23 two-arg call sites in the test suite** to state their quoter, and stub `Quoter` on the mock processors whose scripts expand `$[name]`.
7. **`GenericQuoter` is untouched**, keeping its virtual dispatch and its ~30 provider overrides.

## Why not a fallback quoter

Option G proposed adding a `StandardSqlQuoter` to Abstractions for the no-quoter case. Two problems killed it:

- **It would not remove the duplication.** `StandardSqlQuoter.QuoteValue` and `GenericQuoter.QuoteValue` would both exist and both have to agree. The private duplicate becomes a public one; the count stays at two.
- **The name claimed a conformance `GenericQuoter` does not have.** It emits `1`/`0` for booleans where ANSI has `TRUE`/`FALSE`, `0xdead` for binary where ANSI has `X'DEAD'`, and `'2026-07-28T13:45:06'` where ANSI has `TIMESTAMP '2026-07-28 13:45:06'`. Providers override precisely these to correct them.

Both dissolve because the obsolete overload does not actually need a quoter. It dates from `ce4d9428`, 2018-04-04, is present from `v2.0.0`, and **in 8.0.1 it never expanded `$[name]` at all** — the released package returns `"SELECT $[CurrentTimestamp]"` unchanged. It does not need a quoter; it needs its old behaviour back.

## Compatibility

`ReplaceSqlScriptTokens(sqlText, parameters)` stops compiling. This is free externally: 8.0.1 exposes only `ReplaceSqlScriptTokens(string, IDictionary<string, string>)` and has no `IMigrationProcessor.Quoter` at all, both verified against the released package. The entire quoter-threading design is unreleased, so this is the last release in which changing its signature costs shipped users nothing.

**This also reverts an unannounced behaviour change.** Today's `main` routes the string-dictionary overload through the shared core, so a script whose text contains `$[...]` is now substituted where 8.0.1 returned it untouched. Item 4 restores the 8.0.1 behaviour.

## Acceptance criteria

- [ ] The token substitution matrix produces **byte-identical** `*.verified.md` snapshots. Any moved cell means `QuoteSqlLiteral` and `GenericQuoter.QuoteValue` had already drifted, and the diff identifies where.
- [ ] The string-dictionary overload expands `$(name)` and leaves `$[name]` **verbatim**, matching 8.0.1.
- [ ] `ReplaceSqlScriptTokens(sql, parameters, quoter: null)` throws `ArgumentNullException` when the script contains `$[name]`, and the message names the offending token.
- [ ] The same call **succeeds** when the script contains only `$(name)`. This is the narrowed guard's whole point and the easiest thing to regress by tidying the check up to method entry.
- [ ] Full suite green.

## Out of scope

- Whether `IMigrationProcessor` should forbid a null `Quoter` by contract. Item 5 stops the call sites tolerating one; it does not stop a processor returning one.
- Restoring `$[name]` to the string-dictionary overload. If ever wanted, Option H in the ADR is the right shape and should be decided on its own merits.
- `$(name)` DateTime portability (#2354), which concerns the raw token style and is unaffected either way.

## Release note

> `SqlScriptTokenReplacer.ReplaceSqlScriptTokens` now requires an `IQuoter` when the script uses the `$[name]` token style; scripts using only `$(name)` are unaffected and may pass `null`. The private standard-SQL fallback is removed — supply the processor's `Quoter`. The obsolete `IDictionary<string, string>` overload no longer expands `$[name]`, restoring its 8.0.1 behaviour of leaving such text untouched.

Implements the decision in #2355 as amended by #2364. Refs #2336.

jzabroski:

Implements #2362, per the ADR merged in #2355 and amended in #2364.

## What it does

`SqlScriptTokenReplacer` had **two** implementations of value rendering: the private `QuoteSqlLiteral` in `FluentMigrator.Abstractions`, and `GenericQuoter.QuoteValue` in `Runner.Core`, held in agreement only by a test. This deletes the duplicate rather than relocating it.

- **The `IDictionary<string, object>` overload requires a quoter** — the `= null` default is gone.
- **The guard sits inside the `$[name]` branch**, not at method entry. A script using only `$(name)` never consults the quoter, so rejecting it up front would fail callers with no need of one — including four of the five mock-based expression test files. The message names the token that forced the failure.
- **`QuoteSqlLiteral` and its helpers are deleted.** Nothing is added to Abstractions; `GenericQuoter` is untouched, keeping its virtual dispatch and ~30 provider overrides.
- **The private core gains an `expandSafeTokens` flag.** The obsolete `IDictionary<string, string>` overload passes `false`, so `$[name]` and its `$$[[name]]` escape pass through verbatim — exactly as in 8.0.1, which had no such token.
- **The two production call sites drop the vestigial `?.`** on `processor?.Quoter`. `ExecuteWith` dereferences `processor` unconditionally on the next line, so it never guarded a null processor.

## This reverts a silent behaviour change

The string-dictionary overload dates from `ce4d9428`, **2018-04-04**, and is the only public overload in 8.0.1. It cannot supply a quoter. `main` had started routing it through the shared core, so a script whose text contains `$[...]` was being **substituted** where 8.0.1 returned it untouched — an unannounced change on the oldest API in the file. Item 4 restores the old behaviour, pinned by three tests.

## The matrix column, and the check that matters

The matrix had a `$[x] no quoter` column recording what the fallback rendered. With the fallback gone it reads `<throws ArgumentNullException>` for every value in every dialect — one fact repeated 161 times. Removed, and replaced by three focused tests: the throw, a `$(name)`-only script succeeding without a quoter, and an unresolved `$[name]` not needing one either.

**I verified that removing it moved nothing else.** Discounting that column, all seven snapshots are **byte-identical** to before:

```
MySql:     diff-after-removing-column = 0
Oracle:    diff-after-removing-column = 0
Other:     diff-after-removing-column = 0
Postgres:  diff-after-removing-column = 0
SQLite:    diff-after-removing-column = 0
Snowflake: diff-after-removing-column = 0
SqlServer: diff-after-removing-column = 0
```

That was the ADR's strongest acceptance criterion, and it confirms `QuoteSqlLiteral` and `GenericQuoter.QuoteValue` had not drifted — the deletion is provably behaviour-preserving for every path that still renders.

**One correction to the ADR's wording:** it asked for byte-identical snapshots outright. That was carried over from Option G, where the fallback still rendered. Under Option I the no-quoter column *must* change, because there is no longer a fallback to render it. The criterion holds for every column that produces SQL, which is what it was actually protecting.

## Compatibility

`ReplaceSqlScriptTokens(sqlText, parameters)` stops compiling — free externally, since 8.0.1 exposes only the string-dictionary overload and has no `IMigrationProcessor.Quoter` at all. 23 call sites in our own tests were updated: those expanding only `$(name)` now state `null` explicitly; the rest pass a real quoter.

Two mock-based tests that expand `$[name]` gained a `Quoter` stub, per the ADR. The other twenty mock-processor call sites are untouched — their scripts use `$(name)` only, which is exactly why the guard was narrowed.

## Results

```
5460 passed, 941 skipped, 0 failed   (net48)
```

## Reviewer notes

- **Only `net48` ran locally** — the repo lives on a WSL share. CI covers net8.0.
- `BASELINE-FAILURES.md` is kept as the historical record of what the matrix found, with a note that the column it describes no longer exists. It documents a past run, not the current fixture.
- The `$(name)` DateTime portability question (#2354) is untouched and unaffected either way.

Fixes #2362.

### DIFF
--- a/src/FluentMigrator.Abstractions/Expressions/ExecuteSqlScriptExpressionBase.cs
+++ b/src/FluentMigrator.Abstractions/Expressions/ExecuteSqlScriptExpressionBase.cs
@@ -48,7 +48,7 @@ namespace FluentMigrator.Expressions
         /// <param name="sqlScript">The SQL script to execute</param>
         protected void Execute(IMigrationProcessor processor, [StringSyntax("sql")] string sqlScript)
         {
-            var finalSqlScript = SqlScriptTokenReplacer.ReplaceSqlScriptTokens(sqlScript, GetMergedParameters(), processor?.Quoter);
+            var finalSqlScript = SqlScriptTokenReplacer.ReplaceSqlScriptTokens(sqlScript, GetMergedParameters(), processor.Quoter);
             processor.Execute(finalSqlScript);
         }
 
--- a/src/FluentMigrator.Abstractions/Expressions/ExecuteSqlStatementExpression.cs
+++ b/src/FluentMigrator.Abstractions/Expressions/ExecuteSqlStatementExpression.cs
@@ -43,7 +43,7 @@ namespace FluentMigrator.Expressions
         /// <inheritdoc />
         public override void ExecuteWith(IMigrationProcessor processor)
         {
-            var finalSqlScript = SqlScriptTokenReplacer.ReplaceSqlScriptTokens(SqlStatement, GetMergedParameters(), processor?.Quoter);
+            var finalSqlScript = SqlScriptTokenReplacer.ReplaceSqlScriptTokens(SqlStatement, GetMergedParameters(), processor.Quoter);
             processor.Execute(finalSqlScript);
         }
 
--- a/src/FluentMigrator.Abstractions/SqlScriptTokenReplacer.cs
+++ b/src/FluentMigrator.Abstractions/SqlScriptTokenReplacer.cs
@@ -36,9 +36,9 @@ namespace FluentMigrator
         /// <param name="parameters">The tokens to be replaced</param>
         /// <param name="quoter">
         /// The <see cref="IQuoter"/> used to safely quote/format <c>$[name]</c> parameter values
-        /// (e.g. numbers, dates, booleans, <see langword="null"/>, strings). Pass the processor's
-        /// <c>Quoter</c>. May be <see langword="null"/> only when <paramref name="sqlText"/>
-        /// contains no <c>$[name]</c> token, since nothing else consults it.
+/// (e.g. numbers, dates, booleans, <see langword="null"/>, strings). Pass the processor's
+/// <c>Quoter</c>. May be <see langword="null"/> only when <paramref name="sqlText"/>
+/// contains no <c>$[name]</c> token that matches a key in <paramref name="parameters"/>, since nothing else consults it.
         /// </param>
         /// <returns>The SQL script with the replaced tokens</returns>
         /// <exception cref="ArgumentNullException">
--- a/src/FluentMigrator.Abstractions/SqlScriptTokenReplacer.cs
+++ b/src/FluentMigrator.Abstractions/SqlScriptTokenReplacer.cs
@@ -18,7 +18,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
-using System.Text;
 using System.Text.RegularExpressions;
 
 using FluentMigrator.Generation;
@@ -37,10 +36,15 @@ namespace FluentMigrator
         /// <param name="parameters">The tokens to be replaced</param>
         /// <param name="quoter">
         /// The <see cref="IQuoter"/> used to safely quote/format <c>$[name]</c> parameter values
-        /// (e.g. numbers, dates, booleans, <see langword="null"/>, strings). When <see langword="null"/>,
-        /// a basic string-literal quoting fallback is used, which treats every value as a string.
+        /// (e.g. numbers, dates, booleans, <see langword="null"/>, strings). Pass the processor's
+        /// <c>Quoter</c>. May be <see langword="null"/> only when <paramref name="sqlText"/>
+        /// contains no <c>$[name]</c> token, since nothing else consults it.
         /// </param>
         /// <returns>The SQL script with the replaced tokens</returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="quoter"/> is <see langword="null"/> and <paramref name="sqlText"/>
+        /// contains a <c>$[name]</c> token.
+        /// </exception>
         /// <remarks>
         /// Two token styles are supported:
         /// <list type="bullet">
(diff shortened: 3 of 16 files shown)


jzabroski:
Correct on both counts. Addressed in #2366, since this PR merged before the changes landed.

The docs did overstate the requirement. The guard sits inside the branch that renders a value, so a `$[name]` whose key is **absent** from `parameters` is left verbatim and never consults the quoter. That case has a test — `DoesNotRequireAQuoterForAnUnknownSafeToken` — which the docs directly contradicted.

Corrected in all three places, not just the `<param>` tag:

- **`<param name="quoter">`** — "may be null when no `$[name]` token is *actually expanded* … that includes a `$[name]` whose key is absent from `parameters`".
- **`<exception>`** — now says the throw needs a `$[name]` "whose key is present in `parameters`, so a value has to be rendered".
- **`<remarks>`** — same correction, since it made the shorter version of the same claim.

I kept the Autofix commit in history and rebased on top of it. Two things needed carrying forward from it: its wording covered only the `<param>` tag, and it had flattened the `///` indentation to column 0 for those three lines. The resolved version keeps the corrected meaning at the original indentation.

## Second finding also addressed

`TokenSubstitutionMatrix.md` described the call sites as passing `processor?.Quoter`; this PR changes them to `processor.Quoter`. Updated.

Three mentions of `processor?.Quoter` remain in `adr/proposed/RequireQuoterForTokenSubstitution.md` and are deliberately left alone — two are historical context describing the state before this change, and one is a forward-looking out-of-scope item. That file is a merged decision record, so editing it from an implementation PR would muddy both.

## One I caught myself while verifying

The previous commit introduced **18 CS1573 warnings**: I added a `<param>` tag to the private `ReplaceSqlScriptTokensCore` without documenting its other three parameters. CI does not fail on warnings, so it passed green — but #2331 deliberately cleared every warning in this repo, so it was a regression. The tag is now a plain `//` comment, since the method is private and needs no XML doc.

Verified rather than assumed:

```
CS1573 at 2c461a9b (what CI built):  18
CS1573 now:                           0
```

Full suite: 5460 passed, 941 skipped, 0 failed.

PR #2375, opened 2026-08-03
Clone constraint definitions independently

### LINKED ISSUE #2374: ConstraintDefinition.Clone loses schema and shares the column collection
## Summary

`ConstraintDefinition.Clone()` does not produce an independent copy of a constraint definition:

- `SchemaName` is not copied at all.
- `Columns` is assigned by reference, so adding a column to the clone also changes the source definition.

## Reproduction

```csharp
var definition = new ConstraintDefinition(ConstraintType.PrimaryKey)
{
    ConstraintName = "PK_Users",
    SchemaName = "tenant",
    TableName = "Users",
};
definition.Columns.Add("Id");

var clone = (ConstraintDefinition)definition.Clone();
clone.Columns.Add("TenantId");
```

Expected:

- `clone.SchemaName` is `"tenant"`.
- The source has only `"Id"`; the clone has `"Id"` and `"TenantId"`.

Actual:

- `clone.SchemaName` is `null`.
- Both definitions expose the same `Columns` collection.

## Impact

Code using `ICloneable` to create a modified constraint can silently alter the original expression and can target the default schema instead of the requested schema.

## Suggested fix

Copy `SchemaName` and allocate a new collection from `Columns`, matching the clone implementations of `ForeignKeyDefinition` and `IndexDefinition`.

## Validation

A focused unit test fails on `main` at `85aa53fe`. I searched open and closed issues and PRs for this behavior and found no duplicate.

hopkienne:
Fixes #2374.

## Summary
- Preserve the schema when cloning a constraint definition.
- Create a distinct column collection for the clone.
- Add regression coverage for both behaviors.

## Root cause
`ConstraintDefinition.Clone()` omitted `SchemaName` and assigned the original `Columns` list to the clone. A clone therefore lost schema-qualified constraint information, and editing its columns also edited the original definition.

## Validation
```
dotnet test test\FluentMigrator.Tests\FluentMigrator.Tests.csproj --configuration Debug --no-restore --filter "FullyQualifiedName~ConstraintDefinitionTests|FullyQualifiedName~CreateConstraintExpressionTests|FullyQualifiedName~DeleteConstraintExpressionTests" --verbosity minimal
```

### DIFF
--- a/src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
+++ b/src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
@@ -83,7 +83,7 @@ namespace FluentMigrator.Model
         {
             var result = new ConstraintDefinition(_constraintType)
             {
-                Columns = new List<string>(Columns),
+                Columns = new HashSet<string>(Columns),
                 ConstraintName = ConstraintName,
                 SchemaName = SchemaName,
                 TableName = TableName
--- a/src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
+++ b/src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
@@ -83,8 +83,9 @@ namespace FluentMigrator.Model
         {
             var result = new ConstraintDefinition(_constraintType)
             {
-                Columns = Columns,
+                Columns = new List<string>(Columns),
                 ConstraintName = ConstraintName,
+                SchemaName = SchemaName,
                 TableName = TableName
             };
 
--- a/test/FluentMigrator.Tests/Unit/Definitions/ConstraintDefinitionTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Definitions/ConstraintDefinitionTests.cs
@@ -28,6 +28,25 @@ namespace FluentMigrator.Tests.Unit.Definitions
     public class ConstraintDefinitionTests
     {
         [Test]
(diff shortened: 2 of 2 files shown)


src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
```diff
@@ -83,8 +83,9 @@ public object Clone()
         {
             var result = new ConstraintDefinition(_constraintType)
             {
-                Columns = Columns,
+                Columns = new List<string>(Columns),
```
jzabroski: Should this also clone the Columns? Or should that be the caller's responsibility? I am inclined to think shallow clone semantics is best. Once you start doing deep clone for users, you get into policy vs. mechanism issues as each consumer might have a different definition of what a deep clone does. However, if the shallow clone works correctly, they can always break the problem into smaller sub-steps and build a DeepClone themselves.

jzabroski (changes requested):

src/FluentMigrator.Abstractions/Model/ConstraintDefinition.cs
hopkienne: Thanks, that policy distinction is helpful. The merged version retains a distinct `HashSet<string>` for the columns, so it preserves the source collection's uniqueness semantics while avoiding shared mutable state. I will follow the repository's established clone contracts rather than infer broader deep-clone behavior in future.

PR #2376, opened 2026-08-03
Fix #2367: console --include-untagged crash for explicit values.

### LINKED ISSUE #2367: `Migrate.exe --include-untagged=<value>` always crashes (exit code 3) for every explicit value
## Summary

The FluentMigrator console runner (`Migrate.exe` / `FluentMigrator.Console`) accepts an `--include-untagged[=value]` option to selectively enable/disable untagged **migrations** and untagged **maintenance** objects (`mi`/`migrations`, `ma`/`maintenance`, with an optional `+`/`-` suffix).

Only the **bare** `--include-untagged` (no value) works. Passing **any explicit value** (`mi`, `ma`, `migrations`, `maintenance`, with or without `+`/`-`) makes the runner throw `ArgumentOutOfRangeException` while parsing the argument, which is caught by the top-level
handler and reported as **exit code 3**. The intended value can never be supplied.

The failure happens during **command-line parsing**, before any database connection is made, so it reproduces regardless of provider, connection string, or migrations present.

## Steps to reproduce

Because the exception is thrown at argument-parse time, you do not need a reachable database or valid migrations - any invocation that reaches option parsing is enough.

```bash
# Any explicit value fails. Example with the short "migrations" key:
Migrate.exe --provider SqlServer2008 \
            --connection "Server=.;Database=Whatever;Trusted_Connection=true;" \
            --assembly MyMigrations.dll \
            --task migrate \
            --include-untagged=mi
```

Other values that all fail the same way:

```bash
Migrate.exe ... --include-untagged=ma
Migrate.exe ... --include-untagged=mi-
Migrate.exe ... --include-untagged=ma+
Migrate.exe ... --include-untagged=migrations
Migrate.exe ... --include-untagged=maintenance
```

The only form that does **not** crash is the valueless flag (which enables both):

```bash
Migrate.exe ... --include-untagged
```

## Expected vs. actual

| | |
|---|---|
| **Expected** | `--include-untagged=mi-` disables untagged migrations; `--include-untagged=ma+` enables untagged maintenance; comma-separated combos like `--include-untagged=maintenance+,migrations-` work; the process runs and exits normally. |
| **Actual** | Every explicit value throws `ArgumentOutOfRangeException`, caught as a generic exception => **exit code 3**; the migration task never runs. |

## Workaround

Until fixed, use only the valueless flag (which enables **both** untagged migrations and untagged maintenance):

```bash
Migrate.exe ... --include-untagged
```

Bafyn:
## Summary

`Migrate.exe --include-untagged=<value>` crashed (exit code 3) for **every** explicit value. Only the bare `--include-untagged` flag worked, so untagged migrations and untagged maintenance objects could not be enabled/disabled individually from the console runner.

Fixes #2367

### DIFF
--- a/src/FluentMigrator.Console/MigratorConsole.cs
+++ b/src/FluentMigrator.Console/MigratorConsole.cs
@@ -206,7 +206,7 @@ namespace FluentMigrator.Console
                     },
                     {
                         "include-untagged:",
-                        "Include untagged migrations and/or maintenance objects.",
+                        "Include untagged migrations and/or maintenance objects. Available values are: ma, maintenance, mi, migrations with an optional '+' or '-' at the end to enable or disable the option. Multiple values may be given when separated by a comma.",
                         v =>
                         {
                             if (string.IsNullOrEmpty(v))
--- a/src/FluentMigrator.Console/MigratorConsole.cs
+++ b/src/FluentMigrator.Console/MigratorConsole.cs
@@ -215,38 +215,23 @@ namespace FluentMigrator.Console
                             }
                             else
                             {
-                                var items = v.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                    .Select(x => x.ToLowerInvariant().Trim())
-                                    .Where(x => !string.IsNullOrEmpty(x))
-                                    .Select(
-                                        x =>
-                                        {
-                                            var hasOption = x.EndsWith("+") || x.EndsWith("-");
-                                            var enable = !x.EndsWith("-");
-                                            var name = hasOption ? x.Substring(0, x.Length - 1) : x;
-                                            return new
-                                            {
-                                                FullName = x,
-                                                ShortName = name.Substring(0, Math.Min(2, name.Length)),
-                                                Enable = enable,
-                                            };
-                                        });
-                                foreach (var item in items)
+                                foreach (var raw in v.Split([','], StringSplitOptions.RemoveEmptyEntries))
                                 {
-                                    switch (item.ShortName)
-                                    {
-                                        case "ma":
-                                            IncludeUntaggedMaintenances = item.Enable;
-                                            break;
-                                        case "mi":
-                                            IncludeUntaggedMigrations = item.Enable;
-                                            break;
-                                        default:
-                                            throw new ArgumentOutOfRangeException(
-                                                $"The argument {item.FullName} is not supported. "
-                                              + "Valid values are: ma, maintenance, mi, migrations with an optional '+' or '-' at the end to enable or disable the option. "
-                                              + "Multiple values may be given when separated by a comma.");
-                                    }
+                                    var token = raw.Trim();
+                                    if (token.Length == 0)
+                                        continue;
+
+                                    var enable = !token.EndsWith("-", StringComparison.Ordinal);
+
+                                    if (token.StartsWith("ma", StringComparison.OrdinalIgnoreCase))
+                                        IncludeUntaggedMaintenances = enable;
+                                    else if (token.StartsWith("mi", StringComparison.OrdinalIgnoreCase))
+                                        IncludeUntaggedMigrations = enable;
+                                    else
+                                        throw new ArgumentOutOfRangeException(
+                                            $"The argument {token} is not supported. "
+                                          + "Valid values are: ma, maintenance, mi, migrations with an optional '+' or '-' at the end to enable or disable the option. "
+                                          + "Multiple values may be given when separated by a comma.");
                                 }
                             }
                         }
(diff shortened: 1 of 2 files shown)


src/FluentMigrator.Console/MigratorConsole.cs
```diff
@@ -227,7 +227,7 @@ public int Run(params string[] args)
                                             return new
                                             {
                                                 FullName = x,
-                                                ShortName = name.Substring(Math.Min(2, name.Length)),
+                                                ShortName = name.Substring(0, Math.Min(2, name.Length)),
```
jzabroski: [CA1846](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846) however recommends to use AsSpan over String.Substring.

src/FluentMigrator.Console/MigratorConsole.cs
```diff
@@ -426,7 +426,7 @@ private int ExecuteMigrations()
                         opt.PreviewOnly = PreviewOnly;
                         opt.ProviderSwitches = ProviderSwitches;
                         opt.StripComments = StripComments;
-                        opt.Timeout = Timeout == null ? null : (TimeSpan?) TimeSpan.FromSeconds(Timeout.Value);
+                        opt.Timeout = Timeout == null ? null : (TimeSpan?)TimeSpan.FromSeconds(Timeout.Value);
```
jzabroski: Whitespace changes add complexity to PR review. Please try to keep that out of PRs.

test/FluentMigrator.Tests/Unit/Runners/MigratorConsoleTests.cs
```diff
@@ -77,6 +77,97 @@ public void CanInitMigratorConsoleWithValidArgumentsRegardlessOfCase()
...
+        [TestCase("migrations+,maintenance-", true, false)]
+        [TestCase("migrations-,maintenance+", false, true)]
+        [TestCase("migrations-,maintenance-", false, false)]
+        // long-form order independence
+        [TestCase("maintenance+,migrations-", false, true)]
+        [TestCase("maintenance-,migrations+", true, false)]
+        // long-form / short-form mixed
+        [TestCase("migrations-,ma+", false, true)]
+        [TestCase("mi+,maintenance-", true, false)]
+        public void IncludeUntaggedParsesValue(string value, bool expectedMigrations, bool expectedMaintenances)
```
jzabroski, line 131: This is how I would have written this test. Nice job.

jzabroski (changes requested):
Changes are fairly minor. I'd like to merge this but please incorporate the suggestions.

src/FluentMigrator.Console/MigratorConsole.cs
Bafyn: makes sense, it's just that it happens automatically on file formatting as some pieces do not follow code styling

src/FluentMigrator.Console/MigratorConsole.cs
Bafyn: We cannot use that here really as we're not passing the result of the substring anywhere (like a metho arg) + it becomes a property of an anonymous type that cannot use Spans

src/FluentMigrator.Console/MigratorConsole.cs
Bafyn: reverted

src/FluentMigrator.Console/MigratorConsole.cs
Bafyn: applied a slightly different approach to reduce the number of string allocations

jzabroski: makes sense. I realized I forgot one comment yesterday

src/FluentMigrator.Console/MigratorConsole.cs
```diff
@@ -215,38 +215,23 @@ public int Run(params string[] args)
...
+                                    var enable = !token.EndsWith("-", StringComparison.Ordinal);
+
+                                    if (token.StartsWith("ma", StringComparison.OrdinalIgnoreCase))
+                                        IncludeUntaggedMaintenances = enable;
+                                    else if (token.StartsWith("mi", StringComparison.OrdinalIgnoreCase))
+                                        IncludeUntaggedMigrations = enable;
+                                    else
+                                        throw new ArgumentOutOfRangeException(
+                                            $"The argument {token} is not supported. "
+                                          + "Valid values are: ma, maintenance, mi, migrations with an optional '+' or '-' at the end to enable or disable the option. "
```
jzabroski, line 233: We should add this to the command's long description  too, while we are in here. Forgot to mention this yesterday.

jzabroski (changes requested):

src/FluentMigrator.Console/MigratorConsole.cs
Bafyn: you mean here? 
image

jzabroski: Yes. The ideal is Migrate.exe help menu should mention these as options vs. only the VitePress site.

Bafyn: Example: 

image

jzabroski: Yes, right.

PR #2378, opened 2026-08-04
Fix SQL literal prefix expansion

### LINKED ISSUE #2353: ExpandOverEnclosingQuotes currently walks left over any alphanumeric/underscore
**Describe the bug**
ExpandOverEnclosingQuotes currently walks left over any alphanumeric/underscore run before the opening quote. In SQL, keywords/identifiers can legally touch a string literal delimiter without whitespace (e.g. LIKE'$(name)'), so this code fix can accidentally delete preceding SQL text instead of only stripping literal prefixes (N/E/B/X/_utf8mb4/U&). Please restrict the prefix expansion to known prefixes that occur at an identifier boundary.

_Originally posted by @copilot https://github.com/fluentmigrator/fluentmigrator/pull/2339#discussion_r3668520729_

snowyukitty:
## Summary

- restrict enclosing-quote expansion to recognized SQL literal prefixes at identifier boundaries
- preserve adjacent SQL keywords, identifiers, and operators
- add regression coverage for case-insensitive prefixes, MySQL charset introducers, and boundary cases

## Testing

- `dotnet test test/FluentMigrator.Analyzers.Tests/FluentMigrator.Analyzers.Tests.csproj --configuration Release` (58 passed, net48)

Fixes #2353

### DIFF
--- a/src/FluentMigrator.Analyzers/RawSqlTokenInterpolationCodeFixProvider.cs
+++ b/src/FluentMigrator.Analyzers/RawSqlTokenInterpolationCodeFixProvider.cs
@@ -131,14 +131,58 @@ namespace FluentMigrator.Analyzers
                 return tokenSpan;
             }
 
-            // Walk back over a literal prefix: N'...', E'...', U&'...', _utf8mb4'...'.
-            var prefixStart = start;
-            while (prefixStart > 0 && (char.IsLetterOrDigit(text[prefixStart - 1]) || text[prefixStart - 1] == '_' || text[prefixStart - 1] == '&'))
+            // Consider the entire identifier-like run before the quote. It is part of the
+            // replacement only when that whole run is a recognized literal prefix; SQL such as
+            // LIKE'...' and columnN'...' must keep the adjacent keyword or identifier intact.
+            var candidateStart = start;
+            if (candidateStart >= 2 && text[candidateStart - 1] == '&' &&
+                (text[candidateStart - 2] == 'U' || text[candidateStart - 2] == 'u'))
             {
-                prefixStart--;
+                candidateStart -= 2;
+            }
+            else
+            {
+                while (candidateStart > 0 &&
+                       (char.IsLetterOrDigit(text[candidateStart - 1]) || text[candidateStart - 1] == '_'))
+                {
+                    candidateStart--;
+                }
             }
 
+            var candidate = text.ToString(TextSpan.FromBounds(candidateStart, start));
+            var hasIdentifierBoundary = candidateStart == 0 ||
+                                        !(char.IsLetterOrDigit(text[candidateStart - 1]) || text[candidateStart - 1] == '_');
+            var prefixStart = hasIdentifierBoundary && IsKnownLiteralPrefix(candidate) ? candidateStart : start;
+
             return TextSpan.FromBounds(prefixStart, end + 1);
+        }
+
+        private static bool IsKnownLiteralPrefix(string candidate)
+        {
+            if (candidate.Length == 0)
+            {
+                return false;
+            }
+
+            // MySQL character-set introducers use the open-ended _charset_name form, for
+            // example _utf8mb4'...' or _custom_charset'...'.
+            if (candidate[0] == '_')
+            {
+                return candidate.Length > 1 &&
+                       candidate.Skip(1).All(c => char.IsLetterOrDigit(c) || c == '_');
+            }
+
+            switch (candidate.ToUpperInvariant())
+            {
+                case "N":   // national character literal - T-SQL, Oracle, MySQL, standard
+                case "E":   // PostgreSQL escape-string literal
+                case "B":   // bit-string literal
+                case "X":   // hex-string literal
+                case "U&":  // Unicode escape literal
+                    return true;
+                default:
+                    return false;
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/FluentMigrator.Analyzers/RawSqlTokenInterpolationCodeFixProvider.cs
```diff
@@ -131,14 +131,58 @@ private static TextSpan ExpandOverEnclosingQuotes(SourceText text, TextSpan toke
...
+
+            // MySQL character-set introducers use the open-ended _charset_name form, for
+            // example _utf8mb4'...' or _custom_charset'...'.
+            if (candidate[0] == '_')
+            {
+                return candidate.Length > 1 &&
+                       candidate.Skip(1).All(c => char.IsLetterOrDigit(c) || c == '_');
+            }
+
+            switch (candidate.ToUpperInvariant())
```
jzabroski, line 175: @snowyukitty Thanks for your contribution. I wonder if, as part of a code fixer that acts as an _analyzer silencer_, we could provide a path that hints at which dialect the user is targeting. I'm not suggesting you do this, but more as a discussion.

e.g.
```c#
Execute.Sql([Database(Id = "MySQL")] " SELECT _utf8mb4 0b1000001 COLLATE utf8mb4_danish_ci; ");
```

Many CodeFixProviders in other libraries have a legitimate "Nope" code fixer that silences the warning through either a comment or attribute, depending on whether the Roslyn target area supports attributes or not.

matkoniecz:
note it was likely AI generated without adequate review if it is output of the same process as what ended as https://github.com/watchexec/watchexec/pull/1066#issuecomment-5193395417 was

jzabroski:
I did wonder if the goal of these PRs was a novel form of distributed distillation attack. Curious to get your thoughts. I think it would circumvent any future KYC OpenAi, Microsoft and Anthropic put in place, and would have serious negative consequences for open source eventually if that is the attack vector.

src/FluentMigrator.Analyzers/RawSqlTokenInterpolationCodeFixProvider.cs
snowyukitty: That seems like a useful separate design discussion. An explicit dialect hint or a deliberate suppression fixer could let users document intent without asking this fixer to infer dialect semantics from raw SQL. I agree it should remain separate from this narrow token-boundary correction.

snowyukitty:
That is not the goal. This is a human-owned account, and I use AI-assisted tooling to research and prepare open-source contributions; I remain responsible for what is submitted. The purpose is not to collect maintainer feedback as training data, distill a model, or circumvent KYC or provider controls.

The watchexec PR you linked exposed a real verification failure: I had relied on an analogue before testing the exact mapped-drive boundary. I reproduced the counterexample, acknowledged that the patch was unsafe, and closed it. I am tightening the process around exact-environment validation as a result.

If AI-assisted contributions are outside FluentMigrator’s expectations, please say so and I will withdraw this PR.

jzabroski:
@snowyukitty That's a relief. If you're AI, that last reply fooled me.

There's a competing PR for the same issue with slightly different implementation. I have not yet compared them.

I have, like you, switched to a mostly agentic workflow. I wrote 200,000 lines of code at work in the past 4 weeks. I am curious to hear your thoughts and expertise on how you manage to keep up. I spoke with the an author of the GAN paper, Bing Xu of Nvidia, and he feels strongly we won't be writing code in 18 months. I still do a lot of steering but its becoming like playing tetris or some other button mashing game I would try to speedrun/high-score.

Let me know if you're open to a WhatsApp call to discuss and compare notes.

snowyukitty:
Yeah, I’d be happy to chat and compare notes. WhatsApp can work, though Discord or a browser-based meeting app would probably be a little easier and more flexible for me—and it keeps personal phone details separate. Discord is great for an informal conversation; Google Meet, Zoom, or Jitsi would all work too.

I’m currently on UTC+9. Feel free to suggest whichever platform you prefer and a few times that work for you. I’d especially enjoy comparing how we each handle steering, verification, review quality, and keeping a high-throughput agentic workflow sustainable.

snowyukitty:
Closing this in favour of #2385, which is the better fix.

I went back and read `SqlLiteralScanner` on current main rather than the two pull requests. `ReadLiteralPrefixStart` already walks the prefix, already refuses a run preceded by an identifier character — the comment there is `col='x'` has no prefix but `WHERE N'x'` does — and already delegates to a private `IsKnownLiteralPrefix` carrying the same prefix set. So the rule #2353 asks for exists; what was missing was that `RawSqlTokenInterpolationCodeFixProvider` could not reach it.

#2378 solves that by adding a second private `IsKnownLiteralPrefix` about two hundred lines away in the same assembly. #2385 makes the existing helper `internal` and calls it, which is the right shape — one definition of what a literal prefix is, rather than two that can drift apart — and it brings far more test coverage than I did.

One detail worth carrying over if you like it, offered and then dropped: for the MySQL charset introducer, the existing helper accepts any `_` run with `Length > 1`, so `_utf8mb4'x'` and `_%%%'x'` are treated the same. #2378 additionally required the characters after the underscore to be letters, digits or underscores. It is a small tightening and it belongs in whichever helper survives, not in a separate pull request.

Thanks for the time you spent on this one.

PR #2387, opened 2026-08-06
Fix #2386: DROP TABLE IF EXISTS ignores schema (Postgres & SQLite)

### LINKED ISSUE #2386: `Delete Table IF Exists` ignores the schema (Postgres & SQLite) - drops/targets the wrong schema
## Summary

When deleting a table with both `IfExists()` and a non-default schema, the generated `DROP TABLE IF EXISTS` statement omits the schema.

This affects PostgreSQL and SQLite dialects. On PostgreSQL a missing schema resolves to `public`, so the statement silently targets the wrong table.

## Steps to reproduce

```csharp
using FluentMigrator;

[Migration(1)]
public class DropInCustomSchema : Migration
{
    public override void Up()
    {
        Delete.Table("Foo").InSchema("myschema").IfExists();
    }

    public override void Down() { }
}
```

Run the migration against PostgreSQL in preview mode to see the generated SQL:

```sql
DROP TABLE IF EXISTS "public"."Foo"; -- schema "myschema" was silently replaced by "public"
```

## Expected

`DROP TABLE IF EXISTS "myschema"."Foo";` - the same schema used everywhere else.

## Actual

`DROP TABLE IF EXISTS "public"."Foo";` - schema dropped.

## Notes

- The dialects that override `Generate(DeleteTableExpression)` (SQL Server, Oracle, Db2, Snowflake, Redshift, HANA) already pass the schema and are unaffected.

Bafyn:
## Problem

Fixes #2386 

When deleting a table with both `IfExists()` and a non-default schema, the generated `DROP TABLE IF EXISTS` statement omits the schema.

This affects PostgreSQL and SQLite dialects. On PostgreSQL a missing schema resolves to `public`, so the statement silently targets the wrong table.

Only the dialects using the generic `DROP TABLE IF EXISTS` generation are affected (PostgreSQL, SQLite). Dialects that override `Generate(DeleteTableExpression)` (SQL Server, Oracle, Db2, Snowflake, Redshift, HANA) already pass the schema and are unaffected.

### DIFF
--- a/src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
+++ b/src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
@@ -166,7 +166,7 @@ namespace FluentMigrator.Runner.Generators.Generic
         {
             if (expression.IfExists)
             {
-                return FormatStatement(DropTableIfExists, Quoter.QuoteTableName(expression.TableName));
+                return FormatStatement(DropTableIfExists, Quoter.QuoteTableName(expression.TableName, expression.SchemaName));
             }
 
             return FormatStatement(DropTable, Quoter.QuoteTableName(expression.TableName, expression.SchemaName));
--- a/test/FluentMigrator.Tests/Helpers/PostgresTestTable.cs
+++ b/test/FluentMigrator.Tests/Helpers/PostgresTestTable.cs
@@ -74,10 +74,12 @@ namespace FluentMigrator.Tests.Helpers
 
         public void Drop()
         {
+            // Use IF EXISTS so cleanup stays robust even when the table under test was
+            // already dropped by the migration/expression being executed.
             var sb = new StringBuilder();
-            sb.AppendFormat("DROP TABLE {0}", NameWithSchema);
+            sb.AppendFormat("DROP TABLE IF EXISTS {0}", NameWithSchema);
             if (!string.IsNullOrEmpty(_schemaName))
-                sb.AppendFormat(";DROP SCHEMA \"{0}\"", _schemaName);
+                sb.AppendFormat(";DROP SCHEMA IF EXISTS \"{0}\"", _schemaName);
 
             using (var command = new NpgsqlCommand(sb.ToString(), Connection, Transaction))
                 command.ExecuteNonQuery();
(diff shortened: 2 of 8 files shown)


src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
```diff
@@ -166,7 +166,7 @@ public override string Generate(DeleteTableExpression expression)
         {
             if (expression.IfExists)
             {
-                return FormatStatement(DropTableIfExists, Quoter.QuoteTableName(expression.TableName));
+                return FormatStatement(DropTableIfExists, Quoter.QuoteTableName(expression.TableName, expression.SchemaName));
```
Bafyn, line 169: @jzabroski FYI while this is a bug fix, the change might be breaking if someone is relying on the original behavior

jzabroski: This would break databases like MySQL which map a schema to a database. I wouldn't change this line as  ot every database supports first-class schemas. If we do, the righ6approis via an ADR to document the coverage by provider, showing where the gaps are in provider specific support for schema objects.

jzabroski (changes requested):
Let's revert the GenericGenrator change.

src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs
Bafyn: are you sure about that? 
line below (delete unconditionally) does the same https://github.com/fluentmigrator/fluentmigrator/blob/c7e0747f67d58bad7765882b3fcfc5b54b8ecd58/src/FluentMigrator.Runner.Core/Generators/Generic/GenericGenerator.cs#L172

it does pass the schema.

In addition to this, some db providers override quoters which ignore schemas to cover the cases you mentioned.
See https://github.com/fluentmigrator/fluentmigrator/blob/c7e0747f67d58bad7765882b3fcfc5b54b8ecd58/src/FluentMigrator.Runner.MySql/Generators/MySql/MySqlQuoter.cs#L83-L84

Bafyn: also there is a test that I added for MySQL which checks the generated query and there's no schema in it

jzabroski: Good point. You're right.

Issue #2400, opened 2026-08-18
dotnet-fm 8.0.1: no -p value can apply migrations to Oracle on .NET 8 (same setup works with 6.2.0)


lucasrossoni:
**Describe the bug**

After upgrading from FluentMigrator 6.2.0 to 8.0.1, `dotnet-fm` can no longer apply migrations to an Oracle database on .NET 8. Every Oracle processor fails, in two different ways:

| `-p` value | Result on 8.0.1 |
|---|---|
| `OracleManaged`, `Oracle12cManaged` | Connects to the database, then throws `A migration generator with the ID OracleManaged couldn't be found` |
| `Oracle`, `Oracle12c` | Never connects: `Unable to load the driver. Attempted to load: Oracle.DataAccess` |

The exact same command, against the same database and the same migrations, works with 6.2.0.

Stack trace for `-p OracleManaged`:

```
Beginning Transaction
Rolling back transaction
Unhandled exception. System.InvalidOperationException: A migration generator with the ID OracleManaged couldn't be found. Available generators are: Db2, Db2ISeries, Oracle, Oracle12C, Firebird, Hana, MySql4, MySql5, MySql8, Oracle, Oracle12C, Oracle, Oracle12C, Postgres15_0, Postgres92, Postgres10_0, Postgres11_0, Postgres15_0, Redshift, Snowflake, SQLite, SqlServer2016, SqlServer2000, SqlServer2005, SqlServer2008, SqlServer2012, SqlServer2014, SqlServer2016
   at FluentMigrator.Runner.Generators.SelectingGeneratorAccessor.FindGenerator(IReadOnlyCollection`1 generators, String generatorId) in SelectingGeneratorAccessor.cs:line 86
   at FluentMigrator.Runner.Generators.SelectingGeneratorAccessor..ctor(IEnumerable`1 generators, IOptionsSnapshot`1 options, IOptionsSnapshot`1 processorSelectorOptions) in SelectingGeneratorAccessor.cs:line 63
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Constructor(Object obj, IntPtr* args)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
[...]

PR #2402, opened 2026-08-25
Fix #2400: give Oracle12cManaged a generator of its own

RaphaelFakhri:
Fixes the second half of #2400.

The first symptom reported there, `-p OracleManaged` failing with "A migration generator with the ID OracleManaged couldn't be found", was fixed by #2269 and is awaiting release. This addresses the second one, which is still live on `main`.

`Oracle12CManagedProcessor` answers to the processor id `Oracle12cManaged`. `SelectingGeneratorAccessor` falls back to the processor id when no generator id is configured, which is exactly what `-p` does, and no generator claimed that id: `AddOracle12CManaged()` registered `Oracle12CGenerator`, whose `GeneratorId` is `Oracle12c`. It is the only Oracle processor id with no generator counterpart, so that provider could never resolve one.

`Oracle12CManagedGenerator` and `IOracle12CManagedGenerator` mirror the `OracleManaged` pair that #2269 added, and `AddOracle12CManaged()` now registers them through the same `TryAdd` indirection `RegisterOracleManagedGenerator` uses. The generator overrides `GeneratorId` and nothing else, so no SQL changes.

It also declares `Oracle` and `Oracle12c` as `GeneratorIdAliases`, the way `OracleManagedGenerator` inherits the `Oracle` alias. Without that, an application registering only `AddOracle12CManaged()` and pinning `GeneratorId = Oracle12c` explicitly would have stopped resolving, since the managed generator takes over the `IOracle12CGenerator` registration.

Tests, in `OracleRunnerBuilderExtensionsTests`:

- `EveryOracleProcessorIdSelectsAGenerator` registers all four Oracle providers the way `FluentMigrator.DotNet.Cli/Setup.cs` does, sets the processor id, and resolves `IGeneratorAccessor`. Without the fix, only the `Oracle12cManaged` case fails, with the same exception and call site as the stack trace in #2400: "A migration generator with the ID Oracle12cManaged couldn't be found. Available generators are: Oracle, Oracle12C, OracleManaged, Oracle12C".
- `Oracle12CManagedGeneratorAnswersToTheUnmanagedIds` pins each of the three ids explicitly against a container holding only `AddOracle12CManaged()`.
- The existing `AddOracle12CManaged` case in `IsolatedRegistrationResolvesProcessorAndGenerator` expected `GeneratorIdConstants.Oracle12c`; it now expects `Oracle12cManaged`, matching the `AddOracleManaged` row directly above it.

`FluentMigrator.Tests.Unit` is 5448 passed, 0 failed on net8.0. The solution builds with the same 12 warnings as `main`.

One line of the Oracle token matrix snapshot changes, since the `Oracle12cManaged` section now reports `Oracle12CManagedGenerator` rather than `Oracle12CGenerator`. #2401 also touches that file; the edits are on different lines and independent.

### DIFF
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/IOracle12CManagedGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/IOracle12CManagedGenerator.cs
@@ -1,0 +1,27 @@
+#region License
+// Copyright (c) 2026, Fluent Migrator Project
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace FluentMigrator.Runner.Generators.Oracle;
+
+/// <summary>
+/// Defines the interface for the Oracle 12c Managed SQL generator used in FluentMigrator.
+/// </summary>
+/// <remarks>
+/// This interface extends <see cref="IOracle12CGenerator"/> to provide functionality specific to Oracle 12c via the managed driver.
+/// </remarks>
+public interface IOracle12CManagedGenerator : IOracle12CGenerator
+{
+}
--- a/src/FluentMigrator.Runner.Oracle/Generators/Oracle/Oracle12CManagedGenerator.cs
+++ b/src/FluentMigrator.Runner.Oracle/Generators/Oracle/Oracle12CManagedGenerator.cs
@@ -1,0 +1,32 @@
+#region License
+// Copyright (c) 2026, Fluent Migrator Project
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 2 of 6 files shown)


jzabroski:
@RaphaelFakhri This is not necessarily the way it should work.

Generators are about database engine versions within a dialect.

Processors wire in the database driver.

Ideally, a Generator should have no knowledge of the fact it is using Oracle12cManaged, because that is a database driver detail, not a database engine.

One gap is that [`dotnet-fm migrate` specifies a --provider argument](https://fluentmigrator.github.io/runners/dotnet-fm.html#command-reference), and does not allow differentiating between the two. I think that is the preferred way to handle this versus leaking database driver details into a Generator abstraction. What do you think?

RaphaelFakhri:
I agree with the layering principle, the driver should be invisible to the generator. The reason it leaks here is the fallback in `SelectingGeneratorAccessor`: when no generator id is configured it selects by the processor id, and the processor ids are the one place the driver is encoded. #2269 answered that for `OracleManaged` by minting a generator that claims the driverful id, and this PR mirrors it for 12c, so `main` already carries one generator with driver knowledge either way.

The fix that matches what you describe would be to make that fallback resolve through the processor instead: each processor knows its dialect, so mapping processor id to generator id at selection or registration time would let `-p Oracle12cManaged` pick the plain `Oracle12c` generator with no new generator types, and `OracleManagedGenerator` from #2269 could be retired in a later release. It also answers the dotnet-fm gap from your link: `--provider` stays the only knob users need, with a separate generator override possible for anyone who genuinely wants to split them.

That is a bigger change to selection than this stopgap, and it revisits what #2269 shipped, so I did not want to reach for it uninvited. Happy to rework this PR into that shape if you prefer it; otherwise this version keeps `Oracle12cManaged` consistent with what `OracleManaged` already does today. Which way do you want it?

jzabroski:
> It also answers the dotnet-fm gap from your link: `--provider` stays the only knob users need, with a separate generator override possible for anyone who genuinely wants to split them.

Sorry, I was saying that it is possible dotnet-fm could gain additional parameters: explicit `--generator` and `--processor` parameters, each passing in a distinct ID. When we originally added `--provider`, the reduction of provider to both generator and processor made sense as a short-hand, and it can remain around for simplicity. We should document the full cartesian product of possible mappings and what --provider maps to. This documentation will also help LLMs discover issues e.g. if the user is using Oracle12cManaged and resolution step. In the old https://fluentmigrator.github.io site archived at https://github/com/fluentmigrator/documentation/ , we had a partial matrix here we can start with. https://github.com/fluentmigrator/documentation/blob/master/snippets/supported-databases.md

PR #2404, opened 2026-08-25
Fix #1982: do not carry index creation options into the reversed delete

### LINKED ISSUE #1982: Auto reversing an index creation with the Online option
**Describe the bug**
A migration that inherits from the AutoReversingMigration and creates a non-clustered index with the online option throws an exception with this message:
_"The error was Only a clustered index can be dropped online."_

**To Reproduce**
Create a migration similar to this one:

```
[Migration(123)]
public class MyMigration : AutoReversingMigration
{
    public override void Up()
    {
        Create.Index().OnTable("MyTable")
            .OnColumn("ColumnA").Ascending()
            .WithOptions().Online();
    }
}
```

**Expected behavior**
The index should be dropped with no exception

 - OS: Windows
 - Platform: .NET 8
 - FluentMigrator version: 6.2.0
 - FluentMigrator runner: FluentMigrator.MSBuild
 - Database Management System: SQL Server 2022

**Additional context**
A workaround for this issue is not to use the `AutoReversingMigration`.

RaphaelFakhri:
Fixes #1982.

`CreateIndexExpression.Reverse()` returned `new DeleteIndexExpression { Index = Index.Clone() }`, and `DeleteIndexExpression.AdditionalFeatures` is `Index.AdditionalFeatures`, so the clone carried the creation options onto the drop. An `AutoReversingMigration` that created a non-clustered index with `.WithOptions().Online()` emitted:

```sql
DROP INDEX [TestIndex] ON [dbo].[TestTable1] WITH (ONLINE=ON);
```

which SQL Server rejects with "Only a clustered index can be dropped online." `DATA_COMPRESSION` leaks through the same path.

Additional features on an index describe how it gets created, so `Reverse()` now clears them on the clone when the index is not clustered. A clustered index keeps them: SQL Server can drop a clustered index online, so an auto-reversed `Create.Index(...).WithOptions().Clustered().Online()` still emits `WITH (ONLINE=ON)` on the drop. `IndexDefinition.IsClustered` is a provider-neutral model property set by the fluent `.Clustered()`, and `Clone()` already copies it, so the check adds no provider knowledge to Abstractions.

An explicit `Delete.Index(...).WithOptions().Online()` is untouched and still emits `WITH (ONLINE=ON)`.

I considered fixing it in the generator instead, by emitting `ONLINE` on a drop only where SQL Server accepts it, and rejected it: the delete fluent syntax cannot declare clusteredness, so gating on it there would silently break the existing, tested `Delete.Index(...).WithOptions().Online()` API. Doing it in `Reverse()` also keeps the behaviour consistent with `CreateTableExpression` and `CreateColumnExpression`, which already reverse to identity-only deletes.

Five tests: two in `SqlServer2005IndexTests` pinning the generated SQL for a reversed non-clustered `.Online()` index (bare `DROP INDEX`) and a reversed clustered `.Online()` index (`WITH (ONLINE=ON)` preserved), and three provider-neutral ones in `CreateIndexExpressionTests` covering that `Reverse()` names the same index, drops the additional features for a non-clustered index, and keeps them for a clustered one. Reverting `Reverse()` to a plain clone fails the two non-clustered tests; reverting only the clustered guard fails the two clustered tests. With the change, `FluentMigrator.Tests.Unit` is 5445 passed, 0 failed on net8.0.

### DIFF
--- a/test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
@@ -128,5 +128,53 @@ namespace FluentMigrator.Tests.Unit.Expressions
 
             Assert.That(processed.Index.SchemaName, Is.EqualTo("testdefault"));
         }
+
+        [Test]
+        public void ReverseReturnsDeleteIndexExpressionForTheSameIndex()
+        {
+            var expression = new CreateIndexExpression
+            {
+                Index = new IndexDefinition
+                {
+                    Name = "IX_Name",
+                    SchemaName = "TestSchema",
+                    TableName = "Table",
+                    Columns = new Collection<IndexColumnDefinition>
+                    {
+                        new IndexColumnDefinition { Name = "Name" },
+                    },
+                },
+            };
+
+            var reversed = (DeleteIndexExpression)expression.Reverse();
+
+            Assert.That(reversed.Index.Name, Is.EqualTo("IX_Name"));
+            Assert.That(reversed.Index.SchemaName, Is.EqualTo("TestSchema"));
+            Assert.That(reversed.Index.TableName, Is.EqualTo("Table"));
+        }
+
+        [Test]
+        public void ReverseDoesNotCarryAdditionalFeaturesOverToTheDeleteIndexExpression()
+        {
+            var expression = new CreateIndexExpression
+            {
+                Index = new IndexDefinition
+                {
+                    Name = "IX_Name",
+                    TableName = "Table",
+                    Columns = new Collection<IndexColumnDefinition>
+                    {
+                        new IndexColumnDefinition { Name = "Name" },
+                    },
+                },
+            };
+
+            expression.AdditionalFeatures["SomeCreateOnlyOption"] = true;
+
+            var reversed = (DeleteIndexExpression)expression.Reverse();
+
+            Assert.That(reversed.AdditionalFeatures, Is.Empty);
+            Assert.That(expression.AdditionalFeatures, Is.Not.Empty);
+        }
     }
 }
--- a/test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
+++ b/test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
@@ -176,5 +176,29 @@ namespace FluentMigrator.Tests.Unit.Expressions
             Assert.That(reversed.AdditionalFeatures, Is.Empty);
             Assert.That(expression.AdditionalFeatures, Is.Not.Empty);
         }
+
+        [Test]
+        public void ReverseKeepsAdditionalFeaturesForAClusteredIndex()
+        {
+            var expression = new CreateIndexExpression
+            {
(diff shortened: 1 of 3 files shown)


test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
```diff
@@ -128,5 +128,53 @@ public void WhenDefaultSchemaConventionIsChangedAndSchemaIsNotSetThenSetSchema()
 
             Assert.That(processed.Index.SchemaName, Is.EqualTo("testdefault"));
         }
+
+        [Test]
+        public void ReverseReturnsDeleteIndexExpressionForTheSameIndex()
```
jzabroski, line 133: We need a variation of this test that reverses a clustered index ; the current change to CreateIndexExpression would break users who _do_ have a clustered index on column(s) that are not the primary key and want to reverse it.

```c#
if (!Index.TryGetAdditionalFeature<SqlServerConstraintType>(
                            SqlServerExtensions.ConstraintType,
                            out var indexType)
                     || indexType == SqlServerConstraintType.NonClustered)
{
  // non-clustered case - do not clone OnlineIndex additional feature
}
else
{
  // clustered case - DO preserve OnlineIndex additional feature.
}
```

jzabroski (changes requested):

test/FluentMigrator.Tests/Unit/Expressions/CreateIndexExpressionTests.cs
RaphaelFakhri: Good catch, that behaviour change was the weak point of the first version. It came out cleaner than the snippet suggests: for `Create.Index` the builder's `.Clustered()` sets `IndexDefinition.IsClustered` directly, and the clone already carries it, so `Reverse()` now keeps the additional features when the cloned index is clustered and clears them otherwise. The `SqlServerConstraintType` additional feature is the constraint-side mechanism, so no SQL Server types need to leak into Abstractions for this.

Two tests added: an expression-level one pinning that a clustered index keeps its features through `Reverse()`, and the variation you asked for in `SqlServer2005IndexTests`, where a clustered index created with `.Online()` reverses to `DROP INDEX [TestIndex] ON [dbo].[TestTable1] WITH (ONLINE=ON);`. The non-clustered case still reverses to a bare `DROP INDEX`. Reverting just the `Reverse()` change fails both new tests. The PR description is updated to match.

jzabroski: I am referring to these two scenarios - I think you're only addressing example 1:

```C#
//Example 1
            Create.Index("foo")
                .OnTable("f")
                .InSchema("b")
                .OnColumn("x")
                .Ascending()
                .WithOptions()
                .Clustered();

//Example 2 .Clustered() lives in FluentMigrator.SqlServer namespace and uses AdditionalFeatures mentioned in previous PR comment
            Create.UniqueConstraint("foo")
                .OnTable("f")
                .WithSchema("b")
                .Column("x")
                .Clustered(); // in FluentMigrator.SqlServer, this is a SQL Server-specific extension method that specifies that the unique constraint should be clustered.
```

PR #2405, opened 2026-08-25
Fix #2170: report the version table creation from LoadVersionInfoIfRequired

### LINKED ISSUE #2170: MigrationRunner.LoadVersionInfoIfRequired always returns false even when it had to create the tables
**Describe the bug**
`MigrationRunner.LoadVersionInfoIfRequired` docstring says `<returns><c>true</c> if the table had to be created or <c>false</c> when it already existed</returns>`, but it returns `false` if the version info table had to be created. This happens because `VersionLoader` calls `LoadVersionInfo` in its ctor and `MigrationRunner.LoadVersionInfoIfRequired` checks happen already too late

https://github.com/fluentmigrator/fluentmigrator/blob/fe72ea2fc10687a65bb39832b2e9c3ed3f036d8a/src/FluentMigrator.Runner.Core/MigrationRunner.cs#L498

**To Reproduce**
Simply run `runner.LoadVersionInfoIfRequired()` on an empty database, no migrations needed 

```csharp
var serviceProvider = new ServiceCollection()
    .AddLogging(lb => lb.AddFluentMigratorConsole())
    .AddFluentMigratorCore()
    .ConfigureRunner(
        builder => builder
            .WithVersionTable(new VersionTableMetaData())
            .AddSqlServer2016()
            .WithGlobalConnectionString(connectionString) // empty database connection string
            .ScanIn(typeof(Program).Assembly).For.All())
    .BuildServiceProvider();

using var scope = serviceProvider.CreateScope();
var runner = serviceProvider.GetRequiredService<IMigrationRunner>();
Console.WriteLine(runner.LoadVersionInfoIfRequired()); // prints False
```

**Expected behavior**
`runner.LoadVersionInfoIfRequired()` returns `true` if it had to create version info table

**Additional context**
I'm trying to find out if this is the first time FluentMigrator is run against a database in which case I want run some custom code

```csharp
if (runner.LoadVersionInfoIfRequired())
{
    // custom code
}

runner.MigrateUp();
```

RaphaelFakhri:
Fixes #2170.

The XML doc says "true if the table had to be created or false when it already existed", and on a fresh database it always returns false. The reason is ordering: `LoadVersionInfoIfRequired` reads the `VersionLoader` property, and the default `VersionLoader` creates any missing version table from its own constructor, so the `AlreadyCreatedVersionTable` and `AlreadyCreatedVersionSchema` checks, which query the database live, run after the creation has already happened.

On the shape of the fix, since jzabroski's concern above was about taking on more than needed: correcting the documentation to "always false" would just document a method with no reason to exist, and the reporter's use is exactly what it promises. So this is additive only. `VersionLoader` gains `HasCreatedVersionSchema` and `HasCreatedVersionTable`, backed by the `_versionSchemaMigrationAlreadyRun` and `_versionMigrationAlreadyRun` fields it already tracks, and `LoadVersionInfoIfRequired` notes whether the loader was materialised by this very call and, if that materialisation performed the creation, returns true.

`IVersionLoader` is deliberately untouched. Adding a member to that public interface would break every external implementation on net48 and netstandard2.0, where default interface members are not available. A custom `IVersionLoader` therefore behaves exactly as it does today, and the interface remark says so.

`ConnectionlessVersionLoader` needs no change, since its `LoadVersionInfo` never creates anything.

Two tests in `VersionLoaderTests` drive a real `MigrationRunner` and a real `VersionLoader` through the container against a stateful processor mock whose existence flags flip when the create expressions are processed. Against master's source the fresh-database test fails by returning false. With the change, `FluentMigrator.Tests.Unit` is 5442 passed, 0 failed on net8.0.

No API or binary break. The one behaviour change is the intended one: with the default `VersionLoader`, `LoadVersionInfoIfRequired()` on a fresh database now returns true where it previously returned false, contradicting its own documentation. Second and later calls still return false, which is what the existing test pins.

### DIFF
--- a/test/FluentMigrator.Tests/Unit/VersionLoaderTests.cs
+++ b/test/FluentMigrator.Tests/Unit/VersionLoaderTests.cs
@@ -16,6 +16,7 @@
 //
 #endregion
 
+using System.Data;
 using System.Linq;
 
 using FluentMigrator.Expressions;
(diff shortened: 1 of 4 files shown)


test/FluentMigrator.Tests/Unit/VersionLoaderTests.cs
```diff
@@ -375,5 +376,69 @@ public void VersionDescriptionMigrationOnlyRunOnceEvenIfExistenceChecksReturnFal
 
             runner.Verify(r => r.Up(loader.VersionDescriptionMigration), Times.Once());
         }
+
+        [Test]
+        public void LoadVersionInfoIfRequiredReportsCreationPerformedByVersionLoaderInitialization()
```
jzabroski, line 381: Could this be re-written as a parameter-driven unit test, where the VersionLoader is either VersionLoader or ConnectionlessVersionLoader? This would mean this one unit test would give complete path coverage of `bool LoadVersionInfoIfRequired()`

jzabroski (changes requested):
Thanks, this is helpful. One concern I have is we recently introduced a feature to dotnet-fm and the [Preview](https://github.com/fluentmigrator/fluentmigrator/blob/6343659455e2d633f174959fcb6fa6d535fe8fc8/src/FluentMigrator.DotNet.Cli/MigratorOptions.cs#L54) mode to PreviewFromNothing. https://github.com/fluentmigrator/fluentmigrator/pull/2370 - While it is a draft, the intention is to merge this feature in.

I will work on merging in #2370, but in the mean time, can you add unit test improvements I flagged? I think we want 100% path coverage on this change given it has worked the way it has for ~13 years.

